PR #27, opened 2019-10-17
Standardize APIs for Sending Across Xpring SDK

keefertaylor:
XpringJ and XpringClient both take `send` parameters in the order of `amount`, `destination`, `sender`. Logically, this reads as "Send `amount` drops of XRP to `destination` from `sender`.

This is a breaking API change and will be accompanied with a major version bump. 

In a perfect world, we wouldn't make this change, *but* I'm about to break the public API by removing the protocol buffers anyway, so we might as well take this opportunity to make it better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,17 +150,17 @@ const { Wallet, XRPAmount, XpringClient } = require("xpring-js");
 const remoteURL = "grpc.xpring.tech:80";
 const xpringClient = XpringClient.xpringClientWithEndpoint(remoteURL);
 
-// Wallet which will send XRP
-const senderWallet = Wallet.generateRandomWallet();
-
-// Destination address.
-const address = "r3v29rxf54cave7ooQE6eE7G5VFXofKZT7";
-
 // Amount of XRP to send
 const amount = new XRPAmount();
 amount.setDrops("10");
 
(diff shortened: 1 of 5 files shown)


keefertaylor:
Any sort of model object in Xpring SDK is represented by a protocol buffer. [`Payment`](https://github.com/xpring-eng/xpring-common-protocol-buffers/blob/master/proto/payment.proto) is what you're looking for. 

I'm generally not in favor of having protocol buffers in the public API for the reasons stated in this PR: https://github.com/xpring-eng/Xpring-JS/pull/28. 

I'd also argue that exposing a payment object provides too high of an amount of configurability to the common client, who really doesn't care about what options they *can* set, but just wants sane defaults. 

In the future, when we have more payment types, you could imagine that we'd have a few easy to use APIs which were ultimately sending payments. The implementation of these methods would bind up a `Payment` protocol buffer, then pass it to a common method that accepts and sends a `Payment`. We could make that external method public, if we really wanted to. 

I'm going to merge this now, since this PR doesn't preclude us exposing some sort of internal Payment API and because there's full approvals on it. I'm happy to have a discussion about this (publicly if repo watchers have opinions) or internally, if folks have strong opinions that I have the wrong view here.

PR #31, opened 2019-10-28
Remove Protocol Buffers from Public APIs

keefertaylor:
Protocol buffers have proven hard for external developers to grok and they shouldn't need to learn how protocol buffers work to use Xpring SDK. Instead, provide native types in the API. 

Specifically: 
- Modify `getBalance` to return `BigInt` not a protocol buffer
- Modify `send` to take a `BigInt` amount instead of a protocol buffer.
- Modify `send` to return a transaction hash `string` instead of a protocol buffer
- Hook up tests for `transactionBlobToTransactionHash` (taken from xpring-common-js)

These are breaking API changes and will be accompanied with an appropriate version bump.

### DIFF
--- a/src/xpring-client.ts
+++ b/src/xpring-client.ts
@@ -72,12 +72,13 @@ class XpringClient {
    * @param drops A numeric string indicating the number of drops to send.
    * @param destination A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   public async send(
     amount: XRPAmount,
     destination: string,
     sender: Wallet
-  ): Promise<SubmitSignedTransactionResponse> {
+  ): Promise<string> {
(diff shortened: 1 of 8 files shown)


src/xpring-client.ts
```diff
@@ -61,22 +63,23 @@ class XpringClient {
...
    *
-   * @param drops A numeric string indicating the number of drops to send.
+   * @param drops A `BigInt` representing the number of drops to send.
    * @param destination A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   public async send(
-    amount: XRPAmount,
+    amount: BigInt,
```
Stormtv: Any chance we could make this `BigInt | string | number` then try to cast the amount as a BigInt? Overall I think the developer experience of being able to use types you are familiar with is an advantage.

(Maybe not number because they might not take into account the overflow)

PR #66, opened 2019-11-15
Support `LastLedgerSequence` in transactions

keefertaylor:
- Update `xpring-common-js`
- Wire a new RPC for retrieving the latest ledger sequence
- When sending, retrieve the last ledger sequence, pad by a constant amount, and add to `lastLedgerSequence` 
- Add tests

### DIFF
--- a/src/grpc-network-client.ts
+++ b/src/grpc-network-client.ts
@@ -7,7 +7,9 @@ import {
   GetFeeRequest,
   SubmitSignedTransactionRequest,
   SubmitSignedTransactionResponse,
-  grpcCredentials
+  grpcCredentials,
+  GetLatestValidatedLedgerSequenceRequest,
+  LedgerSequence
 } from "xpring-common-js";
 
 /**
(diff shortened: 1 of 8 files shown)


src/grpc-network-client.ts
```diff
@@ -68,6 +70,23 @@ class GRPCNetworkClient implements Networking.NetworkClient {
...
   }
+
+  public async getLatestValidatedLedgerSequence(
+    getLatestValidatedLedgerSequenceRequest: GetLatestValidatedLedgerSequenceRequest
+  ): Promise<LedgerSequence> {
+    return new Promise((resolve, reject): void => {
+      this.grpcClient.getLatestValidatedLedgerSequence(
+        getLatestValidatedLedgerSequenceRequest,
+        (error, response): void => {
+          if (error != null || response == null) {
```
Stormtv: `!= null` and `== null` checks if the left side value is `null or undefined` if you want to strictly check `null` then use `===`

src/xpring-client.ts
```diff
@@ -146,68 +149,84 @@ class XpringClient {
...
-            );
-          }
-
-          const submitSignedTransactionRequest = new SubmitSignedTransactionRequest();
-          submitSignedTransactionRequest.setSignedTransaction(
-            signedTransaction
-          );
+          return this.getLastValidatedLedgerSequence().then(
+            async ledgerSequence => {
+              if (accountInfo.getSequence() == undefined) {
```
Stormtv: `== undefined` checks if the getSequence() value is `null or undefined` if you want to strictly check `undefined` then use `===`

src/xpring-client.ts
```diff
@@ -146,68 +149,84 @@ class XpringClient {
...
 
   /* eslint-enable no-dupe-class-members */
 
+  private async getLastValidatedLedgerSequence(): Promise<number> {
+    const getLatestValidatedLedgerSequenceRequest = new GetLatestValidatedLedgerSequenceRequest();
+    const ledgerSequence = await this.networkClient.getLatestValidatedLedgerSequence(
+      getLatestValidatedLedgerSequenceRequest
+    );
+    const index = ledgerSequence.getIndex();
+    return index;
```
Stormtv: Is there a reason to not return on line 226?

src/xpring-client.ts
keefertaylor: Done

PR #97, opened 2019-12-16
Consume protocol buffers from Submodule

keefertaylor:
Consume protocol buffers from a submodule rather than reading them from xpring-common-js. 

This is useful because this library will have to choose which protocol buffers it uses when it is compiled for webpack. Consuming protocol buffers from this top level target means that xpring-common-js doesn't need to do anything fancy for webpack to just work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,11 +14,12 @@
   },
   "scripts": {
     "build": "tsc -d",
-    "clean": "rm -rf ./build",
+    "clean": "rm -rf ./generated ./dist ./build",
+    "pretest": "./scripts/regenerate_protos.sh && tsc",
     "lint": "eslint **/*.ts",
     "lint:fix": "eslint **/*.ts --fix",
-    "prepublishOnly": "npm run clean && npm test && tsc -d",
-    "test": "npm run lint && nyc mocha"
+    "test": "npm run lint && nyc mocha",
+    "prepublishOnly": "npm test && tsc -d"
   },
   "homepage": "https://github.com/xpring-eng/Xpring-JS#readme",
   "devDependencies": {
@@ -51,6 +52,7 @@
     "typescript-eslint-parser": "22.0.0"
   },
   "dependencies": {
+    "grpc-tools": "^1.8.0",
     "xpring-common-js": "1.1.13"
   },
   "nyc": {
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "typescript-eslint-parser": "22.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -13,12 +13,13 @@
     "url": "https://github.com/xpring-eng/Xpring-JS/issues"
   },
   "scripts": {
-    "build": "tsc -d",
-    "clean": "rm -rf ./build",
+    "build": "./scripts/regenerate_protos.sh && tsc -d",
```
tedkalaw: this seems fine, but do you think it's worth having an explicit build script? i feel like if we add one more step it would be nice to encapsulate it

scripts/regenerate_protos.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+echo "Regenerating Protocol Buffers"
+
```
tedkalaw: so we've talked about this a little and i think i understand the issue 

i think the comments below are super helpful wrt having there be two different outputs - it would be sorta nice if the motivation was at the top of this script as well, which i think is described pretty reasonably in the description of this PR

scripts/regenerate_protos.sh
```diff
@@ -0,0 +1,32 @@
...
+$PWD/node_modules/grpc-tools/bin/protoc \
+    --js_out=import_style=commonjs,binary:$TS_OUT_DIR \
+    --js_out=import_style=commonjs,binary:$JS_OUT_DIR \
+    --grpc_out=$TS_OUT_DIR \
+    --grpc_out=$JS_OUT_DIR \
+    --plugin=protoc-gen-grpc=`which grpc_tools_node_protoc_plugin` \
+    --proto_path=$PWD/xpring-common-protocol-buffers/proto \
+    $PWD/xpring-common-protocol-buffers/**/*.proto
+
+# Generate tyepscript declaration files.
```
tedkalaw: extreme nit: `typescript`

scripts/regenerate_protos.sh
keefertaylor: This is actually because we need to reference these relative to where they are. 

JS OUT DIR is where the compiled typescript goes, and the TS OUT DIR is where the typescript code running with ts-node will find the files.

package.json
keefertaylor: I think this will shake out when this library converges with Xpring-web.

keefertaylor:
Added updates to https://app.asana.com/0/1143722888069470/1153543128205838

PR #108, opened 2020-01-02
Generate Protocol Buffers from rippled.

keefertaylor:
Long term, [xpring-common-protocol-buffers](https://github.com/xpring-eng/xpring-common-protocol-buffers) is going to disappear and be replaced with an official and well reviewed implementation by the rippled team. 

The implementation in rippled is being done here: https://github.com/ripple/rippled/pull/3159.

To access the new rippled protocol buffers, a new submodule is added. Note that the submodule is `cjcobb23/rippled` (rather than `ripple/rippled`) since I can't figure out how to get a submodule to track an open PR. I filed an Asana task to track this update here: https://app.asana.com/0/1143722888069470/1155647690941937.

This change segregates the protocol buffers so that we can do the transition in several phases. As of now:
- /generated/*.proto : 'new' protocol buffers from rippled. Unused, as of now.
- /generated/legacy/*.proto : The same protocol buffers from xpring-common-protocol-buffers which we have been using. 

When we have fully migrated the library to rippled's protocol buffers we can simply delete the `legacy` folder to delete the migration.


scripts/regenerate_protos.sh
```diff
@@ -18,14 +24,46 @@ $PWD/node_modules/grpc-tools/bin/protoc \
...
-    $PWD/xpring-common-protocol-buffers/**/*.proto
+    --proto_path $PROTO_PATH \
+    ./rippled/src/ripple/proto/rpc/v1/*.proto
 
 # Generate typescript declaration files.
 $PWD/node_modules/grpc-tools/bin/protoc \
     --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts \
     --ts_out=$TS_OUT_DIR \
     --ts_out=$JS_OUT_DIR \
+    --proto_path=$PROTO_PATH \
```
cjcobb23: does this generate classes for ./rippled/src/ripple/proto/ripple.proto too?

keefertaylor:
I added a more complete plan for how I plan to do this migration in https://github.com/xpring-eng/Xpring-JS/pull/109. I think those details are mostly tangential to this stuff since they are actual code changes, whereas this is just infrastructure and scripting, but wanted to share if reviewers wanted more context.

scripts/regenerate_protos.sh
keefertaylor: No, this is the include directory for protocol buffers to look for `imports` in their files.

The protocol buffers that are matched by the wildcard on line 28 and 36, which only matches `.proto` files in the `rpc/v1` directory.

keefertaylor:
> Looks reasonable to me. Your comments in the `sh` script are very helpful for reviewing!
> 
> Out of curiosity, why are `xpring-common-protocol-buffers` and `rippled` imported as git submodules? Is there just no good way to package up a bunch of protobuf definitions and export them? It seems odd to have all of `rippled` as a submodule when we only care about a small portion of that codebase.

There's not a great way to do this. We could put them in their own minimal repository and have both rippled and xpring-js import them, but that just feels more complicated for the rippled team. 

We could also presumably set up some sort of copy / paste script that managed to update these protocols buffers whenever rippled updated theirs, but again, that feels like more complication. 

Lastly, we could choose to place everything at ripple into a mono-repo but I think that's unlikely to happen. 

AFAICT, the only downside of having rippled as a submodule is that it will take longer to clone this repository, but that's a one time problem for developers only. Nothing from rippled gets packaged and shipped, so I'm not positive it's a trade off worth optimizing too fully. 

I am *very* open to clever ideas here

cjcobb23:
> > Looks reasonable to me. Your comments in the `sh` script are very helpful for reviewing!
> > Out of curiosity, why are `xpring-common-protocol-buffers` and `rippled` imported as git submodules? Is there just no good way to package up a bunch of protobuf definitions and export them? It seems odd to have all of `rippled` as a submodule when we only care about a small portion of that codebase.
> 
> There's not a great way to do this. We could put them in their own minimal repository and have both rippled and xpring-js import them, but that just feels more complicated for the rippled team.
> 
> We could also presumably set up some sort of copy / paste script that managed to update these protocols buffers whenever rippled updated theirs, but again, that feels like more complication.
> 
> Lastly, we could choose to place everything at ripple into a mono-repo but I think that's unlikely to happen.
> 
> AFAICT, the only downside of having rippled as a submodule is that it will take longer to clone this repository, but that's a one time problem for developers only. Nothing from rippled gets packaged and shipped, so I'm not positive it's a trade off worth optimizing too fully.
> 
> I am _very_ open to clever ideas here

I am beginning to think it makes sense to put the proto files in their own minimal repo. I don't think it will be complicated for rippled (or anyone) to include that repo as a submodule. Including all of rippled just to get the proto files seems backwards to me, and sort of an anti pattern. Having a dedicated repo just for the proto files seems to be the method others are using to resolve this issue (https://www.bugsnag.com/blog/libraries-for-grpc-services , https://medium.com/namely-labs/how-we-build-grpc-services-at-namely-52a3ae9e7c35).

This is more of a long term goal though, and not incredibly pressing. What you have here is fine for now, in my opinion, and at some point in the future we can yank those proto files out of rippled into a separate repo.

keefertaylor:
> This is more of a long term goal though, and not incredibly pressing. What you have here is fine for now, in my opinion, and at some point in the future we can yank those proto files out of rippled into a separate repo.

Agree. We can always adjust this strategy in the future as needed. It's not hard to do as long as we're just migrating the location of the protocol buffers. As opposed to the protocol buffers themselves.

keefertaylor:
Seems like we have quorum here, so I'm going to go ahead and merge this. I'll see you all over on #109!

PR #109, opened 2020-01-03
Migrate existing code to `Legacy` subfolders

keefertaylor:

I propose doing this upgrade in a few stages:
- [x] Migrate existing code to a `legacy` subdirectory and namespace (This PR)

### DIFF
--- a/src/legacy/legacy-grpc-network-client.ts
+++ b/src/legacy/legacy-grpc-network-client.ts
@@ -1,22 +1,22 @@
-import * as Networking from "./network-client";
-import { XRPLedgerAPIClient } from "../generated/legacy/xrp_ledger_grpc_pb";
-import { AccountInfo } from "../generated/legacy/account_info_pb";
-import { Fee } from "../generated/legacy/fee_pb";
-import { GetFeeRequest } from "../generated/legacy/get_fee_request_pb";
-import { GetAccountInfoRequest } from "../generated/legacy/get_account_info_request_pb";
-import { SubmitSignedTransactionRequest } from "../generated/legacy/submit_signed_transaction_request_pb";
-import { SubmitSignedTransactionResponse } from "../generated/legacy/submit_signed_transaction_response_pb";
-import { GetLatestValidatedLedgerSequenceRequest } from "../generated/legacy/get_latest_validated_ledger_sequence_request_pb";
-import { LedgerSequence } from "../generated/legacy/ledger_sequence_pb";
-import { GetTransactionStatusRequest } from "../generated/legacy/get_transaction_status_request_pb";
-import { TransactionStatus } from "../generated/legacy/transaction_status_pb";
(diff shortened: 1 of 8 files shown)


hbergren:
@keefertaylor , can you rebase this on master so only the relevant commits are included? It looks like it has the same commits that have already been merged into master by https://github.com/xpring-eng/Xpring-JS/pull/108.

keefertaylor:
@hbergren rebased! sorry about that.

src/legacy/legacy-grpc-network-client.ts
```diff
@@ -1,22 +1,22 @@
...
+import { Fee } from "../../generated/legacy/fee_pb";
+import { GetFeeRequest } from "../../generated/legacy/get_fee_request_pb";
+import { GetAccountInfoRequest } from "../../generated/legacy/get_account_info_request_pb";
+import { SubmitSignedTransactionRequest } from "../../generated/legacy/submit_signed_transaction_request_pb";
+import { SubmitSignedTransactionResponse } from "../../generated/legacy/submit_signed_transaction_response_pb";
+import { GetLatestValidatedLedgerSequenceRequest } from "../../generated/legacy/get_latest_validated_ledger_sequence_request_pb";
+import { LedgerSequence } from "../../generated/legacy/ledger_sequence_pb";
+import { GetTransactionStatusRequest } from "../../generated/legacy/get_transaction_status_request_pb";
+import { TransactionStatus } from "../../generated/legacy/transaction_status_pb";
+import { LegacyNetworkClient } from "./legacy-network-client";
```
hbergren: Any specific reason you changed the import from `* as Networking` to `{ LegacyNetworkClient }`?

test/mocha.opts
```diff
@@ -1,4 +1,4 @@
 --require ./node_modules/ts-node/register
 --require ./node_modules/source-map-support/register
 --recursive
-test/*.ts
\ No newline at end of file
+test/**/*.ts
```
hbergren: nit: newline.

If you're using VS Code, you can punch these into your user settings:
```
    "files.trimTrailingWhitespace": true,
```

src/legacy/legacy-grpc-network-client.ts
keefertaylor: No. I think this was actually just a bad VSCode autocomplete at some point. Any reason you'd like me to namespace this object?

test/mocha.opts
keefertaylor: https://github.com/xpring-eng/Xpring-JS/pull/112

keefertaylor: Added newline. Sent PR to put VSCode settings into project repo.

src/legacy/legacy-grpc-network-client.ts
hbergren: I can't think of any, it was just a change that stood out to me as different from the other `import` changes.

PR #111, opened 2020-01-06
Network Clients on rippled protocol buffers

keefertaylor:
Create a new `NetworkClient` interface, `GRPCNetworkClient` which implements `NetworkClient` and a `FakeNetworkClient` for testing. 

These classes are just light syntactical wrappers around the underlying gRPC implementation which give us: 
- A facade abstraction, if we ever wanted to move off of gRPC
- Promise based callbacks on gRPC
- Testability through a Fake

# Background

# Status
   - [x] Create a new `NetworkClient` with fakes (this PR)
   - [ ] Create a new `DefaultXpringClient` with tests

### DIFF
--- a/src/grpc-network-client.ts
+++ b/src/grpc-network-client.ts
@@ -1,0 +1,82 @@
+import {
+  GetAccountInfoRequest,
+  GetAccountInfoResponse
+} from "../generated/rpc/v1/account_info_pb";
+import { GetFeeRequest, GetFeeResponse } from "../generated/rpc/v1/fee_pb";
+import { GetTxRequest, GetTxResponse } from "../generated/rpc/v1/tx_pb";
+import {
+  SubmitTransactionRequest,
+  SubmitTransactionResponse
+} from "../generated/rpc/v1/submit_pb";
+import { XRPLedgerAPIServiceClient } from "../generated/rpc/v1/xrp_ledger_grpc_pb";
+
+import { NetworkClient } from "./network-client";
+import { credentials } from "grpc";
+
+/**
+ * A GRPC Based network client.
+ */
+class GRPCNetworkClient implements NetworkClient {
+  private readonly grpcClient: XRPLedgerAPIServiceClient;
+
+  public constructor(grpcURL: string) {
+    this.grpcClient = new XRPLedgerAPIServiceClient(
+      grpcURL,
+      credentials.createInsecure()
+    );
+  }
+
+  public async getAccountInfo(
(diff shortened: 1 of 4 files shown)


src/grpc-network-client.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.grpcClient = new XRPLedgerAPIServiceClient(
+      grpcURL,
+      credentials.createInsecure()
+    );
+  }
+
+  public async getAccountInfo(
+    request: GetAccountInfoRequest
+  ): Promise<GetAccountInfoResponse> {
+    return new Promise((resolve, reject): void => {
```
Stormtv: Out of curiosity is there a reason you prefer the Promise syntax over the async await syntax?

keefertaylor: Can you elaborate on what you mean? I thought I would need to return a promise in order to use async / await?

Stormtv: By default returning inside an async function returns a Promise. so the code could look like this:

```js
  public async getAccountInfo(
    request: GetAccountInfoRequest
  ): Promise<GetAccountInfoResponse> {
    return this.grpcClient.getAccountInfo(request)
  }
```

No `try {} catch() {}` would be needed since you are just rejecting the promise (throwing) the error anyway.

Also not sure if you would still want to throw an error in the case of the `GetAccountInfoRequest` being null but I don't know if it made much sense.

keefertaylor: TypeScript yells at me for an incorrect signature here: 

```
src/legacy/legacy-grpc-network-client.ts:32:18 - error TS2554: Expected 2-4 arguments, but got 1.

32     return await this.grpcClient.getAccountInfo(getAccountInfoRequest);
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  generated/legacy/xrp_ledger_grpc_pb.d.ts:104:87
    104     public getAccountInfo(request: get_account_info_request_pb.GetAccountInfoRequest, callback: (error: grpc.ServiceError | null, response: account_info_pb.AccountInfo) => void): grpc.ClientUnaryCall;
                                                                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'callback' was not provided.


Found 1 error.
```

PR #114, opened 2020-01-06
Skeleton of new `DefaultXpringClient`

keefertaylor:
Create a new `DefaultXpringClient` which returns an `unimplemented` error for all methods. Future PRs will implement these methods piecemeal.

Bootstrap a single test case that checks unimplemented is thrown. 

# Background

# Status
   - [x] Create a new `DefaultXpringClient` with tests (this PR)


src/default-xpring-client.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param transactionHash The hash of the transaction.
+   * @returns The status of the given transaction.
+   */
+  public async getTransactionStatus(
+    transactionHash: string
+  ): Promise<TransactionStatus> {
+    throw new Error(XpringClientErrorMessages.unimplemented);
+  }
+
+  /* eslint-disable no-dupe-class-members */
```
hbergren: Do you actually have dupe class members?

src/default-xpring-client.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  public async send(
+    amount: BigInt | number | string,
+    destination: string,
+    sender: Wallet
+  ): Promise<string> {
+    throw new Error(XpringClientErrorMessages.unimplemented);
+  }
+
+  /* eslint-enable no-dupe-class-members */
```
hbergren: See above.

PR #115, opened 2020-01-08
Abstraction around raw transaction status

keefertaylor:
`XpringClientDecorator` will share a definition across the legacy protocol buffers (from `xpring-common-protocol-buffers`) and the rippled protocol buffers. 

This PR creates an abstraction that wraps the return type from this method in an interface. This will allow us to create a bridge layer between the old and new protocol buffers during the migration.

# Background

# Status
   - [x] Create a new `NetworkClient` with fakes (this PR)
   - [ ] Create a new `DefaultXpringClient` with tests

### DIFF
--- a/src/legacy/legacy-default-xpring-client.ts
+++ b/src/legacy/legacy-default-xpring-client.ts
@@ -7,7 +7,8 @@ import { GetTransactionStatusRequest } from "../../generated/legacy/get_transact
 import { Payment } from "../../generated/legacy/payment_pb";
 import { SubmitSignedTransactionRequest } from "../../generated/legacy/submit_signed_transaction_request_pb";
 import { Transaction } from "../../generated/legacy/transaction_pb";
-import { TransactionStatus as RawTransactionStatus } from "../../generated/legacy/transaction_status_pb";
+import { TransactionStatus as RawTransactionStatusResponse } from "../../generated/legacy/transaction_status_pb";
+import RawTransactionStatus from "../raw-transaction-status";
 import { XRPAmount } from "../../generated/legacy/xrp_amount_pb";
 import { LegacyNetworkClient } from "./legacy-network-client";
 import LegacyGRPCNetworkClient from "./legacy-grpc-network-client";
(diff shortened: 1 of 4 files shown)


src/legacy/legacy-default-xpring-client.ts
```diff
@@ -7,7 +7,8 @@ import { GetTransactionStatusRequest } from "../../generated/legacy/get_transact
 import { Payment } from "../../generated/legacy/payment_pb";
 import { SubmitSignedTransactionRequest } from "../../generated/legacy/submit_signed_transaction_request_pb";
 import { Transaction } from "../../generated/legacy/transaction_pb";
-import { TransactionStatus as RawTransactionStatus } from "../../generated/legacy/transaction_status_pb";
+import { TransactionStatus as RawTransactionStatusResponse } from "../../generated/legacy/transaction_status_pb";
```
hbergren: I don't see anywhere in the PR diff that `RawTransactionStatusResponse` is used. Do we need this import?

keefertaylor:
> My general question is for my own understanding. It looks like before, we were both importing RawTransactionStatus from xpring-common-js and from ../../generated/legacy/transaction_status_pb, and now Xpring-JS is responsible for its own RawTransactionStatus interface.

Yup. So before:
`TransactionStatus`: An enum we return for `getTransactionStatus` calls. 
`RawTransactionStatus`: A raw protocol buffer returned from the server that represents a transaction status

Now:
`TransactionStatus`: same as before
`RawTransactionStatus`: An interface which the previous `RawTransactionStatus` protocol buffer happens to conform to. 

> That interface is our bridge from the old world to the new, so clients can just use the interface and not worry about the underlying implementation. Is that all correct?

Correct. Now the `XpringClientDecorator` interface is decoupled from any protocol buffer implementation so we can move between them.

keefertaylor:
> Where is the transaction's outcome (transaction metadata) accessible?
It is not.

intelliot:
Transaction metadata is necessary even for payments only. If someone sends you a partial payment, you need to look at the metadata in order to know how much money the payment gave you.

PR #118, opened 2020-01-09
Get balance from rippled

keefertaylor:

Implement `getBalance` for the a Xpring Client which uses official rippled protocol buffers. 

The set of tests are the same set of tests LegacyXpringClient runs against which guarantees consistent behavior between the two implementations.

### Context of Change

xpring-common-js now supports legacy protocol buffers and rippled protocol buffers in version 2.0.0+. Upgrade the library and migrate to the new methods.

*Status*

   - [ ] Read only calls for new `DefaultXpringClient` (In progress in this PR)

### Type of Change

## Before / After

This code is still not accessible. A future PR will add a boolean which will let us switch between implementations. 

## Test Plan

Unit tests on CI continue to pass.  When a boolean switch is implemented then we will be able to run integration tests.

### DIFF
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -2,6 +2,13 @@ import { XpringClientDecorator } from "./xpring-client-decorator";
 import TransactionStatus from "./transaction-status";
 import { Wallet } from "xpring-common-js";
 import { TransactionStatus as RawTransactionStatus } from "../generated/legacy/transaction_status_pb";
+import GRPCNetworkClient from "./grpc-network-client";
+import { NetworkClient } from "./network-client";
+import {
+  GetAccountInfoRequest,
+  GetAccountInfoResponse
+} from "../generated/rpc/v1/account_info_pb";
+import { AccountAddress } from "../generated/rpc/v1/amount_pb";
 
 /* global BigInt */
 
@@ -12,6 +19,7 @@ import { TransactionStatus as RawTransactionStatus } from "../generated/legacy/t
  * Error messages from XpringClient.
  */
 export class XpringClientErrorMessages {
+  public static readonly malformedResponse = "Malformed Response.";
   public static readonly unimplemented = "Unimplemented.";
 }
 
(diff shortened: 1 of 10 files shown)


src/default-xpring-client.ts
```diff
@@ -31,7 +62,29 @@ class DefaultXpringClient implements XpringClientDecorator {
...
   public async getBalance(address: string): Promise<BigInt> {
-    throw new Error(XpringClientErrorMessages.unimplemented);
+    if (!Utils.isValidXAddress(address)) {
+      return Promise.reject(
+        new Error(XpringClientErrorMessages.xAddressRequired)
+      );
+    }
+
+    const account = new AccountAddress();
+    account.setAddress(address);
```
hbergren: This feels like a weird pattern. Why not `const account = new AccountFromAddress(address)`?

src/default-xpring-client.ts
```diff
@@ -31,7 +62,29 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+
+    const request = new GetAccountInfoRequest();
+    request.setAccount(account);
```
hbergren: See above.

test/fakes/fake-network-client.ts
```diff
@@ -49,24 +52,33 @@ export class FakeNetworkClientResponses {
...
     > = FakeNetworkClientResponses.defaultGetTxResponse()
   ) {}
 
   /**
    * Construct a default AccountInfoResponse.
    */
   public static defaultAccountInfoResponse(): GetAccountInfoResponse {
-    return new GetAccountInfoResponse();
+    const balance = new XRPDropsAmount();
+    balance.setDrops(10);
```
hbergren: It doesn't matter that this is a magic number, because we don't assert that we get back `10` in the test, just that a `balance` exists, correct?

keefertaylor: Yeah, we're just asserting existence. I could name this as a constant, export it and assert that it's exactly equal, but I kinda figured that was a bit over kill since this class is already pretty large and unwieldy.

src/default-xpring-client.ts
keefertaylor: `AccountAddress` is a generated class: https://github.com/ripple/rippled/blob/develop/src/ripple/proto/rpc/v1/amount.proto#L19

The JavaScript API for protocol buffers doesn't offer a builder or other pattern to assist here, hence the awkward syntax. I could wrap this into a helper if you think that's better?

keefertaylor:
nah, we should have a changelog, i'll make one in a few seconds.

keefertaylor:
Added a changelog

src/default-xpring-client.ts
hbergren: Hmmm. I think a helper would be beneficial, but I don't think it's necessary for this branch.

hbergren:
Have you rewritten the changelog to be relevant for `Xpring-JS`?

It looks like the changelog for `xpring-common-js`. If the changes/versions are the same, that's fine, but it looks like the last cut release of `xpring-js` was on Dec 17th, and it was 1.2.5, and there has been no `2.0.0` release yet:
https://github.com/xpring-eng/Xpring-JS/releases

PR #119, opened 2020-01-09
Add CODEOWNERS file

keefertaylor:
Add `keefertaylor@` as CODEOWNER


Stormtv:
Can we close this since it is already on master?

PR #130, opened 2020-01-18
Implement boolean switch for rippled protocol buffer implementation

keefertaylor:

Implement a boolean in `XpringClient` which allows clients to switch between protocol buffer implementations. 

This functionality unlocks integration tests. 

### Context of Change

xpring-common-js now supports legacy protocol buffers and rippled protocol buffers in version 2.0.0+. Upgrade the library and migrate to the new methods.

*Status*

   - [ ] Read only calls for new `DefaultXpringClient` (In progress)
   - [ ] Write calls for `DefaultXpringClient`
- [x] Add a boolean parameter to switch between implementations (this PR)

### Type of Change

## Before / After

Clients can now access rippled's protocol buffer implementation if they so desire. Most methods (all besides `getBalance` are stubbed. 

This change lets us add integration tests as we implement the other methods. Before we can turn on integration tests, I need a publicly accessible gRPC enabled node. I've filed https://ripplelabs.atlassian.net/browse/INS-2802?atlOrigin=eyJpIjoiZDcxNWU3OWJmNTU5NDgxYTkwYzgyODcxNzRkZjVjZTciLCJwIjoiaiJ9 to ask for this functionality. 

## Test Plan

CI Continues to work.


src/xpring-client.ts
```diff
@@ -18,11 +19,17 @@ class XpringClient {
    * The XpringClient will use gRPC to communicate with the given endpoint.
    *
    * @param grpcURL The URL of the gRPC instance to connect to.
+   * @param useNewProtocolBuffers If `true`, then the new protocol buffer implementation from rippled will be used. Defaults to false.
    */
-  public constructor(grpcURL: string) {
-    const defaultXpringClient = LegacyDefaultXpringClient.defaultXpringClientWithEndpoint(
+  public constructor(grpcURL: string, useNewProtocolBuffers: boolean = false) {
+    var defaultXpringClient: XpringClientDecorator = LegacyDefaultXpringClient.defaultXpringClientWithEndpoint(
```
Stormtv: I believe let would be more appropriate here.

src/xpring-client.ts
```diff
@@ -18,11 +19,17 @@ class XpringClient {
...
       grpcURL
     );
+    if (useNewProtocolBuffers) {
+      defaultXpringClient = DefaultXpringClient.defaultXpringClientWithEndpoint(
```
Stormtv: So if using the new proto buffers we initialize to the Legacy and then change it to the rippled client? It might be better practice to initialize the defaultXpringClient only once.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add an additional boolean parameter to `XpringClient`'s constructor which allows toggling between protocol buffer implementations.
```
hbergren: 👍 Good job updating the CHANGELOG!

src/xpring-client.ts
hbergren: nit: Prefer `let` to `var`.

`let` got added to JS in 2015, and is block scoped (so if it's inside an if statement, it won't be valid outside the if block), whereas `var` is function scoped (which can lead to odd behavior if your mind isn't trained in the ways of JS).

In general, `let` is preferred by the community now.

CHANGELOG.md
keefertaylor: Thank you. Really feel like I'm getting good at this whole "software" thing.

src/xpring-client.ts
keefertaylor: I changed this to using a ternary operator. It lets us use a `const` but I think it's slightly more readable. I don't care because it is functionally the same code and will go away eventually so just LMK if you prefer I revert.

src/xpring-client.ts
keefertaylor: Prettier actually re-wrote this when I merged in #139. Sorry about that.

keefertaylor: Acknowledged. Prettier rewrote this after I merged in #139 but I can also just use a ternary operator and make it a `const`.

src/xpring-client.ts
Stormtv: @keefertaylor Looks :fire: now!

PR #131, opened 2020-01-18
Implement getRawTransactionStatus

keefertaylor:

Implement `getRawTransactionStatus` with the new rippled protocol buffers. 

### Context of Change

`XpringClient` uses a decorator pattern where each implementation must conform to an interface. The interface required for the `getRawTransactionStatus` method requires it return a `RawTransactionStatus`. 

This PR implements the `getRawTransactionStatus` method of the decorator. 

*Status*

   - [ ] Read only calls for new `DefaultXpringClient` (In progress)
   - [ ] Write calls for `DefaultXpringClient`

### Type of Change

## Before / After

No change. This new class is not exposed to the world or hooked up yet. 

## Test Plan

CI Continues to work.


test/default-xpring-client-test.ts
```diff
@@ -1,6 +1,5 @@
-import chai, { assert } from 'chai'
```
keefertaylor: This was dead code which I removed.

keefertaylor:
I know this slightly lowers code coverage. The next PR will implement `getTransactionStatus` which uses the results of this method and adds full test coverage.

test/fakes/fake-network-client.ts
```diff
@@ -98,7 +99,17 @@ export class FakeNetworkClientResponses {
    * Construct a default getTx response.
    */
   public static defaultGetTxResponse(): GetTxResponse {
-    return new GetTxResponse()
+    const transactionResult = new TransactionResult()
+    transactionResult.setResult('tesSUCCESS')
+
+    const meta = new Meta()
```
hbergren: What is `meta`?

keefertaylor: See “Response Format” here: https://xrpl.org/tx.html

Meta is defined as “ Various metadata about the transaction.”. The protobufs are just mirroring the existing JSON API, so we accepted their structure at face value.

PR #134, opened 2020-01-18
Implement Functionality to Retrieve the Current Fee for the Ledger

keefertaylor:

Implement `getFee` for the  `XpringClient` which uses official rippled protocol buffers. 

This method is not hooked up. Test coverage will temporarily decrease. 

### Context of Change

*Status*

   - [ ] Read only calls for new `DefaultXpringClient` (In progress in this PR)

### Type of Change

## Before / After

Users who use the new protocol buffers will be able to check a transaction's status. 

## Test Plan

Unit tests on CI continue to pass.


src/default-xpring-client.ts
```diff
@@ -1,3 +1,19 @@
+<<<<<<< HEAD
```
hbergren: I'm gonna stop right here, because this can't be right 😄

keefertaylor:
PTAL

src/default-xpring-client.ts
```diff
@@ -174,6 +176,24 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+    const getFeeResponse = await this.networkClient.getFee(getFeeRequest)
+    const fee = getFeeResponse.getDrops()
+    if (!fee) {
+      throw new Error(XpringClientErrorMessages.malformedResponse)
+    }
+
+    const minimumFee = fee.getMinimumFee()
+    if (!minimumFee) {
+      throw new Error(XpringClientErrorMessages.malformedResponse)
+    }
```
tedkalaw: if you wanted to, you could do

```
class MalformedResponseError extends Error {}
```

then you can get tighter typing on the error that was thrown, and you could put the error message in there

keefertaylor: https://app.asana.com/0/1143722888069470/1158635908525161

PR #135, opened 2020-01-18
Implement LastLedgerSequence

keefertaylor:

Implement `getLastLedgerSequence` for the  `XpringClient` which uses official rippled protocol buffers. 

This method is not hooked up. Test coverage will temporarily decrease. 

### Context of Change

*Status*

   - [x] Read only calls for new `DefaultXpringClient` (In progress in this PR)

### Type of Change

## Before / After

Users who use the new protocol buffers will be able to check a transaction's status. 

## Test Plan

Unit tests on CI continue to pass.


src/default-xpring-client.ts
```diff
@@ -163,7 +163,8 @@ class DefaultXpringClient implements XpringClientDecorator {
   }
 
   public async getLastValidatedLedgerSequence(): Promise<number> {
-    throw new Error(XpringClientErrorMessages.unimplemented)
+    const getFeeResponse = await this.getFee()
```
intelliot: We need to get the fee in order to get the validated ledger sequence?

keefertaylor: Yes. The `Fee` rpc happens to contain this information and we only implemented the minimal subset of these APIs in rippled to make this happen so this is what we're using for now.

src/default-xpring-client.ts
```diff
@@ -177,11 +178,9 @@ class DefaultXpringClient implements XpringClientDecorator {
     return new GetTxResponseWrapper(getTxResponse)
   }
 
-  // TODO(keefertaylor): Add tests for this method once send is hooked up. 
-  private async getFee(): Promise<XRPDropsAmount> {
-    const getFeeRequest = new GetFeeRequest()
+  private async getMinimumFee(): Promise<XRPDropsAmount> {
```
hbergren: is `getMinimumFee` used right now? Maybe `git` is just making the diff look weird.

keefertaylor: Not yet. We're building toward #136 (implement send) which needs to get a fee and generate a last ledger sequence number (which requires the current ledger sequence number) in order to generate a transaction. 

In general, these helper methods are just mirroring the existing helper methods in `default-xpring-client`: https://github.com/xpring-eng/Xpring-JS/blob/master/src/legacy/legacy-default-xpring-client.ts#L200. There are differences in implementation in both that we're using a different set of protocol buffers and we implemented a slightly different subset of RPCs in rippled (they functionally serve the same set of data though).

keefertaylor:
It's fine. I wish that GitHub had a way to distinguish a chain of PRs and link them together. 

I should also be providing this context in my descriptions and will do so in the future.

keefertaylor:
We know that this is an outstanding task, and it's tracked here: https://app.asana.com/0/1143722888069470/1155647690941929

PR #136, opened 2020-01-18
Implement `send` with new protocol buffers

keefertaylor:

Implement `send` for the  `XpringClient` which uses official rippled protocol buffers. 

Add unit and integration tests. The unit tests are taken from the `LegacyDefaultXpringClient` tests, which ensure that the implementation behaves consistently across protocol buffers. 

### Context of Change

*Status*

- [x] Create code that works with the new protocol buffers from the rippled team
   - [x] Read only calls for new `DefaultXpringClient` (many PRs)
   - [x] Write calls fro `DefaultXpringClient`

### Type of Change

## Before / After

Users who use the new protocol buffers will be able to send XRP.
## Test Plan

Unit tests on CI continue to pass.  Integration test passes.


src/default-xpring-client.ts
```diff
@@ -159,11 +170,70 @@ class DefaultXpringClient implements XpringClientDecorator {
...
-    _sender: Wallet,
+    amount: BigInteger | number | string,
+    destination: string,
+    sender: Wallet,
   ): Promise<string> {
-    throw new Error(XpringClientErrorMessages.unimplemented)
+    if (!Utils.isValidXAddress(destination)) {
+      return Promise.reject(
+        new Error(XpringClientErrorMessages.xAddressRequired),
+      )
```
Stormtv: throw is same as return Promise.reject

src/default-xpring-client.ts
```diff
@@ -159,11 +170,70 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+    }
+
+    const classicAddress = Utils.decodeXAddress(sender.getAddress())
+    if (!classicAddress) {
+      return Promise.reject(
+        new Error(XpringClientErrorMessages.xAddressRequired),
+      )
```
Stormtv: Same

src/default-xpring-client.ts
```diff
@@ -159,11 +170,70 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+    const account = new AccountAddress()
+    account.setAddress(sender.getAddress())
+
+    const transaction = new Transaction()
+    transaction.setAccount(account)
+    transaction.setFee(fee)
+    transaction.setSequence(accountData.getSequence())
+    transaction.setPayment(payment)
+    transaction.setLastLedgerSequence(
+      lastValidatedLedgerSequence + ledgerSequenceMargin,
```
Stormtv: So do we need to use `lastValidatedLedgerSequence.getIndex()` here instead? This is similar to that balance thing we were talking about yesterday.

keefertaylor: `getLastValidatedLedgerSequence()` returns a `number` already, so this is a `number` + `number`.

Stormtv: My bad on this one, the old legacy protobuffs return an object.

src/default-xpring-client.ts
```diff
@@ -205,6 +271,43 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+   */
+  private static toBigInt(value: string | number | BigInteger): BigInteger {
+    if (typeof value === 'string') {
+      return bigInt(value)
+    }
+    if (typeof value === 'number') {
+      return bigInt(value)
+    }
+    // Value is already a BigInteger.
+    return value
```
intelliot: There should be no harm in returning `bigInt(someBigInteger)` so I think this entire function can be replaced with just `return bigInt(value)`

src/default-xpring-client.ts
```diff
@@ -8,20 +8,31 @@ import GRPCNetworkClientWeb from './grpc-network-client.web'
...
+import { Payment, Transaction } from './generated/web/rpc/v1/transaction_pb'
+import { AccountRoot } from './generated/web/rpc/v1/ledger_objects_pb'
 
 // TODO(keefertaylor): Re-enable this rule when this class is fully implemented.
 /* eslint-disable @typescript-eslint/require-await */
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable class-methods-use-this */
 
+/** A margin to pad the current ledger sequence with when submitting transactions. */
+const ledgerSequenceMargin = 10
```
intelliot: I recommend renaming `ledgerSequenceMargin` to `maxLedgerVersionOffset` for consistency with existing APIs, unless we have a good reason to introduce new terminology. https://xrpl.org/rippleapi-reference.html#transaction-instructions

keefertaylor: Done. Thanks for the tip.

src/default-xpring-client.ts
keefertaylor: This actually throws a bunch of typing errors (it wants a unary value, not a complex value type).

keefertaylor: Oh, this does work. Weirdly, VSCode complains but it compiles!

PR #139, opened 2020-01-21
Add ESLint + Prettier

hbergren:

Add the standardized JS/TS linting configuration to `xpring-common-js`.

### Notes for Reviewers

My commits are way less organized than this PR for `xpring-common-js`, but the idea is the same.

### Context of Change

The goal here is to get all of our JS/TS projects on the same configuration so we have some consistency across projects. This linting configuration is live on `ripple-keypairs` and will be migrated to the other `ripple-lib` style libraries moving forward.

Next steps after this PR is perhaps the wallet (with some tweaks for Jest + React).

### Type of Change

## Before / After

Now we have a more robust linting configuration, and Prettier for auto-formatting. Some linting errors were also fixed.

## Test Plan

Run the tests, they all still pass.


hbergren:
@keefertaylor , ESLint is pretty upset about the emptiness of `default-xpring-client`. I'm not sure what to do about that.

We could disable the violating ESLint rules like you do at the top under:
```
// TODO(keefertaylor): Re-enable this rule when this class is fully implemented.
/* eslint-disable @typescript-eslint/require-await */
```

keefertaylor:
That would be fine with me. I have a stack of PRs that I'd like to put out to start implementing this stuff but it will probably take at least a week. We can either hold this PR or disable the rules. Up to you. 

If you want to disable rules, could you file an Asana issue?

test/fakes/fake-xpring-client.ts
```diff
@@ -1,38 +1,44 @@
...
 
-    public async getLastValidatedLedgerSequence(): Promise<number> {
-        return Promise.resolve(this.getLastValidatedLedgerSequenceValue)
-    }
-    
-    public async getRawTransactionStatus(transactionHash: string): Promise<RawTransactionStatus> {
-       return Promise.resolve(this.getRawTransactionStatusValue);
-    }
+  public async send(
+    _amount: any,
```
Stormtv:
```suggestion
    _amount: BigInt | number | string,
```

src/default-xpring-client.ts
```diff
@@ -1,9 +1,7 @@
...
-import { Wallet } from "xpring-common-js";
-import { TransactionStatus as RawTransactionStatus } from "../generated/legacy/transaction_status_pb";
-
-/* global BigInt */
+import { Wallet } from 'xpring-common-js'
+import { XpringClientDecorator } from './xpring-client-decorator'
+import TransactionStatus from './transaction-status'
+import { TransactionStatus as RawTransactionStatus } from '../generated/legacy/transaction_status_pb'
 
 // TODO(keefertaylor): Re-enable this rule when this class is fully implemented.
```
Stormtv: I think we should just disable these under the TODO to re-enable these rules when completed
```suggestion
 // TODO(keefertaylor): Re-enable these rules when this class is fully implemented.
/* eslint-disable @typescript-eslint/require-await */
/* eslint-disable @typescript-eslint/no-empty-function */
/* eslint-disable @typescript-eslint/no-useless-constructor */
/* eslint-disable @typescript-eslint/no-unused-vars */
/* eslint-disable class-methods-use-this */
/* eslint-disable max-classes-per-file */
```

src/default-xpring-client.ts
```diff
@@ -1,9 +1,7 @@
-import { XpringClientDecorator } from "./xpring-client-decorator";
-import TransactionStatus from "./transaction-status";
-import { Wallet } from "xpring-common-js";
-import { TransactionStatus as RawTransactionStatus } from "../generated/legacy/transaction_status_pb";
-
-/* global BigInt */
+import { Wallet } from 'xpring-common-js'
```
Stormtv:
```suggestion
/* eslint-disable max-classes-per-file */
import { Wallet } from 'xpring-common-js'
```

hbergren:
I pulled in all the changes Tyler suggested, with one exception.

ESLint doesn't like defining multiple classes in the same file like you do for `default-xpring-client` with the helper `class XpringClientErrorMessages`.

i think in general that the 1 class per file thing is a good rule, but for this tiny helper class, I don't really care. I want to treat inline `/* eslint-disable */` rules as things we rip out, but since the 2 class per file for `xpring-default-client` is a reasonable exception, I added that directly to the `.eslintrc.js` config file.

I made an Asana task for re-enabling all the ESLint rules for this file in the backlog:
https://app.asana.com/0/1143722888069470/1158300650916617/f

PR #140, opened 2020-01-24
Integration tests for read gRPC calls on new rippled

keefertaylor:

Implement integration tests for the read only methods of rippled. 

Two bug fixes:
1) Address passed to rippled needs to be a classic address, not an x-address
2) tx needs a serialized transaction hash in bytes, not a hex string

### Context of Change

I implemented functionality for `getBalance` and `getTransactionStatus`. These new tests wire these against a live server. In practice, these tests are not hermetic, but they have proven not to be too flakey thus far. 

The server on the other side is a rippled node that I have stood up on an Ubuntu AWS box. It is publicly accessible and attached to the testnet network. 

These are analogs to the legacy tests in the same file. 

### Type of Change

[x] Tests

## Before / After

Code had no integration tests. Now code is tested! 

## Test Plan

I will not be testing the tests.


keefertaylor:
Note that these tests will fail because my new rippled node is not synced yet. It should be reasonably synced in a few hours and have passing tests then.

hbergren:
When you say "they have proven not to be too flakey thus far", what does that mean. They only fail 20% of the time? 2% of the time?

I prefer non-flakey tests so I can have faith in the test suite, but I understand that's hard for integration tests that are about actually talking to a server that we may not control.

Or can we make them non-flakey by having CircleCI spin up a Docker container with a `rippled` node? (Really just spitballing).

> Note that these tests will fail because my new rippled node is not synced yet. It should be reasonably synced in a few hours and have passing tests then.

The CircleCI build is still red, should it be green now?

Stormtv:
They are red due to a lack of typings in the xpring-common-js 3.0 I put a PR out to fix it here https://github.com/xpring-eng/xpring-common-js/pull/157

src/default-xpring-client.ts
```diff
@@ -171,9 +172,12 @@ class DefaultXpringClient implements XpringClientDecorator {
     transactionHash: string,
   ): Promise<RawTransactionStatus> {
     const getTxRequest = new GetTxRequest()
-    getTxRequest.setHash(transactionHash)
+    getTxRequest.setHash(
+      '95D30FED388C1B4BC1CC73BE3B8D79875995A26F7FB46061CF1A77972AEB9A7A'.toLowerCase(),
+    )
 
     const getTxResponse = await this.networkClient.getTx(getTxRequest)
+    console.log('response!')
```
Stormtv: Probably don't need this console.

keefertaylor:
> When you say "they have proven not to be too flakey thus far", what does that mean. They only fail 20% of the time? 2% of the time?
Fail 0% of the time. Once a month the transaction hash we use falls out of the rippled history (because the nodes are not full history). 

So like, < 1% of the time.

> I prefer non-flakey tests so I can have faith in the test suite, but I understand that's hard for integration tests that are about actually talking to a server that we may not control.

Same. I think that if they became flakey maybe we could have separate test suite that can run on demand. My point is that until they prove to be flakey it's probably not worth it. 

> Or can we make them non-flakey by having CircleCI spin up a Docker container with a rippled node? (Really just spitballing).

You need the rippled server to be in sync, which becomes prohibitively hard. 

> The CircleCI build is still red, should it be green now?

Looks like @Stormtv  has a fix.

keefertaylor:
I removed the install step which always installed the latest xpring-common-js.

test/integration-test.ts
```diff
@@ -14,12 +14,16 @@ const wallet = Wallet.generateWalletFromSeed('snYP7oArxKepd3GPDcrjMsJYiJeJB')!
...
 const transactionHash =
-  '9A88C8548E03958FD97AF44AE5A8668896D195A70CF3FF3CB8E57096AA717135'
+  '95D30FED388C1B4BC1CC73BE3B8D79875995A26F7FB46061CF1A77972AEB9A7A'
 
 // A XpringClient that makes requests. Uses the legacy protocol buffer implementation.
 const legacyGRPCURL = 'grpc.xpring.tech:80'
 const legacyXpringClient = new XpringClient(legacyGRPCURL)
 
+// A XpringClient that makes requests. Uses rippled's gRPC implementation.
+const rippledURL = '3.14.64.116:50051'
```
intelliot: Is it ok to hardcode this IP?

keefertaylor: This is a server that I am running and admin-ing outside of the ripple network as we are unsure if we want to open gRPC ports on a TestNet node.

PR #143, opened 2020-01-27
Polyfill BigInteger (2 of 2)

Stormtv:
BigInt is still not supported on the latest version of Safari. This PR remedies that problem by using [bit-integer ](https://www.npmjs.com/package/big-integer). Big-integer acts as a thin wrapper over the native implementation when BigInt exists and polyfills it where it doesn't

### Context of Change
This removes the native BigInts and switches to the polyfill version. In addition I removed the `Lib` settings in the tsconfig back to the defaults as it is no longer required

### Type of Change

## Before / After
No more BigInt only big-integer.

## Test Plan
Tests pass and npm link version in safari wallet works (working safari screenshot below)

Screen Shot 2020-01-21 at 6 31 21 PM


hbergren:
Should CircleCI be failing?

Stormtv:
> Should CircleCI be failing?

Its failing as the 1 of 2 PR is now failing. I'm working on a solution. I might rebase this onto master instead.

Stormtv:
@tedkalaw @intelliot I had to rebase this PR https://github.com/xpring-eng/Xpring-JS/pull/150

PR #150, opened 2020-01-28
Polyfill BigInt

Stormtv:
Had to rebase this https://github.com/xpring-eng/Xpring-JS/pull/143 (same code here)

BigInt is still not supported on the latest version of Safari. This PR remedies that problem by using [bit-integer ](https://www.npmjs.com/package/big-integer). Big-integer acts as a thin wrapper over the native implementation when BigInt exists and polyfills it where it doesn't

### Context of Change
This removes the native BigInts and switches to the polyfill version. In addition I removed the `Lib` settings in the tsconfig back to the defaults as it is no longer required

### Type of Change

## Before / After
No more BigInt only big-integer.

## Test Plan
Tests pass and npm link version in safari wallet works (working safari screenshot below)

Screen Shot 2020-01-21 at 6 31 21 PM

### DIFF
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -1,5 +1,6 @@
 import { Utils, Wallet } from 'xpring-common-js'
 import { XRPDropsAmount } from 'xpring-common-js/build/generated/rpc/v1/amount_pb'
+import bigInt, { BigInteger } from 'big-integer'
 import { XpringClientDecorator } from './xpring-client-decorator'
 import TransactionStatus from './transaction-status'
 import RawTransactionStatus from './raw-transaction-status'
@@ -95,9 +96,9 @@ class DefaultXpringClient implements XpringClientDecorator {
    * Retrieve the balance for the given address.
    *
    * @param address The X-Address to retrieve a balance for.
-   * @returns A `BigInt` representing the number of drops of XRP in the account.
+   * @returns A `BigInteger` representing the number of drops of XRP in the account.
    */
-  public async getBalance(address: string): Promise<BigInt> {
+  public async getBalance(address: string): Promise<BigInteger> {
     const classicAddress = Utils.decodeXAddress(address)
     if (!classicAddress) {
       return Promise.reject(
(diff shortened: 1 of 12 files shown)


Stormtv:
Added reminder to check in 6 months
https://app.asana.com/0/1143722888069470/1159440450999273

src/default-xpring-client.ts
```diff
@@ -121,7 +122,7 @@ class DefaultXpringClient implements XpringClientDecorator {
     if (!balance) {
       throw new Error(XpringClientErrorMessages.malformedResponse)
     }
-    return BigInt(balance)
+    return bigInt(balance.getDrops())
```
keefertaylor: why do you need to add getDrops here?

test/integration-test.ts
```diff
@@ -1,5 +1,6 @@
 import { Wallet } from 'xpring-common-js'
 import { assert } from 'chai'
+import bigInt from 'big-integer'
```
keefertaylor: Is this a class? If so should it be capitalized?

Stormtv: Sure the default BigInt is BigInt so we don't want to use that and the type is called BigInteger so I'm not sure we would want to use that either, but we can use something else.

src/default-xpring-client.ts
Stormtv: balance is an object
```
export namespace XRPDropsAmount {
    export type AsObject = {
        drops: number,
    }
}
```

We have to get the number to make it a bigInt (this should really be a string also).

Stormtv:
> Would it be idiomatic to extern BigInt from index.ts since it's an input parameter to a public API?

I don't think we need to as the only external API that takes BigInteger also takes string and number. In addition to that the native BigInt will also work when passing into that external function.

test/integration-test.ts
keefertaylor: That makes sense enough for me. I'm fine with this as is.

src/default-xpring-client.ts
keefertaylor: I investigated this. Tyler is definitely right here. What I think is happening is that the balance object has only one field, drops , if you call balance.toString() it just prints the number of drops. 

Presumably BigInt calls this method internally and thus this actually works, despite being semantically incorrect.

PR #156, opened 2020-01-30
Abstraction Cleanup for gRPC-Web + gRPC-Node

Stormtv:
Refactor the gRPC implementation to keep the web / node types outside of the xpring-clients.

### Context of Change
This refactor moves almost all of the generated content to live inside the gRPC network clients. I think going forward it makes a lot of sense to abstract away the generated content and not to have it in any other files besides the gRPC files.

### Type of Change

## Before / After
We no longer rely on generated types in the xpring clients, they now only live in the grpc network clients.

## Test Plan
CI passes

## Future Tasks
Re-organize the folder / file structor to better firewall this process.


src/default-xpring-client.ts
```diff
@@ -2,68 +2,19 @@ import { Utils, Wallet } from 'xpring-common-js'
...
-import RawTransactionStatus from './raw-transaction-status'
-import GRPCNetworkClient from './grpc-network-client'
+import RawTransaction from './raw-transaction'
+import GRPCNetworkClientNode from './grpc-network-client'
 import GRPCNetworkClientWeb from './grpc-network-client.web'
-import { NetworkClient } from './network-client'
-import { GetTxResponse } from './generated/web/rpc/v1/tx_pb'
-import { GetFeeResponse } from './generated/web/rpc/v1/fee_pb'
-import { XRPDropsAmount } from './generated/web/rpc/v1/amount_pb'
+import { XRPDropsAmount } from './generated/web/rpc/v1/amount_pb' // TODO would be nice to have 0 references to generated files here
```
keefertaylor: Humbly noted.  Want to add this as an actionable todo and format it with an owner:

```
// TODO(keefertaylor@): Do X which removes XRPDropsAmount dependency.
```

src/default-xpring-client.ts
```diff
@@ -2,68 +2,19 @@ import { Utils, Wallet } from 'xpring-common-js'
...
 import isNode from './utils'
+import { NetworkClient } from './network-client'
+import XpringClientErrorMessages from './xpring-client-error-messages'
```
keefertaylor: Sweet

src/legacy/legacy-grpc-network-client.ts
```diff
@@ -109,39 +115,45 @@ class LegacyGRPCNetworkClient implements LegacyNetworkClient {
...
-    return new GetTransactionStatusRequest()
-  }
-
-  public GetAccountInfoRequest(): GetAccountInfoRequest {
-    return new GetAccountInfoRequest()
-  }
-
-  public GetFeeRequest(): GetFeeRequest {
-    return new GetFeeRequest()
+  public async createSignedTransaction(
```
keefertaylor: What's the rationale for putting this inside of this class? Is it stylistic (you're trying to encapsulate protocol buffers here) or required by types?

src/legacy/legacy-grpc-network-client.web.ts
```diff
@@ -119,39 +126,45 @@ class LegacyGRPCNetworkClient implements LegacyNetworkClient {
...
-    return new GetTransactionStatusRequest()
-  }
-
-  public GetAccountInfoRequest(): GetAccountInfoRequest {
-    return new GetAccountInfoRequest()
-  }
-
-  public GetFeeRequest(): GetFeeRequest {
-    return new GetFeeRequest()
+  public async createSignedTransaction(
```
keefertaylor: Also: Can we dedupe this with the other class?

src/raw-transaction.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface RawTransactionStatus {
+  getValidated(): boolean
+  getTransactionStatusCode(): string
+  getLastLedgerSequence(): number
+}
+
+/**
+ * A wrapper class which conforms `GetTxResponse` to the `RawTransaction` interface.
+ */
+export default class RawTransaction {
```
keefertaylor: I'm not in love with this class name. This is an implementation of the RawTransactionStatus class so I think it should be 
<XYZ>RawTransactionStatus.  GetTxResponseRawTransactionStatus sounds like a handful though, which is why I had used the `Wrapper` name before. 

`RawTransaction` sounds like a transaction model object though, when really this is encapsulating a transaction status which is:
- the transaction
- some metadata

src/legacy/legacy-grpc-network-client.ts
Stormtv: The idea is to encapsulate all protocol buffers inside of the GRPC network clients to better handle the variation between Node buffers and Web buffers.

src/legacy/legacy-grpc-network-client.web.ts
Stormtv: I'll look into it, this would be a great improvement. The code here looks highly duplicative but the problem is that all the types here are technically different.

src/raw-transaction.ts
Stormtv: I agree 100%. I should have tried to make this a separate PR but this is something I really wanted to look into and optimize the functionality of. This is the only class outside of the grpc-network-client that uses the generated buffers which is something that needs to be resolved.

src/default-xpring-client.ts
```diff
@@ -2,68 +2,19 @@ import { Utils, Wallet } from 'xpring-common-js'
 import bigInt, { BigInteger } from 'big-integer'
 import { XpringClientDecorator } from './xpring-client-decorator'
 import TransactionStatus from './transaction-status'
-import RawTransactionStatus from './raw-transaction-status'
-import GRPCNetworkClient from './grpc-network-client'
+import RawTransaction from './raw-transaction'
```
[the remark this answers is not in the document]
Stormtv: It was renamed to raw transaction as it contains the class for the compatible transaction and also contains the status interface. This was a way to simplify the xpring-client

PR #160, opened 2020-01-30
Fix React Build Process

Stormtv:
Fix process checking for React Build.

React build has a process but not a process.release

### Context of Change
Check for existence for process release

### Type of Change

## Before / After
React should now build properly.

## Test Plan
npm link this into Xpring-Wallet-System ui on my `replace-xpring-web-xpring-js` branch

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,5 +1,9 @@
 const isNode = (): boolean => {
-  return typeof process !== 'undefined' && process.release.name === 'node'
+  return (
+    typeof process !== 'undefined' &&
+    typeof process.release !== 'undefined' &&
+    process.release.name === 'node'
+  )
 }
 
 export default isNode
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -1,5 +1,9 @@
 const isNode = (): boolean => {
-  return typeof process !== 'undefined' && process.release.name === 'node'
+  return (
+    typeof process !== 'undefined' &&
+    typeof process.release !== 'undefined' &&
+    process.release.name === 'node'
```
hbergren: This `typeof x !== 'undefined` seems a bit weird to me.

Can we not do optional chaining like `process?.release?.name === node`?

Stormtv: @hbergren 🤦‍♂ Yeah, much better lol

PR #179, opened 2020-02-10
Upgrade for breaking changes in rippled protocol buffers

keefertaylor:

Update to be compatible with breaking protocol buffer changes in rippled.  Breaking changes were introduced in  https://github.com/ripple/rippled/pull/3254. .

### Context of Change

No logic changes are introduced. This change does not touch the public API. 

Changes for protocol buffer compatibility and upgrade xpring-common-js to a version that has compatibility wiht the new protocol buffers

Changes in the protocol buffers are:
- Change folder / package location of files
- Use numeric strings instead of numbers to prevent overflows when working with 64 bit integers.
- Introduce new layers of abstraction internally.

### Type of Change

## Before / After

Compatible with latest rippled. 

## Test Plan
 CI

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -18,5 +18,7 @@
     },
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
-    "files.trimTrailingWhitespace": true
+    "files.trimTrailingWhitespace": true,
+
+    "mochaExplorer.files": "test/*.ts"
 }
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -141,16 +141,16 @@ class DefaultXpringClient implements XpringClientDecorator {
    * Send the given amount of XRP from the source wallet to the destination address.
    *
    * @param drops A `BigInteger`, number or numeric string representing the number of drops to send.
-   * @param destination A destination address to send the drops to.
+   * @param destinationAddress A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
    * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   public async send(
-    amount: BigInteger | number | string,
-    destination: string,
+    drops: BigInteger | number | string,
+    destinationAddress: string,
     sender: Wallet,
   ): Promise<string> {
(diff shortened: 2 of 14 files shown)


.vscode/settings.json
```diff
@@ -18,5 +18,7 @@
     },
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
-    "files.trimTrailingWhitespace": true
+    "files.trimTrailingWhitespace": true,
+
+    "mochaExplorer.files": "test/*.ts"
```
Stormtv:
```suggestion
    "mocha.files.glob": "test/**/*-test.ts",
    "mocha.requires": [
        "ts-node/register"
    ]
```
This fixed the sidebar for me

src/default-xpring-client.ts
```diff
@@ -143,39 +159,54 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+    destinationAccountAddress.setAddress(destinationAddress)
+
+    const destination = new Destination()
+    destination.setValue(destinationAccountAddress)
 
     const account = new AccountAddress()
     account.setAddress(sender.getAddress())
 
+    const account2 = new Account()
+    account2.setValue(account)
```
Stormtv: Better name for this, or at least explain why we have to do this?

src/default-xpring-client.ts
```diff
@@ -143,39 +159,54 @@ class DefaultXpringClient implements XpringClientDecorator {
...
+      lastValidatedLedgerSequence + maxLedgerVersionOffset,
+    )
+
+    const signingPublicKeyBytes = Utils.toBytes(sender.getPublicKey())
+    const signingPublicKey = new SigningPublicKey()
+    signingPublicKey.setValue(signingPublicKeyBytes)
+
     const transaction = new Transaction()
-    transaction.setAccount(account)
+    transaction.setAccount(account2)
```
hbergren: What is `account2` here? Why did this switch from `account` to `account2`? What is `account`?

Misc: Can we name `account2` something more descriptive?

keefertaylor: Yeah, sorry about that. leftover code from prototyping.

PR #181, opened 2020-02-10
Add `RippledFlags` class

keefertaylor:

XpringKit will need to interact with flags to work with rippled. Examples of these interactions include:
- Checking for partial payments
- Requiring canonical signatures

### Context of Change

This is a trio of changes implemented across Xpring SDK:
- JS: https://github.com/xpring-eng/Xpring-JS/pull/181
- Java: https://github.com/xpring-eng/xpring4j/pull/94
- Swift: https://github.com/xpring-eng/XpringKit/pull/110

### Type of Change

## Before / After

N/A - This new struct has no clients.

## Test Plan

N/A

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,5 +7,6 @@ export {
   WalletGenerationResult,
   Utils,
 } from 'xpring-common-js'
+export { default as RippledFlags } from './rippled-flags'
 export { default as TransactionStatus } from './transaction-status'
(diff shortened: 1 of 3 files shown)


keefertaylor:
I like to build at the leaf nodes and push things down into core libraries when they're needed. Otherwise, your core libraries are often just littered with things that people thought might be useful one day, which makes them hard to work with.

PR #182, opened 2020-02-11
[Partial Rollback] Rollback Mocha fix

keefertaylor:
## High Level Overview of Change

Partial rollback of https://github.com/xpring-eng/Xpring-JS/pull/138.

### Context of Change

Broke Mocha test explorer in sidebar. 

### Type of Change

[x] Rollback

## Before / After
After: Tests in sidebar of VSCode work, can attach debugger. 

## Test Plan

N/A

PR #186, opened 2020-02-11
Improves accountExists method to discern error codes

amiecorso:
accountExists method in XpringClient was treating any exception raised as a "false" conclusion (account does not exist).  This PR updates the method to conclude false only if a grpc.status.NOT_FOUND error code is in the exception.  Other codes will cause the exception to be re-thrown.  Unit and integration tests also updated to test ability to distinguish error codes.

### Type of Change

## Before / After
Before: accountExists could have returned misleading false-negatives. 
After: will propagate exceptions that it's not qualified to interpret.

## Test Plan
CI

### DIFF
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -1,5 +1,6 @@
 import { Signer, Utils, Wallet } from 'xpring-common-js'
 import bigInt, { BigInteger } from 'big-integer'
+import grpc from 'grpc'
 import { XpringClientDecorator } from './xpring-client-decorator'
 import TransactionStatus from './transaction-status'
 import RawTransactionStatus from './raw-transaction-status'
@@ -299,7 +300,10 @@ class DefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
+        return false
+      }
+      throw e // error code other than NOT_FOUND should re-throw error
     }
   }
 }
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -300,7 +300,10 @@ class DefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
+      if (
+        Object.prototype.hasOwnProperty.call(e, 'code') &&
+        e.code === grpc.status.NOT_FOUND
+      ) {
         return false
       }
       throw e // error code other than NOT_FOUND should re-throw error
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -300,10 +300,7 @@ class DefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      if (
-        Object.prototype.hasOwnProperty.call(e, 'code') &&
-        e.code === grpc.status.NOT_FOUND
-      ) {
+      if (e.code && e.code === grpc.status.NOT_FOUND) {
         return false
       }
       throw e // error code other than NOT_FOUND should re-throw error
--- a/src/legacy/legacy-default-xpring-client.ts
+++ b/src/legacy/legacy-default-xpring-client.ts
@@ -1,6 +1,7 @@
 import { Signer, Utils, Wallet } from 'xpring-common-js'
 
 import bigInt, { BigInteger } from 'big-integer'
+import grpc from 'grpc'
 import { AccountInfo } from '../generated/web/legacy/account_info_pb'
 import { XRPAmount } from '../generated/web/legacy/xrp_amount_pb'
 
@@ -244,7 +245,10 @@ class LegacyDefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
+        return false
+      }
+      throw e // error code other than NOT_FOUND should re-throw error
     }
   }
 }
--- a/src/legacy/legacy-default-xpring-client.ts
+++ b/src/legacy/legacy-default-xpring-client.ts
@@ -245,10 +245,7 @@ class LegacyDefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      if (
-        Object.prototype.hasOwnProperty.call(e, 'code') &&
-        e.code === grpc.status.NOT_FOUND
-      ) {
+      if (e.code && e.code === grpc.status.NOT_FOUND) {
         return false
       }
       throw e // error code other than NOT_FOUND should re-throw error
--- a/src/legacy/legacy-default-xpring-client.ts
+++ b/src/legacy/legacy-default-xpring-client.ts
@@ -245,7 +245,10 @@ class LegacyDefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
+      if (
+        Object.prototype.hasOwnProperty.call(e, 'code') &&
+        e.code === grpc.status.NOT_FOUND
+      ) {
         return false
       }
       throw e // error code other than NOT_FOUND should re-throw error
--- a/test/default-xpring-client-test.ts
+++ b/test/default-xpring-client-test.ts
@@ -439,8 +439,7 @@ describe('Default Xpring Client', function(): void {
   })
 
   it('Check if account exists - failing network request w/ NOT_FOUND error', async function() {
-    // GIVEN a XpringClient which wraps an erroring network client that returns an Error containing
-    // a property 'code' with value grpc.status.NOT_FOUND.
+    // GIVEN a XpringClient with a network client that will report accounts as not found
     const notFoundError = new FakeGRPCError(
       'FakeGRPCError: account not found',
       grpc.status.NOT_FOUND,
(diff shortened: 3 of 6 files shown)


.vscode/settings.json
```diff
@@ -18,5 +18,11 @@
     },
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
-    "files.trimTrailingWhitespace": true
+    "files.trimTrailingWhitespace": true,
+    "mocha.options": {},
+    "mocha.requires": [
+        "ts-node/register"
+    ],
+    "mocha.files.glob": "test/**/*.js"
```
Stormtv: This should probably be `"test/**/*-test.ts"`

test/mocha.opts
```diff
@@ -0,0 +1,5 @@
+--require ./node_modules/ts-node/register	
+--require ./node_modules/source-map-support/register	
+--recursive	
+test/**/*.ts	
+
```
Stormtv: Mocha opts is deprecated and was replaced by the .mocharc.js

https://mochajs.org/#mochaopts

src/default-xpring-client.ts
```diff
@@ -299,7 +300,10 @@ class DefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
```
Stormtv: I believe best practice here might be to use `hasOwnProperty` vs `in`

src/legacy/legacy-default-xpring-client.ts
```diff
@@ -244,7 +245,10 @@ class LegacyDefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if ('code' in e && e.code === grpc.status.NOT_FOUND) {
```
Stormtv: I believe best practice here might be to use `hasOwnProperty` vs `in`

.mocharc.js
```diff
@@ -1,23 +0,0 @@
...
-
-  // Warn if test exceed 75ms duration
-  slow: 75,
-
-  // Fail if tests exceed 10000ms
-  timeout: 10000,
-
-  // Check for global variable leaks
-  'check-leaks': true,
-}
```
Stormtv: Why was this deleted?

.vscode/settings.json
amiecorso: Cool! Do you mind explaining what this line is for?  I think it may have been auto-generated because I definitely didn't write it :)

.mocharc.js
amiecorso: hmm, I also don't know.  Also something I didn't touch.

src/legacy/legacy-default-xpring-client.ts
amiecorso: great, I was actually wondering which one was better so I'll change that.

amiecorso: So I just pushed that change  - the linter wanted me to write it like this: Object.prototype.hasOwnProperty.call(e, 'code')   instead of just e.hasOwnProperty('code').  Is that ok?

Stormtv: Yeah, thats perfect 👍

.vscode/settings.json
Stormtv: This is the workspace mocha configuration for the mocha test panel extension in VSCode.

`test/**/*.js` is telling it to look for files of the extension `.js` in a folder called test and all the folders underneath it. 

Our tests are written in Typescript so they have the `.ts` extension and they always follow the executable tests always follow the format of `name-test.ts`.

Glob patterns are interesting there is a great resource at https://globster.xyz/ to practice and see how they work

Stormtv:
I changed the branch this PR is compared to as I see now that #182 is what you were building on top of which is where those mocha changes came in.

test/mocha.opts
keefertaylor: Agree this change shouldn't be part of this PR. 

This change broke VSCode's test explorer and I'm not sure why.  I'm rolling it back here: https://github.com/xpring-eng/Xpring-JS/pull/182

test/default-xpring-client-test.ts
```diff
@@ -32,6 +33,23 @@ const fakeErroringNetworkClient = new FakeNetworkClient(
   FakeNetworkClientResponses.defaultErrorResponses,
 )
 
+/**
+ * Convenience class for creating more specific Error objects to mimick grpc error responses.
+ */
+class FakeServiceError extends Error {
```
keefertaylor: Maybe call this FakeGRPCError since it's mirroring gRPC semantics?

test/default-xpring-client-test.ts
```diff
@@ -32,6 +33,23 @@ const fakeErroringNetworkClient = new FakeNetworkClient(
...
+  /**
+   * Construct a new FakeServiceError.
+   *
+   * @param message The text details of the error.
+   * @param code The grpc.status code.
+   */
+  constructor(public message: string, public code: grpc.status) {
```
keefertaylor: I think that lines 48/49 are redundant (they get automatically added by putting a `public` modifier in front of these properties.

Could you also make them `readonly`? It doesn't really matter, but immutability is a good habit.

test/default-xpring-client-test.ts
```diff
@@ -436,17 +454,52 @@ describe('Default Xpring Client', function(): void {
...
+    const fakeNetworkClientResponses = new FakeNetworkClientResponses(
+      errorWithCode5, // getAccountInfoResponse
+    )
+    const fakeErroringNetworkClient = new FakeNetworkClient(
+      fakeNetworkClientResponses,
+    )
     const xpringClient = new DefaultXpringClient(fakeErroringNetworkClient)
 
-    // WHEN accountExists throws an exception while calling getBalance
+    // WHEN accountExists encounters this Error while calling getBalance
```
keefertaylor: Generally, encourage you to think of GIVEN, WHEN, THEN as a human reading it, rather than just writing what the test is doing.

You could make this as simple as:
// GIVEN A xpring client with a network client that will report accounts as not found
// WHEN Account existence is checked
// THEN The account is reported to exist

test/default-xpring-client-test.ts
```diff
@@ -436,17 +454,52 @@ describe('Default Xpring Client', function(): void {
     assert.equal(exists, true)
   })
 
-  it('Check if account exists - failing network request', async function() {
-    // GIVEN a XpringClient which wraps an erroring network client.
+  it('Check if account exists - failing network request w/ NOT_FOUND error', async function() {
+    // GIVEN a XpringClient which wraps an erroring network client that returns an Error containing
+    // a property 'code' with value grpc.status.NOT_FOUND.
+    const errorWithCode5 = new FakeServiceError(
```
keefertaylor: I would call this `notFoundError`, since `5` is kind of a magic number and this isn't a really descriptive name.

test/default-xpring-client-test.ts
```diff
@@ -436,17 +454,52 @@ describe('Default Xpring Client', function(): void {
...
     const exists = await xpringClient.accountExists(testAddress)
 
     // THEN accountExists returns false
     assert.equal(exists, false)
   })
 
+  it('Check if account exists - failing network request w/ CANCELLED error', function(done) {
+    // GIVEN a XpringClient which wraps an erroring network client that returns an Error containing
+    // a property 'code' with value grpc.status.CANCELLED.
+    const errorWithCode1 = new FakeServiceError(
```
keefertaylor: s/errorWithCode1/cancelledError

test/integration-test.ts
```diff
@@ -136,10 +136,25 @@ describe('Xpring JS Integration Tests', function(): void {
...
+    void
+  > {
     this.timeout(timeoutMs)
 
     const doesExist = await xpringClient.accountExists(recipientAddress)
+
     assert.equal(doesExist, true)
   })
+
+  // Omitting false-case tests for LegacyNode and LegacyWeb
```
keefertaylor: Can we be slightly more descriptive here? 

// Tests for account existence on the legacy shim are intentionally omitted as the legacy shim does have the required granularity of information.

test/legacy/legacy-default-xpring-client-test.ts
```diff
@@ -35,6 +35,23 @@ const transactionStatusFailureCodes = [
 
 const transactionHash = 'DEADBEEF'
 
+/**
```
keefertaylor: Definitely don't write this class twice. 

You could just make it's own file in `tests/fakes/fake-grpc-error.ts`) or stick it in `fake-network-client`. I'd probably recommend the former as `fake-network-client` is already becoming pretty long, but don't have strong opinions if you want to keep all the fake network stuff as one logical unit.

amiecorso:
> I changed the branch this PR is compared to as I see now that #182 is what you were building on top of which is where those mocha changes came in.

test/default-xpring-client-test.ts
amiecorso: agreed!

test/default-xpring-client-test.ts
amiecorso: way better

test/legacy/legacy-default-xpring-client-test.ts
amiecorso: great, this is in its own file now.

test/integration-test.ts
amiecorso: updated

test/default-xpring-client-test.ts
amiecorso: this makes sense too - less reference to actual names of things, and more of an abstract conception of what's happening

test/fakes/fake-grpc-error.ts
```diff
@@ -0,0 +1,18 @@
+import grpc from 'grpc'
+/**
```
keefertaylor: nit: one vertical blank line after imports

test/fakes/fake-grpc-error.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Convenience class for creating more specific Error objects to mimick grpc error responses.
```
keefertaylor: also super nit: `gRPC`

src/default-xpring-client.ts
```diff
@@ -299,7 +300,13 @@ class DefaultXpringClient implements XpringClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if (
+        Object.prototype.hasOwnProperty.call(e, 'code') &&
```
keefertaylor: I'm surprised you can't use typescripts optional chaining here:
```
if e.code? === grpc.status.NOT_FOUND
```

If code is `undefined` we'll safely unwrap it to undefined and the check will fail. Otherwise, it'll be unwrapped to a status code and compared.

test/fakes/fake-grpc-error.ts
amiecorso: does this mean we want to ```import grpc from 'gRPC'``` ?

amiecorso: oh you must mean gRPC in the comment, because it doesn't know what gRPC is as an import

src/default-xpring-client.ts
amiecorso: So I'm pretty sure we actually can use this construct.  I'll double-check with typescript people here to make sure nothing stupid happens, but if they approve I'll change that.

amiecorso: so the optional chaining syntax doesn't seem to work in that particular way, but I updated the if statement so it's a bit shorter.

PR #192, opened 2020-02-19
Refactor `RawTransactionStatus` to be a class rather than an interface

keefertaylor:

In order to bucket transaction statuses as `unknown` we need to have additional derived fields in the `RawTransactionStatus` object.

### Context of Change

Recall that there are two sets of protocol buffers (one from xpring-common-protocol-buffers, "legacy") and a new set from rippled. `RawTransactionStatus` is a wrapper class that provides a bridge between these two protocol buffer implementations. 

### Type of Change

## Before / After

No change in behavior

## Test Plan

CI

### DIFF
--- a/src/raw-transaction-status.ts
+++ b/src/raw-transaction-status.ts
@@ -6,7 +6,7 @@ export default class RawTransactionStatus {
   /**
    * Create a RawTransactionStatus from a TransactionStatus legacy protocol buffer.
    */
-  public static fromTransactionStatus(
+  static fromTransactionStatus(
     transactionStatus: TransactionStatus,
   ): RawTransactionStatus {
     return new RawTransactionStatus(
@@ -19,9 +19,7 @@ export default class RawTransactionStatus {
   /**
    * Create a RawTransactionStatus from a GetTxResponse protocol buffer.
    */
-  public static fromGetTxResponse(
-    getTxResponse: GetTxResponse,
-  ): RawTransactionStatus {
+  static fromGetTxResponse(getTxResponse: GetTxResponse): RawTransactionStatus {
     return new RawTransactionStatus(
       getTxResponse.getValidated(),
       getTxResponse
@@ -35,7 +33,7 @@ export default class RawTransactionStatus {
   /**
    * Note: This constructor is exposed for testing purposes. Clients of this code should favor using a static factory method.
    */
-  public constructor(
+  constructor(
     public isValidated,
     public transactionStatusCode,
     public lastLedgerSequence,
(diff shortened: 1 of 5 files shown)


src/raw-transaction-status.ts
```diff
@@ -1,6 +1,43 @@
+import { TransactionStatus } from './generated/web/legacy/transaction_status_pb'
+import { GetTxResponse } from './generated/web/rpc/v1/tx_pb'
+
 /** Abstraction around raw Transaction Status for compatibility. */
-export default interface RawTransactionStatus {
-  getValidated(): boolean
-  getTransactionStatusCode(): string
-  getLastLedgerSequence(): number
+export default class RawTransactionStatus {
```
hbergren: Was this class/interface part of the public API of `xpring-js`? Is this a breaking change?

src/raw-transaction-status.ts
```diff
@@ -1,6 +1,43 @@
...
+   */
+  public static fromGetTxResponse(
+    getTxResponse: GetTxResponse,
+  ): RawTransactionStatus {
+    return new RawTransactionStatus(
+      getTxResponse.getValidated(),
+      getTxResponse
+        .getMeta()
+        ?.getTransactionResult()
+        ?.getResult(),
```
hbergren: what a deep chain of properties

src/raw-transaction-status.ts
```diff
@@ -1,6 +1,43 @@
...
 /** Abstraction around raw Transaction Status for compatibility. */
-export default interface RawTransactionStatus {
-  getValidated(): boolean
-  getTransactionStatusCode(): string
-  getLastLedgerSequence(): number
+export default class RawTransactionStatus {
+  /**
+   * Create a RawTransactionStatus from a TransactionStatus legacy protocol buffer.
+   */
+  public static fromTransactionStatus(
```
hbergren: Everything in a TypeScript class is `public` by default.

Do you find it valuable to explicitly state that things are `public`? (If you do, I wonder if we should enforce that in ESLint across our projects somehow).

src/raw-transaction-status.ts
keefertaylor: No. It's not exported in `index.js` and it's also not available in the public API of anything that is. 
 This is just an internal construct. 

When the legacy protocol buffers go away, so will this class.

src/raw-transaction-status.ts
keefertaylor: I tend to value being explicit and in other languages you don't get public by default. I'll roll these changes back.

keefertaylor:
Removed the public modifier. Happier to conform to what everyone else is doing than to make us all change based on my opinions.

PR #193, opened 2020-02-19
(1/2) Add isFullPayment to RawTransactionStatus

keefertaylor:

Adds a new `isFullPayment` to `RawTransactionStatus` which allows us to know if XpringSDK can bucket the transaction. 

### Context of Change

Recall that there are two sets of protocol buffers (one from xpring-common-protocol-buffers, "legacy") and a new set from rippled. `RawTransactionStatus` is a wrapper class that provides a bridge between these two protocol buffer implementations. 

Xpring SDK buckets transactions into several states: pending, failed, succeeded and unknown. We can only bucket non-partial payment transactions at the moment. 

The legacy protocol buffers will go away soon and do not report enough information to determine if they are not a full payment. For this reason, we always report them as full payments. This decision got cleared with product. 

The new protocol buffers do provide enough information. 

### Type of Change

## Before / After

No change.  A future PR will hook up `XpringClient` to return an UNKNOWN status code.

## Test Plan

Added unit tests

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,5 +2,10 @@
 src/generated
 build
 
+# Do not lint submodules
+hermes-ilp
+xpring-common-protocol-buffers
+rippled
+
 # Don't lint node_modules
 node_modules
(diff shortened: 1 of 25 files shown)


keefertaylor:
@Stormtv Thanks. That was probably an awful name. 

I went with 'isFullPayment'.

PR #194, opened 2020-02-19
Skeleton for Transaction History

keefertaylor:

Wire a skeleton for Transaction history. Note these methods are not exposed publicly via `XpringClient` and just raise exceptions. 

### Context of Change

rippled is going to implement the `account_history` endpoint in [this PR](https://github.com/ripple/rippled/pull/3254). 

This is a pre-factor to put infrastructure in place to implement that end point while we wait for the code to land. 

### Type of Change

## Before / After

N/A

## Test Plan

CI

### DIFF
--- a/src/default-xpring-client.ts
+++ b/src/default-xpring-client.ts
@@ -15,6 +15,7 @@ import {
 import isNode from './utils'
 import { Payment, Transaction } from './generated/web/rpc/v1/transaction_pb'
 import { AccountRoot } from './generated/web/rpc/v1/ledger_objects_pb'
+import XRPTransaction from './xrp-transaction'
 
 /** A margin to pad the current ledger sequence with when submitting transactions. */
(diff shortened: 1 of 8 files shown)


src/default-xpring-client.ts
```diff
@@ -267,6 +268,17 @@ class DefaultXpringClient implements XpringClientDecorator {
       return false
     }
   }
+
+  // TODO(keefertaylor): Re-enable
+  /* eslint-disable class-methods-use-this */
+  /* eslint-disable @typescript-eslint/no-unused-vars */
```
tedkalaw: seems like it might be worth adding this lint rule: https://github.com/xpring-eng/xpring-wallet-system/pull/373

keefertaylor:
I'm going to pull this back and bring this more in line with: https://github.com/xpring-eng/XpringKit/pull/124

PR #205, opened 2020-02-28
Update documentation to point towards hosted nodes rather than shimming service

keefertaylor:

We're confident in the rippled node's ability to serve the demo SDK and want to point SLOHack users to these endpoints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,10 @@ To get developers started right away, Xpring currently hosts nodes. These nodes 
 
 ```
 # TestNet
-3.14.64.116:50051
+alpha.test.xrp.xpring.io:50051
 
 # MainNet
-18.222.225.81:50051
+alpha.xrp.xpring.io:50051
 ```
 
 ## Usage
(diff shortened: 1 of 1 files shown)


dpurdy:
could we get:
xrp.xpring.io
test.xrp.xpring.io

when we can and update this text when we can?

@wliu ? @asana: make me a task?

dpurdy:
this is better than what we have now? but would love to get to some cool hostnames for the cool kids? how can i help?

README.md
```diff
@@ -126,8 +126,8 @@ wallet.verify(message, signature); // true
 ```javascript
 const { XpringClient } = require("xpring-js");
 
-const remoteURL = "grpc.xpring.tech:80";
-const xpringClient = XpringClient.xpringClientWithEndpoint(remoteURL);
+const remoteURL = "3.14.64.116:50051"; // TestNet URL, use 18.222.225.81:50051 for MainNet
+const xpringClient = XpringClient.xpringClientWithEndpoint(remoteURL, true);
```
amiecorso: This is the only xpringClient that is constructed with `xpringClientWithEndpoint`... is that correct or should this also use `new` ?

keefertaylor: Thank you!

PR #212, opened 2020-03-03
(2/2) Implement skeleton for getTransactionHistory

keefertaylor:

Implement a skeleton of `getTransactionHistory` and tests inside of XpringClient.

Analogs:
JS: https://github.com/xpring-eng/Xpring-JS/pull/212
Java: https://github.com/xpring-eng/xpring4j/pull/116
Swift: https://github.com/xpring-eng/XpringKit/pull/124

### Context of Change

rippled recently implemented this RPC which we'd like to use.

Note that this functionality is not exposed in `XpringClient`, so this is still internal only. Currently `Transaction` objects are just empty containers. 

A future PR will:
- Update `Transaction` to have fields and to set them correctly from the protocol buffers
- Expose this as a public API

### Type of Change

## Before / After

N/A - still internal only functionality. 

## Test Plan

New unit tests provided for CI.


test/default-xrp-client-test.ts
```diff
@@ -466,4 +466,44 @@ describe('Default Xpring Client', function(): void {
...
+  > {
+    // GIVEN a DefaultXpringClient.
+    const xpringClient = new DefaultXRPClient(fakeSucceedingNetworkClient)
+
+    // WHEN the transaction history for an address is requested.
+    const transactionHistory = await xpringClient.getTransactionHistory(
+      testAddress,
+    )
+
+    // THEN the transaction hisotry is returned.
```
mayurbhandary: Small typo

PR #213, opened 2020-03-03
Update ILP client to align with Xpring4j

nkramer44:

- `sendMoney` and `getBalance` now return gRPC objects instead of single values
- `sendMoney` and `getBalance` now take an optional `bearerToken`.  If using the SDK from node, user will have to specify the `bearerToken` (ie API key), but from the web this will come automatically from a cookie (at least in xpring.io)

### Context of Change

Want to have Xpring4j and Xpring-JS aligned.  Original go at Xpring-JS ILP client was a rough draft, whereas this implementation will actually work with Hermes gRPC.

### Type of Change

However this hasn't been released/used by anyone yet, so not much to worry about there.

## Test Plan

Need to expand on `ilp-integration-tests.ts`.  Currently it does not work with `hermes-grpc.ilpv4.dev`, our main instance of Hermes, because SSL isn't set up.  I have run these tests locally with a local Hermes running and they work.  Once we set up SSL we should be able to test ILP client functionality on real infrastructure


src/ilp-network-client.ts
```diff
@@ -8,19 +8,27 @@ export interface IlpNetworkClient {
...
    *
    * @param request the details required for fetching the balance
-   * @returns a response with details about the balance including the type of currency and amount
+   * @param bearerToken
+   * @returns A {@link GetBalanceResponse} with balance information of the specified account
    */
-  getBalance(request: GetBalanceRequest): Promise<GetBalanceResponse>
+  getBalance(
+    request: GetBalanceRequest,
+    bearerToken?: string,
```
keefertaylor: I think you can default these to be `bearerToken?: string = null` and then you don't have to pass dummy values in your tests, etc.

test/ilp-integration-test.ts
```diff
@@ -1,19 +1,49 @@
+/*
```
keefertaylor: Did you mean to disable this?

keefertaylor:
\+ devx-reviewers for a second pair of eyes

test/ilp-integration-test.ts
nkramer44: Yeah these don't work because Hermes has SSL set up.  Once we set that up in the SDK we can enable these tests

src/ilp-network-client.ts
nkramer44: I actually don't think I have to pass in dummy values in tests anyway.  The `?` will default it to `undefined` if not passed in, if I'm understanding TS correctly

nkramer44:
@keefertaylor @xpring-eng/devx-reviewers I enabled SSL in `GrpcIlpNetworkClient` and uncommented ILP integration tests

test/default-ilp-client-test.ts
```diff
@@ -22,10 +22,10 @@ describe('Default ILP Client', function(): void {
     const client = fakeSuceedingNetworkClient()
 
     // WHEN the balance for an account is requested
-    const amount = await client.getBalance('test.foo.bar')
+    const amount = await client.getBalance('test.foo.bar', 'Bearer blah')
```
dino-rodriguez: I think it might be worth keeping the non-bearer token case as well since it is an optional argument

nkramer44: Woops, forgot to take it out of one test.

src/grpc-ilp-network-client.ts
```diff
@@ -18,20 +18,26 @@ class GrpcIlpNetworkClient implements IlpNetworkClient {
       // FIXME wrong credentials
```
keefertaylor: I think you can remove this FIXME now?

test/ilp-integration-test.ts
```diff
@@ -1,19 +1,48 @@
...
 
+import assert = require('assert')
+
 // A timeout for these tests.
 const timeoutMs = 60 * 1000 // 1 minute
 
 // A ILP Client that makes requests.
 const ILPAddress = 'hermes-grpc.ilpv4.dev'
-const ILPClient = new IlpClient(ILPAddress)
+const ILPClientNode = new IlpClient(ILPAddress, false)
```
keefertaylor: nit: You can omit `false` as a second parameter here because it's the default.

nkramer44:
@dino-rodriguez, @keefertaylor can better answer that question.  My understanding is that it better hides the implementation details of the SDK, and allows us to easily change things without affecting the public API

PR #215, opened 2020-03-04
[PayID] Initial API / Skeleton

keefertaylor:

Initial commit of PayID objects in JS. Main classes:
- `PayIDError`: Errors thrown from a PayID component
- `PayIDClient`: GET request client, most folks will use this.
- `PayIDAdminClient`: PUT request client, requires admin token
- `PayIDClientTest` / `PayIDAdminClientTest`: Tests for classes

### Context of Change

This change defines a client with a public API that will work with the initial payID service. All methods are unimplemented and throw unimplemented errors.

Note I'm specifically not mentioning this in the CHANGELOG because it's really not ready to be used in any meaningful way and I don't want to call attention to it, yet.

I've also started a subfolder for PayID. I will refactor XRP / ILP to be in similar subfolders when I have a bit more breathing more. 

### Type of Change

## Before / After

A new class is introduced.

## Test Plan

Basic unit tests skeleton included.

### DIFF
--- a/src/PayID/pay-id-admin-client.ts
+++ b/src/PayID/pay-id-admin-client.ts
@@ -19,7 +19,7 @@ import PayIDError from './pay-id-error'
  * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
  */
 export default class PayIDAdminClient {
-  private static putEndpoint = '/put'
+  private static putEndpoint = '/v1/users'
 
   /**
    * Initialize a new PayIDAdminClient.
@@ -35,7 +35,9 @@ export default class PayIDAdminClient {
   ) {}
 
   /**
-   * Update a PayID to XRP address mapping.
+   * Create or update a PayID to XRP address mapping.
+   *
+   * If the given PayID has no XRP address mapping then a new mapping is created. If the given PayID has an existing mapping, then the mapping us updated and the previous value is removed.
    *
    * @note The input address to this method must be in X-Address format.
    *
@@ -43,7 +45,7 @@ export default class PayIDAdminClient {
    * @param xrpAddress The new XRP address to associate with the payID.
    * @returns A boolean indicating success of the operation.
    */
-  public async updateXRPAddressMapping(
+  public async addOrUpdateXRPAddressMapping(
     _payID: string,
     _xrpAddress: string,
   ): Promise<boolean> {
@@ -53,6 +55,23 @@ export default class PayIDAdminClient {
     console.log(`Hitting: ${serviceEndpoint}`)
     console.log(`Using Token: ${this.authorizationToken}`)
 
+    // TODO(keefertaylor): Implement writes from backend service.
+    throw PayIDError.unimplemented
+  }
+
+  /**
+   * Remove a PayID to XRP address mapping.
+   *
+   * @note The input address to this method must be in X-Address format.
+   *
+   * @param payID The payID to update.
+   * @param xrpAddress The XRP address to remove.
+   * @returns A boolean indicating success of the operation.
+   */
+  public async removeXRPAddressMapping(
+    _payID: string,
+    _xrpAddress: string,
+  ): Promise<boolean> {
     // TODO(keefertaylor): Implement writes from backend service.
     throw PayIDError.unimplemented
   }
--- a/src/PayID/pay-id-admin-client.ts
+++ b/src/PayID/pay-id-admin-client.ts
@@ -1,0 +1,59 @@
+import PayIDError from './pay-id-error'
+
+/* eslint-disable @typescript-eslint/require-await */
+// TODO(keefertaylor): Enable lint rules when method is implemented.
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable class-methods-use-this */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable max-classes-per-file */
+
+/**
+ * An admin for Xpring's PayID service.
+ *
+ * Admin clients can update address mappings. To resolve PayIDs to addresses, please use `PayIDClient`.
+ *
+ * Inputs to this class are always in X-Addresses.
+ * @see https://xrpaddress.info/
+ *
+ * @warning This class is experimental and should not be used in production applications.
+ * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
+ */
+export default class PayIDAdminClient {
+  private static putEndpoint = '/put'
+
+  /**
+   * Initialize a new PayIDAdminClient.
+   *
+   * TODO(keefertaylor): Add a remoteURL configuration property when it is used, otherwise the linter is upset about an unused property.
+   *
+   * @param authorizationToken An authorization token.
+   * @param remoteURL The remote URL for the PayID Service.
+   */
+  public constructor(
+    private readonly authorizationToken: string,
+    private readonly serviceURL: string,
+  ) {}
+
+  /**
+   * Update a PayID to XRP address mapping.
+   *
+   * @note The input address to this method must be in X-Address format.
+   *
+   * @param payID The payID to update.
+   * @param xrpAddress The new XRP address to associate with the payID.
+   * @returns A boolean indicating success of the operation.
+   */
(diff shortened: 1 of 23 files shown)


src/PayID/pay-id-client.ts
```diff
@@ -0,0 +1,31 @@
...
+ * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
+ */
+export default class PayIDClient {
+  /**
+   * Retrieve the XRP Address authorized with a PayID.
+   *
+   * @note The returned value will always be in an X-Address format.
+   *
+   * @param payID The payID to resolve for an address.
+   * @returns An XRP address representing the given PayID if one exists, otherwise undefined.
```
hbergren: You may want to add a (optional?) `@param network` here. You can default to `MAINNET` in your code, or you can not pass a `Network` header at all and it will (currently) default to returning the associated Mainnet address.

TBD, @austin-king and I need to decide if `Network` is a mandatory or optional header when requesting a PayID, and whether it's a part of the service or the core protocol. I personally think that an optional header is more user-friendly, but perhaps a mandatory header makes for a simpler API and ensures it doesn't become a hidden feature.

src/PayID/pay-id-error.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Types of errors that originate from PayID.
+ */
+export enum PayIDErrorType {
+  Unimplemented,
+}
+
+/**
+ * Represents errors thrown by PayID components of the Xpring SDK.
+ */
+export default class PayIDError extends Error {
```
hbergren: 👍 You're always diligent about defining custom errors, which is something I need to get better at.

src/PayID/pay-id-admin-client.ts
```diff
@@ -0,0 +1,59 @@
...
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable max-classes-per-file */
+
+/**
+ * An admin for Xpring's PayID service.
+ *
+ * Admin clients can update address mappings. To resolve PayIDs to addresses, please use `PayIDClient`.
+ *
+ * Inputs to this class are always in X-Addresses.
+ * @see https://xrpaddress.info/
```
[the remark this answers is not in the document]
dpurdy: xAddress is our SDK parameters seems right.

src/PayID/pay-id-admin-client.ts
```diff
@@ -0,0 +1,59 @@
...
+  ) {}
+
+  /**
+   * Update a PayID to XRP address mapping.
+   *
+   * @note The input address to this method must be in X-Address format.
+   *
+   * @param payID The payID to update.
+   * @param xrpAddress The new XRP address to associate with the payID.
+   * @returns A boolean indicating success of the operation.
```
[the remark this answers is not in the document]
dpurdy: $doug.purdy.im -> mainnet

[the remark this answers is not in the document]
dpurdy: $test.doug.purdy.im -> testnet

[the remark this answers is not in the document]
dpurdy: at the base and we can add more later.

src/PayID/pay-id-client.ts
dpurdy: Is the network in the URL + Content-Type tuple?

src/PayID/pay-id-admin-client.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Admin clients can update address mappings. To resolve PayIDs to addresses, please use `PayIDClient`.
+ *
+ * Inputs to this class are always in X-Addresses.
+ * @see https://xrpaddress.info/
+ *
+ * @warning This class is experimental and should not be used in production applications.
+ * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
+ */
+export default class PayIDAdminClient {
+  private static putEndpoint = '/put'
```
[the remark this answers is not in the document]
hbergren: An `add`/`POST` creates a new "payment pointer account" in the PayID service.

An `update/PUT` can CRUD addresses associated with that account _or_ the payment pointer associated with that account. So if I change my `xpringID` from `hbergren` to `hansbergren`, that looks like updating my `payment_pointer` using a `PUT` request rather than deleting my old PayID service account and creating a new one with all new mappings.

Does that sound reasonable?

src/PayID/pay-id-client.ts
hbergren: Right now we've been playing with a separate (optional?) HTTP header.

So a request looks like:
```
GET https://xpring.io/hbergren
 Accept: "application/xrp+json"
 Network: "Testnet"
```

Currently if network isn't defined, it defaults to Mainnet.

src/PayID/pay-id-admin-client.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Initialize a new PayIDAdminClient.
+   *
+   * TODO(keefertaylor): Add a remoteURL configuration property when it is used, otherwise the linter is upset about an unused property.
+   *
+   * @param authorizationToken An authorization token.
+   * @param remoteURL The remote URL for the PayID Service.
+   */
+  public constructor(
+    private readonly authorizationToken: string,
+    private readonly serviceURL: string,
```
hbergren: `remoteURL` in the docstring vs `serviceURL` here?

src/PayID/pay-id-admin-client.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Admin clients can create, update, and delete address mappings. To resolve PayIDs to addresses, please use `PayIDClient`.
+ *
+ * Inputs to this class are always in X-Addresses.
+ * @see https://xrpaddress.info/
+ *
+ * @warning This class is experimental and should not be used in production applications.
+ * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
+ */
+export default class PayIDAdminClient {
+  private static putEndpoint = '/v1/users'
```
hbergren: This feels weird to me.

You take a `remoteURL`, implying you support PayID services besides the Xpring PayID service, but you hardcode this endpoint, which may be different for other PayID services. Thoughts?

src/PayID/pay-id-client.ts
```diff
@@ -0,0 +1,28 @@
...
+ * A client for PayID.
+ *
+ * @warning This class is experimental and should not be used in production applications.
+ * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
+ */
+export default class PayIDClient {
+  /**
+   * Retrieve the XRP Address authorized with a PayID.
+   *
+   * @note The returned value will always be in an X-Address format.
```
austin-king: Not X-address + (classic address + dest. tag)?

src/PayID/pay-id-client.ts
austin-king: Deciding on final format with @sappenin later today.

src/PayID/pay-id-admin-client.ts
keefertaylor: Done

src/PayID/pay-id-admin-client.ts
keefertaylor: I'm going to merge this PR and we can talk about this later. 

Semantically I'd assume a payID server would just work with this. Why would we want this to be configurable?

src/PayID/pay-id-client.ts
keefertaylor: Xpring SDK only serves X-addresses. I'm happy to revisit this API with you at another point but it's all the same information. 

Serving redundant information feels confusing to users.

keefertaylor:
@dpurdy 
> Is the network in the URL + Content-Type tuple?

I'm basically going to omit the concept of network right now. When we have it properly defined I'm happy to add it. Getting this PR landed lets us riff on this and start building which is really what I'd like to do. 

happy to open another PR to debate on network semantics

keefertaylor:
In the interest of pushing forward I'm going to merge this since I have 3x approvals. 

Lots of questions still hanging about so folks should just let me know if they feel they aren't sufficiently answered.

PR #216, opened 2020-03-04
(1/n)  Implement enum for TransactionType

keefertaylor:

Provide an enum for transactions.

Reference implementation: https://github.com/xpring-eng/XpringKit/pull/128

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan

N/A

### DIFF
--- a/src/XRP/xrp-transaction-type.ts
+++ b/src/XRP/xrp-transaction-type.ts
@@ -1,0 +1,12 @@
+/**
+ * Types of transactions on the XRP Ledger.
+ *
+ * This is a partial list. Please file an issue if you have a use case that requires additional types.
+ *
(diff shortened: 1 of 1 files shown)


src/XRP/xrp-transaction-type.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Types of transactions on the XRP Ledger.
+ *
+ * This is a partial list. Please file an issue if you have a use case that requires additional types.
```
mayurbhandary: Do we have a publicly available list of types that are on the roadmap?

keefertaylor: Our plan is basically to add things as there is external demand, due to the limited engineering resources we have versus other priorities.

PR #224, opened 2020-03-05
Implement PayID Get

keefertaylor:

Implement `PayIDClient` for GET/read requests. 

Implement unit tests with mocked networking and integration tests which run against the `doug.purdy.im` PayID service. 

### Context of Change

This change is large to show how we will handle these requests holistically. The following components are deferred to minimize PR size / until there is a proper spec:
- `network` parameters
- parsing of the response

Note that I haven't updated docs / changelog since I think this feature is still in flux / not worth mentioning. 

One thing I'm not sure about is whether to call `$keefertaylor.com/keefer` a `PayID` or a `PaymentPointer`. As such, the class names end up somewhat mixed in code. If/When we determine a consistent terminology I can do a global find / replace. 

### Type of Change

## Before / After

PayID client is now useable. 

## Test Plan

New unit / integration tests provided for continuous integration.


.mocharc.js
```diff
@@ -1,13 +1,9 @@
-'use strict';
+'use strict'
```
keefertaylor: This file just got formatted on save; no code changes.

src/PayID/pay-id-admin-client.ts
```diff
@@ -23,11 +23,9 @@ export default class PayIDAdminClient {
 
   /**
    * Initialize a new PayIDAdminClient.
-   *
-   * TODO(keefertaylor): Add a remoteURL configuration property when it is used, otherwise the linter is upset about an unused property.
-   *
+   *   *
```
keefertaylor: This is just minor cleanup from #215

src/PayID/pay-id-client.ts
```diff
@@ -14,6 +9,13 @@ import PayIDError from './pay-id-error'
  * TODO(keefertaylor): Export this class in index.ts when it's ready for external consumption.
  */
 export default class PayIDClient {
+  /** An HTTP client. */
+  private readonly axiosInstance: AxiosInstance
+
+  public constructor() {
+    this.axiosInstance = axios.create({})
+  }
+
```
Stormtv: Do we need an axios instance here? We could have this class only have static methods.

src/PayID/pay-id-client.ts
```diff
@@ -22,7 +24,45 @@ export default class PayIDClient {
...
+    const requestConfig = {
+      headers: {
+        Accept: 'application/xrp+json',
+      },
+    }
+
+    return this.axiosInstance
+      .get(url, requestConfig)
+      .then((result: AxiosResponse) => {
+        // TODO(keefertaylor): properly extract xrp address and tag, format as X-Address.
```
hbergren: I think we may try to "standardize" the XRP format for PayID to be something like:
```
{
  "xAddress": "XVdQPCEqWheLp7JNnWE3HbenVZof8jfEHcbWd7Vr9MVXPWb",
  "classicAddress": "rMFmeAsBdfiQFcdFszwPmsj4GT82EFfLHe",
  "destinationTag": "12345"
}
```

In that case, you wouldn't need to parse the address and tag (although it might still be worth handling that case).

src/PayID/pay-id-client.ts
```diff
@@ -22,7 +24,45 @@ export default class PayIDClient {
...
    */
-  public async xrpAddressForPayID(_payID: string): Promise<string | undefined> {
-    throw PayIDError.unimplemented
+  public async xrpAddressForPayID(payID: string): Promise<string | undefined> {
+    const paymentPointer = PayIDUtils.parsePaymentPointer(payID)
+    if (!paymentPointer) {
+      throw PayIDError.invalidPaymentPointer
+    }
+
+    const url = `https://${paymentPointer.host}${paymentPointer.path}`
```
hbergren: Does `path` here include the leading `/`? Or does `host` have a trailing `/`?

src/PayID/pay-id-client.ts
```diff
@@ -22,7 +24,45 @@ export default class PayIDClient {
...
+          // Otherwise re-throw the error as an unexepected response.
+          throw new PayIDError(
+            PayIDErrorType.UnexpectedResponse,
+            `${error.response.status} ${error.response.statusText}: ${error.response.data}`,
+          )
+        } else {
+          // Generically errors from the server which contained no response (timeouts, etc).
+          throw new PayIDError(PayIDErrorType.Unknown, error.message)
+        }
+      })
```
hbergren: Great error handling, as usual.

src/PayID/pay-id-client.ts
hbergren: Although, `doug.purdy.im` just returns `address: r58...`, so this is in flux too, and if you want to prioritize working with `doug.purdy.im`, that's fine.

https://doug.purdy.im/.well-known/pay

src/PayID/pay-id-client.ts
```diff
@@ -22,7 +24,45 @@ export default class PayIDClient {
...
+    const url = `https://${paymentPointer.host}${paymentPointer.path}`
+    // TODO(keefertaylor): Attach network parameter.
+    // TODO(keefertaylor): Generalize the below to allow requesting in other types.
+    const requestConfig = {
+      headers: {
+        Accept: 'application/xrp+json',
+      },
+    }
+
+    return this.axiosInstance
```
Stormtv: Why promise structure here versus async await?

src/PayID/pay-id-client.ts
keefertaylor: Path has a leading /. 

This is following the conventions set in Quilt (which is the implementation I'm mirroring) that `path` always starts with `/` .

Invariant explicitly defined in Quilt Code: https://github.com/hyperledger/quilt/blob/master/spsp-parent/spsp-core/src/main/java/org/interledger/spsp/PaymentPointer.java#L90

Tests showing this is the case: https://github.com/hyperledger/quilt/blob/master/spsp-parent/spsp-core/src/test/java/org/interledger/spsp/PaymentPointerTest.java

src/PayID/pay-id-client.ts
keefertaylor: I actually would rather use `await` but it threw exceptions which sucked because then I had to type them. Do you have an example of this in action?

src/PayID/pay-id-client.ts
keefertaylor: I assumed there was an overhead in running `axios.create`. I also assumed that when I get a PayID spec that we might have some common headers which run across methods, etc. 

Is this an un-idiomatic use of axios?

tedkalaw: there's a global axios instance you can use if you want, which seems fine here. you can also set configs for it

Stormtv: If there is common headers, common errors we are intercepting, or a common url then keep this but if we are just having one api just use the `axios.get` function and don't require a class initialization

src/PayID/pay-id-client.ts
Stormtv: I think we can type them as AxiosError. I can pull this down and check though

Stormtv: Try catch method of async await
```ts
    try {
      const response = await this.axiosInstance.get<{ address: string }>(
        url,
        requestConfig,
      )
      if (!response) return
      const { address } = response.data
      if (!address) {
        throw new PayIDError(
          PayIDErrorType.UnexpectedResponse,
          'Sucessful response was in an unknown format',
        )
      }
      return address
    } catch (error) {
      // Handle erroneous responses from the server.
      if (error.response) {
        // 404 means no mapping was found.
        if (error.response.status === 404) {
          return undefined
        }

        // Otherwise re-throw the error as an unexepected response.
        throw new PayIDError(
          PayIDErrorType.UnexpectedResponse,
          `${error.response.status} ${error.response.statusText}: ${error.response.data}`,
        )
      } else {
        // Generically errors from the server which contained no response (timeouts, etc).
        throw new PayIDError(PayIDErrorType.Unknown, error.message)
      }
    }
```

or

catch model (saves a level of indentation and reads easier IMO)
```ts
    const response = await this.axiosInstance
      .get<{ address: string }>(url, requestConfig)
      .catch((error) => {
        // Handle erroneous responses from the server.
        if (error.response) {
          // 404 means no mapping was found.
          if (error.response.status === 404) {
            return undefined
          }

          // Otherwise re-throw the error as an unexepected response.
          throw new PayIDError(
            PayIDErrorType.UnexpectedResponse,
            `${error.response.status} ${error.response.statusText}: ${error.response.data}`,
          )
        } else {
          // Generically errors from the server which contained no response (timeouts, etc).
          throw new PayIDError(PayIDErrorType.Unknown, error.message)
        }
      })
    if (!response) return
    const { address } = response.data
    if (!address) {
      throw new PayIDError(
        PayIDErrorType.UnexpectedResponse,
        'Sucessful response was in an unknown format',
      )
    }
    return address
```

keefertaylor: I agree both of these are better. In my original implementation (and both of yours) the throw from the unexpected response format gets caught in the catch, which misbehaves. 

The only way around this I can find is to wrap the whole thing in a promise, which I've done but which is really ugly.

PR #225, opened 2020-03-05
Update `.mocharc.js` to glob subfolders

keefertaylor:

PayID tests are in `test/pay-id-client-test` which isn't automatically picked up by mocha. Update the mocha config to run these tests.

Fix failures in tests which now run.

### Context of Change

The functionality of Xpring SDK is rapidly expanding to XRP, ILP and PayID. I've started sub-foldering source files and tests to disambiguate. PayID is in a new subfolder.

For some reason, the `assert.typeOf` call always returns `Error` rather than a `PayIDError`. I'm not sure why this is. The following asserts about `PayIDErrorCode` being set correctly cover this case though, so I'm deferring for now. Reviewers can let me know if there's a more elegant way to do this. 

### Type of Change

## Before / After

Tests run properly

## Test Plan

CI

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -8,6 +8,7 @@ module.exports = {
   extension: [
     'ts'
   ],
+  spec: 'test/**/*-test.ts',
 
   // Do not look for mocha opts file
   opts: false,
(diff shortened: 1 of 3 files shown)


keefertaylor:
Done: https://app.asana.com/0/1143722888069470/1165282936179569

PR #230, opened 2020-03-07
(2/n) Provide conversion code and testing for `Currency` protocol buffer

amiecorso:
Provide a native swift wrapper for the [Currency protocol ](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41). 

Swift analogue: https://github.com/xpring-eng/XpringKit/pull/126

### Context of Change

### Type of Change

## Before / After
No change

## Test Plan
new CI introduced for unit tests.

### DIFF
--- a/test/protocol-buffer-conversion-test.ts
+++ b/test/protocol-buffer-conversion-test.ts
@@ -10,6 +10,7 @@ describe('Protocol Buffer Conversion', function(): void {
     const currencyProto: Currency = new Currency()
     currencyProto.setCode(currencyCode)
     currencyProto.setName(currencyName)
+    
     // WHEN the protocol buffer is converted to a native Typescript type.
     const currency = new XRPCurrency(currencyProto)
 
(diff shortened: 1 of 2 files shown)


test/protocol-buffer-conversion-test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe('Protocol Buffer Conversion', function(): void {
+  it('Convert Currency protobuf to XRPCurrency object', function(): void {
+    // GIVEN a Currency protocol buffer with a code and a name.
+    const currencyCode: Uint8Array = new Uint8Array([1, 2, 3])
+    const currencyName = 'abc'
+    const currencyProto: Currency = new Currency()
+    currencyProto.setCode(currencyCode)
+    currencyProto.setName(currencyName)
+    // WHEN the protocol buffer is converted to a native Typescript type.
```
keefertaylor: nit: +1 vertical line space

PR #231, opened 2020-03-08
(3/n) Path / PathElement

amiecorso:
Provide a native JS wrapper for the [Path Element protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L227) and [Path protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L237).

Analogue in Swift SDK: https://github.com/xpring-eng/XpringKit/pull/127

### Context of Change

### Type of Change

## Before / After
No change

## Test Plan


amiecorso:
Please consider my TODO note in the code about using non-null assertion... is this best practice?

src/xrp-path-element.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  public issuer: string | null
+
+  // TODO(amiecorso): is it ok to disable non-null assertion operator in these lines? (or perhaps for full file?)
+  // These fields are non-optional in the protobuf, so if for example hasAccount() is true, we would expect
+  // getAccount() to be non-null.
+  public constructor(pathElement: PathElement) {
+    this.account = pathElement.hasAccount()
+      ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        pathElement.getAccount()!.getAddress()
```
keefertaylor: I think you should use optional chaining rather than a non-null assertion. It's syntactic sugar:
```
pathElement.getAccount()?.getAddress()
```

Which says "if account is null, just stop and return null". In practice the above expands to:
```
const account = pathElement.getAccount()
if (!account) {
  return null
}
else account.getAddress()
```

You can do this indefinitely:
```
foo()?.bar()?.abc()?.xyz()
```

src/xrp-path.ts
```diff
@@ -0,0 +1,20 @@
+import { Payment } from './generated/web/org/xrpl/rpc/v1/transaction_pb'
+import XRPPathElement from './xrp-path-element'
+
+type Path = Payment.Path
+type PathElement = Payment.PathElement
```
hbergren: I like the type aliases here!

src/xrp-path-element.ts
```diff
@@ -0,0 +1,34 @@
+import { Payment } from './generated/web/org/xrpl/rpc/v1/transaction_pb'
+import XRPCurrency from './xrp-currency'
+
+type PathElement = Payment.PathElement
+
+/*
+ * A path in the XRP Ledger.
```
hbergren:
```suggestion
 * A path element in the XRP Ledger.
```

hbergren: Also, is this a PathElement? Or a Path Step?
https://xrpl.org/paths.html#path-steps

test/protocol-buffer-conversion-test.ts
```diff
@@ -1,6 +1,27 @@
...
+testAccountAddress.setAddress('r123')
+
+const testAccountAddressIssuer = new AccountAddress()
+testAccountAddressIssuer.setAddress('r456')
+
+// Populate test PathElement
+const testPathElement = new Payment.PathElement()
+testPathElement.setCurrency(testCurrencyProto)
+testPathElement.setAccount(testAccountAddress)
+testPathElement.setIssuer(testAccountAddressIssuer)
```
hbergren: Is this effectively a `beforeAll` that you need to run before your tests?

Does our test library here have a `beforeAll` method? If not, could you add some comments here to clarify that this is setup (if it is) and why this is necessary?

test/protocol-buffer-conversion-test.ts
```diff
@@ -17,4 +38,76 @@ describe('Protocol Buffer Conversion', function(): void {
...
+    const path = new XRPPath(pathProto)
+
+    // THEN there are zero paths in the output.
+    assert.equal(path.pathElements.length, 0)
+  })
+
+  it('Convert Path protobuf with one Path to XRPPath', function(): void {
+    // GIVEN a set of paths with one path.
+    const pathProto = new Payment.Path()
+    pathProto.addElements(testPathElement)
```
hbergren: Can we not do `new Payment.Path(testPathElement)`?

src/xrp-path-element.ts
amiecorso: oh interesting that it's called a Path Step here in the documentation... It's called a PathElement in the rippled protocol buffers, so we'll stick with that for consistency.

amiecorso: But I am going to update the comment/link to the path steps you have here

test/protocol-buffer-conversion-test.ts
amiecorso: the Protobuf constructors don't take arguments - they use getters and setters to interact with all the fields after object construction

test/protocol-buffer-conversion-test.ts
amiecorso: So apparently Mocha does have a before method... and a beforeEach, but these don't seem to put variables in the proper scope.  I've spent a ton of time trying to get them to work and I think it's clearer and simpler to just have some code above the describe method with an additional comment.

hbergren: That's reasonable. I just ran into the variable scoping with Mocha as well. It makes sense, given what JS scoping is, but definitely annoying.

I still managed to use a `before/after` block, because I basically do:
`const app = new Express()` on the outside, so it's in scope, and then I have:
```
before() {
  app.init();
}

after() {
  app.close();
}
```

But keeping everything outside is just as reasonable.

src/xrp-path-element.ts
```diff
@@ -8,27 +8,20 @@ type PathElement = Payment.PathElement
...
  */
 export default class XRPPathElement {
-  public account: string | null
+  public account: string | undefined
 
-  public currency: XRPCurrency | null
+  public currency!: XRPCurrency | undefined
 
-  public issuer: string | null
+  public issuer: string | undefined
```
hbergren: Can these never be `null`?

hbergren: Can these never be `null`?

amiecorso: That's a great question.  I believe they can never be null because they are never assigned null in our code, and the native TypeScript implementations of the Protobuf classes either return a value or undefined with the getters and setters, as you can see here (sry for the black and white)
Screen Shot 2020-03-09 at 5 08 36 PM

src/PayID/pay-id-client.ts
```diff
@@ -22,7 +24,53 @@ export default class PayIDClient {
...
+    if (!paymentPointer) {
+      throw PayIDError.invalidPaymentPointer
+    }
+
+    const url = `https://${paymentPointer.host}${paymentPointer.path}`
+    // TODO(keefertaylor): Attach network parameter.
+    // TODO(keefertaylor): Generalize the below to allow requesting in other types.
+    const requestConfig = {
+      headers: {
+        Accept: 'application/xrp+json',
```
austin-king: Fine for now, but I believe we're going to update this to reflect the formats in this doc: https://docs.google.com/document/d/1_Jl12pWqfVGmpq5hhOXClOJXuMck0dEaF_NvgvYkWSg/edit like 'xrp-mainnet' and 'xrp-testnet'

PR #233, opened 2020-03-08
Default to rippled's protocol buffer implementation

keefertaylor:

Change the default behavior of the SDK to connect to rippled. 

### Context of Change

We're using these protocol buffers in the Xpring Wallet and they'll be released officially in rippled 1.5. Xpring believes this implementation to be more robust and safer than the legacy API. We provide stable, well administered nodes which are hosted in our documentation.  

### Type of Change

## Before / After

Users will be defaulted to a new implementation which will provide them a better and safer experience. The only change they will need to make is to change the URL they point to. 

## Test Plan

CI

## Future Tasks

This is the first step in a multi-phase turn down of legacy:
- [x] Spin up new nodes
- [x] Dogfood rippled protocol buffers in wallet system
- [x] Default to rippled implementation
- [ ] Remove legacy protocol buffers / service from Xpring-SDK
- [ ] Turn down legacy service


keefertaylor:
@stephengu  ping

PR #234, opened 2020-03-08
[Documentation] Add Endpoints for Xpring-JS in the browser

keefertaylor:

Update Xpring-JS documentation 

### Context of Change

In browser based applications, Xpring-JS needs to hit an envoy proxy instead of a node. Xpring hosts these nodes. Add these nodes to the documentation. 

### Type of Change

[x] Documentation updates

## Before / After

N/A

## Test Plan

N/A


keefertaylor:
I added an asana task for  updating the Japanese readme: https://app.asana.com/0/1143722888069470/1165758769580083

Also added in some plain text so it isn't wrong, although obviously this would be better if it had accompanying text which made sense.

PR #238, opened 2020-03-10
[WIP] (5/N) Convert CurrencyAmount protocol buffer

amiecorso:
Provide a native swift wrapper for the [CurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L10).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/130

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


amiecorso:
This is failing due to the same big-integer type system issue I'm trying to figure out here: https://github.com/xpring-eng/Xpring-JS/pull/237

PR #241, opened 2020-03-10
Rebase ILP Client classes to ILP/

nkramer44:
- Moved all ILP classes to `src/ILP/`
- Moved all ILP test classes to `test/ILP`

### Context of Change
We should split XRP and ILP clients into their own directories to better organize classes/files.  Before I add wrapper objects for protobufs in Xpring-JS, I thought it would be helpful to rebase the ILP classes.

### Type of Change

## Before / After
Before: All ILP classes and XRP classes in `src`
After: All ILP classes in `src/ILP`

## Test Plan
CI


keefertaylor:
You're my hero.

test/fakes/fake-ilp-network-client.ts
```diff
@@ -1,4 +1,4 @@
-import { IlpNetworkClient } from '../../src/ilp-network-client'
+import { IlpNetworkClient } from '../../src/ILP/ilp-network-client'
```
keefertaylor: I'd move this file to `tests/ILP/fakes/` instead of `tests/fakes`

PR #245, opened 2020-03-11
Remove `getTransactionStatus` method.

keefertaylor:

This method was deprecated in the last release. Remove it in favor of the method named `getPaymentStatus` which is more descriptive.

### Context of Change

Generally we should provide a longer deprecation cycle but the next release is going to:
- Remove `XpringClient`
- Change the default to use rippled's proto implementation

So I think this is a minor cut compared to those two changes and we should just rip the bandage off at once. 

### Type of Change

## Before / After

No deprecated methods. 

## Test Plan

CI


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+- `getTransactionStatus` is deprecated. Please use `getPaymentStatus` instead.
```
hbergren: was deprecated in `1.6.0`, and has now been removed?

I get this is under a `Removed` header, but `deprecated` implies to me that it's still there.

test/xrp-client-integration-test.ts
```diff
@@ -70,7 +70,7 @@ describe('Xpring JS XRPClient Integration Tests', function(): void {
   > {
     this.timeout(timeoutMs)
 
-    const transactionStatus = await legacyXRPClientWeb.getTransactionStatus(
+    const transactionStatus = await legacyXRPClientWeb.getPaymentStatus(
```
hbergren: So even though we're doing breaking changes, we still have the `legacyXRPClient` around?

Should cutting that be part of `2.0`?

keefertaylor: This will get cut when we move to rippled 1.5. 

I'm waiting for two things to happen:
1) rippled 1.5 is actually released (it's currently just an RC)
2) we've had the backend of the wallet on rippled in prod for a week or two with no issues

CHANGELOG.md
keefertaylor: Done. I think i just copy and pasted the wrong header.

PR #246, opened 2020-03-11
Protobuf Wrapper Objects (1/3)

nkramer44:
Asana task: https://app.asana.com/0/1143722888069470/1165556580466065

- Added data classes that will get converted to/from protobuf objects to avoid exposing protobufs in the public API

### Context of Change
This PR is the JS counterpart to https://github.com/xpring-eng/xpring4j/pull/125.

Essentially, we shouldn't be exposing generated protobuf objects in the public ILP API.  This introduces some data objects that mirror those generated objects, but are nicer to look at and can be exposed in the public API.

### Type of Change

## Before / After
Before: `IlpClient` consumes and produces protobuf objects
After: `IlpClient` still consumes and produces protobuf, but in part 2 of this PR, it will consume and produce these new wrapper objects.

## Test Plan
The existing test suite will suffice for this PR, but will need some refactoring (ie changing request/response types) once these objects are in use.

## Future Tasks
- Change `IlpClient` signatures and refactor tests to use wrapped objects


keefertaylor:
\+ Amie - who is doing this for XRP protobufs
\- Dino - who is busy

src/ILP/model/account-balance.ts
```diff
@@ -0,0 +1,92 @@
...
+import { GetBalanceResponse } from '../../generated/node/ilp/get_balance_response_pb'
+
+/**
+ * Response object for requests to get an account's balance
+ */
+export class AccountBalance {
+  /**
+   * The accountId for this account balance.
+   */
+  readonly accountId: string
```
keefertaylor: Please use `ID` instead of `Id` (here + comments)

src/ILP/model/account-balance.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.prepaidAmount = options.prepaidAmount
+    this.netBalance = options.clearingBalance.add(options.prepaidAmount)
+  }
+
+  /**
+   * Constructs an AccountBalance from a GetBalanceResponse
+   *
+   * @param getBalanceResponse a GetBalanceResponse (protobuf object) whose field values will be used
+   *                           to construct an AccountBalance
+   * @return an AccountBalance with its fields set via the analogous protobuf fields.
```
keefertaylor: hilariously pedantic nit: be consistent if you use periods at the end of your one liner documentation :)

src/ILP/model/payment-request.ts
```diff
@@ -0,0 +1,57 @@
...
+   * network.
+   *
+   * @see "https://github.com/interledger/rfcs/blob/master/0026-payment-pointers/0026-payment-pointers.md"
+   */
+  readonly destinationPaymentPointer: string
+
+  /**
+   * @return The accountID of the sender.
+   */
+  readonly senderAccountId: string
```
keefertaylor: s/Id/ID

src/ILP/model/account-balance.ts
nkramer44: I'm OK with this, but the Hermes protobufs use `accountId`, so there will be some inconsistency there.  We also use `accountId` across the whole java ILP stack, so it will be inconsistent with that.

theotherian: I'm against doing this as it represents a relatively drastic refactoring to a number of applications and libraries in our domain.

src/ILP/model/account-balance.ts
nkramer44: lol.... I have a bunch of cleanup in documentation that I have to do.  Lots of inconsistencies and wrong documentation.  Captured in https://app.asana.com/0/1154633712169520/1166011127952369.

Going to get this done in a separate PR

src/ILP/model/account-balance.ts
nhartner: Is there a desire to follow naming conventions of the language, or use the same naming standard across all languages. In Javascript, `accountId` is more common. For example it's `document.getElementById()` not `document.getElementByID()`

PR #248, opened 2020-03-11
(2/n) Provide conversion code and testing for `Currency` protocol buffer

amiecorso:
Provide a native TypeScript wrapper for the [Currency protocol](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41) .

Swift analogue: https://github.com/xpring-eng/XpringKit/pull/126

Note: Re-doing [this PR](https://github.com/xpring-eng/Xpring-JS/pull/230), which was approved but only merged into a non-master branch.  All current (x/n) PRs for building up protobuf wrapper types will be done anew. 

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan
new CI introduced for unit tests.


src/XRP/xrp-currency.ts
```diff
@@ -0,0 +1,20 @@
+import { Currency } from '../generated/web/org/xrpl/rpc/v1/amount_pb'
+
+/**
+ * An issued currency on the XRP LEdger
+ * - SeeAlso: https://xrpl.org/currency-formats.html#currency-codes
```
keefertaylor: Use @see in JSDoc: https://jsdoc.app/tags-see.html

src/XRP/xrp-currency.ts
```diff
@@ -0,0 +1,20 @@
...
+export default class XRPCurrency {
+  public name: string
+
+  public code: string | Uint8Array
+
+  public static from(currency: Currency): XRPCurrency | undefined {
+    return new XRPCurrency(currency)
+  }
+
+  private constructor(currency: Currency) {
```
keefertaylor: Typescript lets you create readonly iVars + bindings in one constructor.  So you can pretty drastically reduce boiler plate code here:
```
export class Foo {
  private constructor(public readonly accountID: string, public readonly assetCode: string, ...)

  static from(getBalanceResponse: GetBalanceResponse) {
      return Foo(getBalanceResponse.getAccountID(), ...)
   }
}
```

https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties

amiecorso:
awesome, this is prettier!

keefertaylor:
I think you need to push updates here.

PR #249, opened 2020-03-11
(3/n) Path / PathElement

amiecorso:
Provide a native TypeScript wrapper for the [Path Element protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L227) and [Path protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L237).

Analogue in Swift SDK: [xpring-eng/XpringKit#127](https://github.com/xpring-eng/XpringKit/pull/127)

Re-doing [this PR](https://github.com/xpring-eng/Xpring-JS/pull/231), which I will close.

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


src/XRP/xrp-currency.ts
```diff
@@ -2,19 +2,15 @@ import { Currency } from '../generated/web/org/xrpl/rpc/v1/amount_pb'
...
+    return new XRPCurrency(currency.getName(), currency.getCode())
   }
 
-  private constructor(currency: Currency) {
-    this.name = currency.getName()
-    this.code = currency.getCode()
-  }
+  private constructor(
+    readonly name: string,
+    readonly code: string | Uint8Array,
```
keefertaylor: We should document this type. If it's a string, then it's a base64 encoded string. If it's an array, it's an array of bytes. 

You can use a more specific helper:
```
# Gets string | Uint8Array
currency.getCode() 

# Gets a uint8array
currency.getCode_asU8()

# Gets a base64 encoded string
currency.getCode_asB64()
```

I found this by inspecting the generated code on disk: 

Screen Shot 2020-03-16 at 4 30 44 PM

src/XRP/xrp-currency.ts
```diff
@@ -2,19 +2,15 @@ import { Currency } from '../generated/web/org/xrpl/rpc/v1/amount_pb'
...
  * An issued currency on the XRP LEdger
- * - SeeAlso: https://xrpl.org/currency-formats.html#currency-codes
+ * @see: https://xrpl.org/currency-formats.html#currency-codes
  */
 export default class XRPCurrency {
-  public name: string
-
-  public code: string | Uint8Array
-
   public static from(currency: Currency): XRPCurrency | undefined {
```
keefertaylor: I think you can drop `undefined` here. This will always be able to convert.

src/XRP/xrp-path-element.ts
```diff
@@ -0,0 +1,29 @@
...
+/*
+ * A path step in an XRP Ledger Path.
+ * @see: https://xrpl.org/paths.html#path-steps
+ */
+export default class XRPPathElement {
+  public static from(pathElement: PathElement): XRPPathElement | undefined {
+    const account = pathElement.getAccount()?.getAddress()
+    const currency = pathElement.getCurrency()
+    let xrpCurrency
+    if (currency) {
```
keefertaylor: if you drop `undefined` from the definition above, this if statement is removeable.

src/XRP/xrp-path.ts
```diff
@@ -0,0 +1,19 @@
...
+import XRPPathElement from './xrp-path-element'
+
+type Path = Payment.Path
+
+/*
+ * A path in the XRP Ledger.
+ * @see: https://xrpl.org/paths.html
+ */
+export default class XRPPath {
+  public static from(path: Path): XRPPath | undefined {
```
keefertaylor: You can also drop `undefined` from the initializer here.

test/protocol-buffer-conversion-test.ts
```diff
@@ -15,7 +37,82 @@ describe('Protocol Buffer Conversion', function(): void {
...
+    // WHEN the protocol buffer is converted to a native TypeScript type.
+    const pathElement = XRPPathElement.from(pathElementProto)
+
+    // THEN the currency converted as expected.
+    assert.isUndefined(pathElement?.account)
+    assert.isUndefined(pathElement?.currency)
+    assert.isUndefined(pathElement?.issuer)
+  })
+
+  // MARK: - Org_Xrpl_Rpc_V1_Transaction.Path
```
keefertaylor: nit: JS doesn't recognize `MARK: -` AFAIK. 

Maybe one of the more knowledgeable JS devs can weigh in if there's a better option.

amiecorso: Ah yes, this is deleted in later PRs where I just label each test section with the pb conversion being tested, such as // Path.

src/XRP/xrp-currency.ts
amiecorso: great, took care of this on (2/n) and merged here

src/XRP/xrp-path-element.ts
amiecorso: I can't actually get rid of this if statement, because `currency` here comes from the protobuf and has type `Currency | undefined` based on how the protos are written. (it's not an XRPCurrency)

src/XRP/xrp-path-element.ts
```diff
@@ -0,0 +1,29 @@
...
+ * A path step in an XRP Ledger Path.
+ * @see: https://xrpl.org/paths.html#path-steps
+ */
+export default class XRPPathElement {
+  public static from(pathElement: PathElement): XRPPathElement {
+    const account = pathElement.getAccount()?.getAddress()
+    const currency = pathElement.getCurrency()
+    let xrpCurrency
+    if (currency) {
+      xrpCurrency = XRPCurrency.from(currency)
```
hbergren: Is it worth making the `from()` return `undefined` if given `undefined`? @keefertaylor

test/protocol-buffer-conversion-test.ts
```diff
@@ -1,6 +1,28 @@
...
+// Set up global variables for use in test cases
+const testCurrencyProto: Currency = new Currency()
+testCurrencyProto.setCode(new Uint8Array([1, 2, 3]))
+testCurrencyProto.setName('currencyName')
+
+const testAccountAddress = new AccountAddress()
+testAccountAddress.setAddress('r123')
+
+const testAccountAddressIssuer = new AccountAddress()
+testAccountAddressIssuer.setAddress('r456')
```
hbergren: Should you be able to set the address to an invalid address?

test/protocol-buffer-conversion-test.ts
```diff
@@ -18,4 +40,78 @@ describe('Protocol Buffer Conversion', function(): void {
...
+
+    // WHEN the protocol buffer is converted to a native TypeScript type.
+    const path = XRPPath.from(pathProto)
+
+    // THEN there is one path in the output.
+    assert.equal(path?.pathElements.length, 1)
+  })
+
+  it('Convert Path protobuf with many Paths to XRPPath', function(): void {
+    // GIVEN a set of paths with one path.
```
hbergren: I think this comment is wrong. We have 3 path elements, right?

src/XRP/xrp-path-element.ts
keefertaylor: I, personally, feel that makes APIs less specific. That is to say `string|undefined` (optional string) is harder to reason about than `string` (string). 

While it would provide neat syntactical sugar at this level, I'm not sure the cost outweighs the benefit. Happy to hear if you feel strongly.

test/protocol-buffer-conversion-test.ts
keefertaylor: This is a protocol buffer so it is just typed as 'string' as an underlying data type and thus any string is valid. This is just test data so we're just making our lives easy and abusing this privilege.

src/XRP/xrp-path-element.ts
amiecorso: Ok, I'm going to leave this as it is for now, and if we decide to change later we can do them all at once.

hbergren: Ok, I do have one suggestion.

Instead of:
```js
let xrpCurrency
if (currency) {
  xrpCurrency = XRPCurrency.from(currency)
} else {
  xrpCurrency = undefined
}
```

I think you can write that as:
```js
const xrpCurrency = currency && XRPCurrency.from(currency)
```

This works because your `if` statement is only asserting that `currency` is truthy. I forget what exactly this usage of the `&&` operator is called, but basically, it should function the same as your `if/else` above.

amiecorso: nice!! cleaner.  I think there are a few places where this type of logic appears while converting from protos to native objects - I'll take another pass on the rest of this PR chain as well.

PR #250, opened 2020-03-12
(4/n) Convert IssuedCurrency protocol buffer

amiecorso:
Provide a native TypeScript wrapper for the [IssuedCurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L28).

Analogue in Swift: [xpring-eng/XpringKit#129](https://github.com/xpring-eng/XpringKit/pull/129)
Re-doing [this PR](https://github.com/xpring-eng/Xpring-JS/pull/237), which I will close.

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


test/protocol-buffer-conversion-test.ts
```diff
@@ -25,6 +30,7 @@ testPathElement.setAccount(testAccountAddress)
 testPathElement.setIssuer(testAccountAddressIssuer)
 
 describe('Protocol Buffer Conversion', function(): void {
+  // Currency
```
keefertaylor: Nit: I'd put on vertical space above and below these headers so they're more visible.

Feel free to ignore if you've seen this style elsewhere in Xpring's JS Codebase

src/XRP/xrp-issued-currency.ts
```diff
@@ -0,0 +1,36 @@
...
+export default class XRPIssuedCurrency {
+  public static from(
+    issuedCurrency: IssuedCurrencyAmount,
+  ): XRPIssuedCurrency | undefined {
+    const currency = issuedCurrency.getCurrency()
+    let xrpCurrency
+    if (currency) {
+      xrpCurrency = XRPCurrency.from(currency)
+    }
+    let value
```
keefertaylor: nit: I'd probably use some vertical spacing to break up this function after you set each variable.

PR #252, opened 2020-03-12
(5/N) Convert CurrencyAmount protocol buffer

amiecorso:
Provide a native TypeScript wrapper for the[ CurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L10).

Analogue in Swift: [xpring-eng/XpringKit#130](https://github.com/xpring-eng/XpringKit/pull/130)
Re-doing [this PR](https://github.com/xpring-eng/Xpring-JS/pull/238), which I will close.

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


test/protocol-buffer-conversion-test.ts
```diff
@@ -166,4 +182,49 @@ describe('Protocol Buffer Conversion', function(): void {
     // THEN the result is undefined
     assert.isUndefined(issuedCurrency)
   })
+
+  // CurrencyAmount tests
```
keefertaylor: +1 vertical space after this.

test/protocol-buffer-conversion-test.ts
```diff
@@ -29,6 +33,18 @@ testPathElement.setCurrency(testCurrencyProto)
 testPathElement.setAccount(testAccountAddress)
 testPathElement.setIssuer(testAccountAddressIssuer)
 
+// test IssuedCurrencyAmount
+const testIssuedCurrency = new IssuedCurrencyAmount()
+testIssuedCurrency.setCurrency(testCurrencyProto)
+testIssuedCurrency.setIssuer(testAccountAddress)
+testIssuedCurrency.setValue('100')
+
+// test Invalid IssuedCurrencyAmount
```
keefertaylor: refactor line 154 to use this new object rather than creating it twice.

test/protocol-buffer-conversion-test.ts
```diff
@@ -166,4 +182,49 @@ describe('Protocol Buffer Conversion', function(): void {
...
+
+  it('Convert CurrencyAmount with Issued Currency', function(): void {
+    // GIVEN a currency amount protocol buffer with an issued currency amount.
+    const currencyAmountProto = new CurrencyAmount()
+    currencyAmountProto.setIssuedCurrencyAmount(testIssuedCurrency)
+
+    // WHEN the protocol buffer is converted to a native TypeScript type.
+    const currencyAmount = XRPCurrencyAmount.from(currencyAmountProto)
+
+    // THEN the result has an issued currency set and no amount.
```
keefertaylor: This 'THEN' could probably be phrased as "no xrp amount" or "no drops amount" which is slightly more specific.

keefertaylor: Apologies since I'm guessing you're copying me :)

test/protocol-buffer-conversion-test.ts
```diff
@@ -166,4 +182,49 @@ describe('Protocol Buffer Conversion', function(): void {
...
+    assert.deepEqual(
+      currencyAmount?.issuedCurrency,
+      XRPIssuedCurrency.from(testIssuedCurrency),
+    )
+    assert.isUndefined(currencyAmount?.drops)
+  })
+
```
keefertaylor: Can you add a test case for converting with the bad issued amount proto you made on line 42? Analagous to https://github.com/xpring-eng/XpringKit/pull/130/files#diff-9307f8a129e7063d9086218b91281a2bR154

test/protocol-buffer-conversion-test.ts
amiecorso: haha, I am! but I agree nonetheless

test/protocol-buffer-conversion-test.ts
amiecorso: nice catch

test/protocol-buffer-conversion-test.ts
amiecorso: I think this test is already in here, but I just refactored it to use the testInvalidIssuedCurrency instead of re-creating it.

PR #253, opened 2020-03-12
Protobuf Wrapper Objects (2+3/3)

nkramer44:
- Change `getBalance` and `sendPayment` signatures to use proto wrapper objects

### Context of Change
Javascript counterpart to https://github.com/xpring-eng/xpring4j/pull/126 and https://github.com/xpring-eng/xpring4j/pull/127.

Putting both of these changes into a singular PR for the sake of velocity, and because this basically mirrors what we did in Xpring4j

### Type of Change

## Before / After
Before: `getBalance` and `sendPayment` consumed and returned proto objects
After: `getBalance` and `sendPayment` consume and return wrapper objects

## Test Plan
Updated `DefaultIlpClientTest` to check each field of responses, which expands the scope of those tests to the conversion process between protos and non-protos


src/ILP/default-ilp-client.ts
```diff
@@ -37,45 +37,42 @@ class DefaultIlpClient implements IlpClientDecorator {
   public constructor(private readonly networkClient: IlpNetworkClient) {}
 
   /**
-   * Retrieve the balance for the given address.
+   * Retrieve the balance for the given accountId.
```
keefertaylor: nit: Id to ID to be consistent with the rest of the SDK code. 

I'm guessing by the amount of time I make this comment that the ILP team favors "Id" so maybe we can change this at some point in the future.

PR #254, opened 2020-03-12
(6/n) Convert Payment Protocol Buffers

amiecorso:
Provide a native TypeScript wrapper for the [Payment protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L224).
Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/131

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+    const paymentAmountValue = payment.getAmount()?.getValue()
+    if (paymentAmountValue) {
+      amount = XRPCurrencyAmount.from(paymentAmountValue)
+      if (!amount) {
+        return undefined // amount is required
+      }
+    } else {
+      return undefined
+    }
+    const destination = payment
```
keefertaylor: I'd put one vertical space in between each variable assignment

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (paymentDeliverMinValue) {
+        deliverMin = XRPCurrencyAmount.from(paymentDeliverMinValue)
+        if (!deliverMin) {
+          return undefined
+        }
+      }
+    } else {
+      deliverMin = undefined
+    }
+    const invoiceID = payment.getInvoiceId()?.getValue()
```
keefertaylor: `getValue_u8()`

test/protocol-buffer-conversion-test.ts
```diff
@@ -227,4 +235,239 @@ describe('Protocol Buffer Conversion', function(): void {
     // THEN the result is empty
     assert.isUndefined(currencyAmount)
   })
+
+  // Payment
```
keefertaylor: +1 vertical space below this

test/protocol-buffer-conversion-test.ts
```diff
@@ -227,4 +235,239 @@ describe('Protocol Buffer Conversion', function(): void {
...
+  it('Convert Payment with all fields set', function(): void {
+    // GIVEN a pyament protocol buffer with all fields set.
```
keefertaylor: s/pyament/payment

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+import XRPCurrencyAmount from './xrp-currency-amount'
+import XRPPath from './xrp-path'
+
+// TODO(amiecorso): Modify this object to use X-Address format.
+export default class XRPPayment {
+  public static from(payment: Payment): XRPPayment | undefined {
+    let amount
+    const paymentAmountValue = payment.getAmount()?.getValue()
+    if (paymentAmountValue) {
+      amount = XRPCurrencyAmount.from(paymentAmountValue)
```
hbergren: What is `XRPCurrencyAmount.from(undefined)`? Would that let us collapse some of this conditional stuff?

Then this could look like:
```ts
    const paymentAmountValue = payment.getAmount()?.getValue()
    const amount = XRPCurrencyAmount.from(paymentAmountValue)
    if (!amount) {
      return undefined // amount is required
    }
```

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    const destination = payment
+      .getDestination()
+      ?.getValue()
+      ?.getAddress()
+    if (!destination) {
+      return undefined // destination is required
+    }
+    const destinationTag = payment.getDestinationTag()?.getValue()
+    // If the deliverMin field is set, it must be able to be transformed into a XRPCurrencyAmount.
```
hbergren: nit: blank line to indicate that `destinationTag` is not related to `deliverMin`

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+      const paymentDeliverMinValue = payment.getDeliverMin()?.getValue()
+      if (paymentDeliverMinValue) {
+        deliverMin = XRPCurrencyAmount.from(paymentDeliverMinValue)
+        if (!deliverMin) {
+          return undefined
+        }
+      }
+    } else {
+      deliverMin = undefined
+    }
```
hbergren: Again, what is `XRPCurrencyAmount.from(undefined)`?

Maybe this could look like this:
```ts
    // If the deliverMin field is set, it must be able to be transformed into a XRPCurrencyAmount.
    let deliverMin = undefined
    if (payment.hasDeliverMin()) {
      const paymentDeliverMinValue = payment.getDeliverMin()?.getValue()
      deliverMin = XRPCurrencyAmount.from(paymentDeliverMinValue)
      if (!deliverMin) {
        return undefined
      }
    }
```

We could even make this shorter (depending on return values):
```ts
    // If the deliverMin field is set, it must be able to be transformed into a XRPCurrencyAmount.
    const paymentDeliverMinValue = payment.getDeliverMin()?.getValue()
    const deliverMin = XRPCurrencyAmount.from(paymentDeliverMinValue)
    if (payment.hasDeliverMin() && !deliverMin) {
      return undefined
    }
```

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,79 @@
...
+      const paymentSendMaxValue = payment.getSendMax()?.getValue()
+      if (paymentSendMaxValue) {
+        sendMax = XRPCurrencyAmount.from(paymentSendMaxValue)
+        if (!sendMax) {
+          return undefined
+        }
+      } else {
+        sendMax = undefined
+      }
+    }
```
hbergren: Same sort of process for `sendMax` as `deliverMin`.

src/XRP/xrp-payment.ts
amiecorso: I think the hiccup with this approach is that `paymentAmountValue` potentially has a value of `undefined`, based on how the protobufs are defined... so in order to feed it to `XRPCurrencyAmount.from(paymentAmountValue)`, the signature of `.from(...)` would have to include `undefined` as a type option for its parameter.   Do we think it's better to give all the `.from()` methods the option to take an `undefined` parameter and then simply immediately return `undefined`?

src/XRP/xrp-payment.ts
amiecorso: same question as above... Keefer do you want to weigh in on whether we should update the `.from` initializers so that we can feed them `undefined` values and use their return to make things more succinct?

src/XRP/xrp-payment.ts
hbergren: @keefertaylor ☝️

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,84 @@
...
+    )
+  }
+
+  private constructor(
+    readonly amount?: XRPCurrencyAmount,
+    readonly destination?: string,
+    readonly destinationTag?: number,
+    readonly deliverMin?: XRPCurrencyAmount,
+    readonly invoiceID?: Uint8Array,
+    readonly paths?: (XRPPath | undefined)[],
```
keefertaylor: What is this type? An array of XRPPath or undefined? 

I would think that this is a `Array<XRPPath>|undefined`. Is this just alternative syntax? If so, I've generally been using `Array<Foo>` syntax in Xpring-JS, would you mind changing for consistency.

amiecorso: So the issue with this is here: https://github.com/xpring-eng/Xpring-JS/blob/fa65158ff80240c92ac862702b52cd90fdb90736/src/XRP/xrp-payment.ts#L47  where we use the map function to convert an array of protobuf paths to an array of XRPPaths... since `XRPPath.from(...)` can return `undefined`, then it's possible that the array contains a mixture of `XRPPath` and `undefined` values.  (In practice, it never should, but the type system is obviously behaving this way).  I'm not sure how else I would go about that except perhaps to not use the map function?  And instead get rid of a value if it's `undefined` instead of ever allowing it to appear in the array?  The possibility of `undefined` is worse than coronavirus.

amiecorso: and to answer your question specifically, I don't think this is alternative syntax, I think that what I have `(XRPPath | undefined)[]` means an array of mixed values? v.s. a single value that's not an array at all but just undefined.

src/XRP/xrp-payment.ts
hbergren: Like my comment in https://github.com/xpring-eng/Xpring-JS/pull/249#discussion_r395053932, I think you could do a:
```

Comment (hbergren) [reply src/XRP/xrp-payment.ts]: Like my comment in https://github.com/xpring-eng/Xpring-JS/pull/249#discussion_r395053932

I think you could do a:
```js
const amount = paymentAmountValue && XRPCurrencyAmount.from(paymentAmountValue)
```

I do see you have an additional check on if the `XRPCurrencyAmount.from()` fails, even with a truthy `paymentAmountValue`, so you could move that:
```
if (!amount) return undefined
```
to the outside.

src/XRP/xrp-payment.ts
hbergren: See comment above about `amount`.

src/XRP/xrp-payment.ts
hbergren: See comment above for `amount`.

src/XRP/xrp-payment.ts
hbergren: I think it is alternative syntax. You should be able to write:
```
Array<XRPPath | undefined>
```

right?

hbergren: > I'm not sure how else I would go about that except perhaps to not use the map function? And instead get rid of a value if it's undefined instead of ever allowing it to appear in the array? The possibility of undefined is worse than coronavirus.

I'm not sure where the relevant `.map()` call is, but you could do a `.filter()` afterwards to filter the array to truthy values.

test/protocol-buffer-conversion-test.ts
```diff
@@ -220,4 +228,240 @@ describe('Protocol Buffer Conversion', function(): void {
...
+    sendMaxCurrencyAmountProto.setIssuedCurrencyAmount(
+      testInvalidIssuedCurrency,
+    )
+    const sendMaxProto = new SendMax()
+    sendMaxProto.setValue(sendMaxCurrencyAmountProto)
+
+    const paymentProto = new Payment()
+    paymentProto.setAmount(amountProto)
+    paymentProto.setDestination(destinationProto)
+    paymentProto.setSendMax(sendMaxProto)
```
hbergren: This is a lot of boilerplate for the `GIVEN`s in these tests. Is it worth writing a helper method to take care of these? (Your call).

src/XRP/xrp-payment.ts
amiecorso: So I think it's alternative syntax for ```Array<XRPPath | undefined>``` but not for ```Array<XRPPath> | undefined```, which is what I thought Keefer was asking.  Should I re-write as ```Array<XRPPath | undefined>```  stylistically?

amiecorso: oh, yes

test/protocol-buffer-conversion-test.ts
amiecorso: I think the reason to wait on a helper method like this is because each test case is creating a slightly different Payment proto in terms of which fields are actually set and what their values are, so for the moment I think the overhead of that helper wouldn't be worth it.

amiecorso: With too many more test cases though... it probably would because it is such a pain in the ass to set all of these things up without the option for chaining.

src/XRP/xrp-payment.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// TODO(amiecorso): Modify this object to use X-Address format.
+export default class XRPPayment {
+  public static from(payment: Payment): XRPPayment | undefined {
+    const paymentAmountValue = payment.getAmount()?.getValue()
+    const amount =
+      paymentAmountValue && XRPCurrencyAmount.from(paymentAmountValue)
+    if (!amount) {
+      return undefined // amount is required
+    }
```
hbergren: @keefertaylor might hate this, but you _can_ do
```
if (!amount) return undefined
```

Some people really hate the 1-line `if return` style though, so this isn't necessarily a recommendation. Not sure what the general code style of this repo is.

amiecorso: the linter is making me do it like this so I'll just leave it consistent

PR #255, opened 2020-03-12
(7/9) Convert Memo Protocol Buffers

amiecorso:
Provide a native TypeScript wrapper for the [Memo protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L80).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/136

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


src/XRP/xrp-memo.ts
```diff
@@ -0,0 +1,20 @@
+import { Memo } from '../generated/web/org/xrpl/rpc/v1/transaction_pb'
+
+/*
+ * Represents a memo on the XRPLedger.
+ * @see: https://xrpl.org/transaction-common-fields.html#memos-field
+ */
+export default class XRPMemo {
+  public static from(memo: Memo): XRPMemo | undefined {
+    const data = memo.getMemoData()?.getValue()
```
keefertaylor: Use `getValue_u8()` to get byte arrays so we don't have polymorphic types.

src/XRP/xrp-memo.ts
```diff
@@ -0,0 +1,20 @@
...
+    const format = memo.getMemoFormat()?.getValue()
+    const type = memo.getMemoType()?.getValue()
+    return new XRPMemo(data, format, type)
+  }
+
+  private constructor(
+    readonly data?: Uint8Array | string,
```
keefertaylor: just `Uint8Array` after the change above.

PR #256, opened 2020-03-12
(8/9) Convert Signer Protocol Buffers

amiecorso:
Provide a native TypeScript wrapper for the [Signer protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L90).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/137

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


src/XRP/xrp-signer.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/*
+ * Represents a signer of a transaction on the XRP Ledger.
+ * @see: https://xrpl.org/transaction-common-fields.html#signers-field
+ */
+export default class XRPSigner {
+  public static from(signer: Signer): XRPSigner | undefined {
+    const account = signer
+      .getAccount()
+      ?.getValue()
```
stephengu: really small nitpick: can you make the ? go on getAccount()'s line ie getAccount()? since that is the thing it belongs to.

amiecorso: that is a good nit, but it actually produces a parsing error when I write it that way, and the linter is automatically forcing me to write it as is.

PR #261, opened 2020-03-13
(9/9) Convert Transaction Protocol Buffers

amiecorso:
Provide a native TypeScript wrapper for the [Transaction protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/135

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


amiecorso:
If y'all could look explicitly at my use of the RippledFlags class in this case... It seemed that the functionality was a little different from how it works in the Swift case and I wasn't sure I was using it right.

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,109 @@
+import { Transaction } from '../generated/web/org/xrpl/rpc/v1/transaction_pb'
+import RippledFlags from '../rippled-flags'
+import XRPSigner from './xrp-signer'
+import XRPTransactionType from './xrp-transaction-type'
+import XRPPayment from './xrp-payment'
+import XRPMemo from './xrp-memo'
+
+/* A transaction on the XRP Ledger.
```
keefertaylor: nit: bad comment formatting:

``` 
/**
 * A transaction ...
```

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,109 @@
...
+    const account = transaction
+      .getAccount()
+      ?.getValue()
+      ?.getAddress()
+    const fee = transaction.getFee()?.getDrops()
+    const sequence = transaction.getSequence()?.getValue()
+    const signingPublicKey = transaction.getSigningPublicKey()?.getValue()
+    const transactionSignature = transaction
+      .getTransactionSignature()
+      ?.getValue()
```
keefertaylor: use `getValue_u8()`

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const accountTransactionID = transaction
+      .getAccountTransactionId()
+      ?.getValue()
+    const rawValue = transaction.getFlags()?.getValue()
+    let flags
+    if (rawValue) {
+      flags = RippledFlags.checkFlag(rawValue, RippledFlags.TF_PARTIAL_PAYMENT)
```
keefertaylor: This method will check if the partial payment flag is set in the `rawValue`. 
```
 RippledFlags.checkFlag(rawValue, RippledFlags.TF_PARTIAL_PAYMENT)
```

You really want to pass along this value, which you could just do as:
```
const flags = transaction.getFlags()?.getValue()
```

In swift, we make flags a struct because Swift has some built in magic for working with bit sets. In JS it seems the best way is to just work with raw numbers.

test/protocol-buffer-conversion-test.ts
```diff
@@ -555,4 +564,286 @@ describe('Protocol Buffer Conversion', function(): void {
     assert.deepEqual(signer?.signingPublicKey, signingPublicKey)
     assert.deepEqual(signer?.transactionSignature, transactionSignature)
   })
+
+  // Transaction
```
keefertaylor: +1 vertical space below.

test/protocol-buffer-conversion-test.ts
```diff
@@ -555,4 +564,286 @@ describe('Protocol Buffer Conversion', function(): void {
...
+    signerSigningPublicKeyProto.setValue(signingPublicKey)
+
+    const signerTransactionSignatureProto = new TransactionSignature()
+    signerTransactionSignatureProto.setValue(transactionSignature)
+    const signerProto = new Signer()
+    signerProto.setAccount(signerAccountProto)
+    signerProto.setSigningPublicKey(signerSigningPublicKeyProto)
+    signerProto.setTransactionSignature(signerTransactionSignatureProto)
+
+    const sourceTag = 6
```
keefertaylor: I'd move this down to where you build the source tag on line 645

src/XRP/xrp-transaction.ts
amiecorso: I did this in two other places that previously had polymorphic fields - now it's all byte arrays

test/protocol-buffer-conversion-test.ts
amiecorso: moved it up to where all the other real values are defined, for consistency

src/XRP/xrp-transaction.ts
amiecorso: hmm, I don't totally understand this second part. are you saying that ```const flags = transaction.getFlags()?.getValue()``` is sufficient and I don't need the ```if (rawValue) { 
flags = RippledFlags.checkFlag(....)``` ?

keefertaylor: correct

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const transactionSignature = transaction
+      .getTransactionSignature()
+      ?.getValue_asU8()
+
+    const accountTransactionID = transaction
+      .getAccountTransactionId()
+      ?.getValue_asU8()
+
+    const flags = transaction.getFlags()?.getValue()
```
dino-rodriguez: Really like the use of the `?` operator chaining here to get a value | undefined. Nice!

dino-rodriguez: Wondering if these return types should be more explicit?

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const lastLedgerSequence = transaction.getLastLedgerSequence()?.getValue()
+
+    let memos
```
dino-rodriguez: can assign to `undefined` here instead of having the else. Up to you

src/XRP/xrp-transaction.ts
```diff
@@ -0,0 +1,110 @@
...
+      case Transaction.TransactionDataCase.PAYMENT: {
+        const payment = transaction.getPayment()
+        if (payment) {
+          paymentFields = XRPPayment.from(payment)
+          type = Transaction.TransactionDataCase.PAYMENT
+          if (!paymentFields) {
+            return undefined
+          }
+        } else {
+          return undefined
```
dino-rodriguez: Nice of an either monad (undefined | tx)! Wondering if you need any more information than `undefined` at the layer above? Or is it sufficient information for all of these to return `undefined`?

src/XRP/xrp-transaction.ts
amiecorso: What would that look like?  type annotations in the variables?  This is number 9/9 in a chain of classes like this so for the moment I'm going to leave as-is, since the types of these variables are essentially documented in the constructor.

src/XRP/xrp-transaction.ts
amiecorso: nice, shortened

src/XRP/xrp-transaction.ts
amiecorso: it's sufficient for these to return undefined.  the .from() initializers in all of these related classes should return undefined given certain problems with conversation from the protobufs.  However I did manage to shorten this logic quite a bit

amiecorso: (I've learned some stuff from PRs 1 - 8 and had forgotten to incorporate here.)

src/XRP/xrp-transaction.ts
dino-rodriguez: Yeah, sounds good

src/XRP/xrp-transaction.ts
dino-rodriguez: Great!

PR #263, opened 2020-03-13
Migrate to using Swagger for public PayID API

keefertaylor:

Migrate to using Swagger for PayID.

### Context of Change

[Swagger](swagger.io) is a code generator that runs on an OpenAPI spec to build network clients. I believe this is a more scalable way to write PayID in a rapidly iterating API. 

I have verified that this change works similarly in Java. Additionally, it *can* work in Swift if we change the definition around, as Swift doesn't allow changing hosts in generated code. I am hopeful that in the future the quality of swagger generated code in Swift will improve, that Xpring can submit a patch to Swagger, or that we can hand roll swift code. 

This doesn't touch headers. It seems custom headers can be added later. I'm working on this as an area of research. 

The CI build system is pretty janky. I filed https://app.asana.com/0/1143722888069470/1166944618521393 to track as I'm extremely frustrated after spending days on trying to get this configured.

### Type of Change

## Before / After

Code is less crufty, easier to maintain. 

## Test Plan

CI continues to pass for tests.

## Future Tasks
- Move Open API spec somewhere consumeable by multiple libs (probably it's own repo)
- Investigate if we can autogenerate typescript bindings


test/mocha.opts
```diff
@@ -1,3 +0,0 @@
---require ./node_modules/ts-node/register
```
keefertaylor: Note this file shouldn't be here as configuring mocha via a `mocha.opts` file is deprecated. Instead a `.mocha.rc.js` file in the root does this.

keefertaylor:
/- Stephen \+ Dino for folks with more context.

src/PayID/pay-id-client.ts
```diff
@@ -24,53 +18,40 @@ export default class PayIDClient {
...
-    } catch (error) {
-      // Handle erroneous responses from the server.
-      if (error.response) {
-        // 404 means no mapping was found.
-        if (error.response.status === 404) {
-          return undefined
-        }
+    const client = new ApiClient()
+    client.basePath = `https://${paymentPointer.host}`
+    const apiInstance = new DefaultApi(client)
```
hbergren: Looks like you tore out headers, how do they work?

pay-id-api-spec/pay-id.v1.yml
```diff
@@ -0,0 +1,33 @@
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  address:
+                    type: string
```
hbergren: Might be worth pulling this schema out into a named schema?

Here's a named schema with examples I wrote for the Github Payout spec:

```
components:
  schemas:
    Repository:
      type: object
      required:
        - name
      properties:
        id:
          type: string
          format: uuid
          example: acd48109-dbf5-4436-8f84-56d8686533a4
        organization_id:
          type: string
          format: uuid
          example: d290f1ee-6c54-4b01-90e6-d701748f0851
        name:
          type: string
          example: 'xpring-common-js'
```

hbergren: Also, the return payload also has a `network` property, which will match the Accept header.

So:
`application/xrpl-testnet+json` =>
```
{
  "network": "xrpl-testnet",
  "address": "T-Address",
}
```

keefertaylor:
@hbergren:

> So are we just trying to document the public GET API for XRP specifically?

I think that we're just going to create a generic API. We can add headers in code specifically.

pay-id-api-spec/pay-id.v1.yml
keefertaylor: > Might be worth pulling this schema out into a named schema?

Done.

> Also, the return payload also has a network property, which will match the Accept header.

Noted. My goal here is to keep things compatible to make sure this change is a zero logic change. A future change will implement network / move to the proper PayID service.

keefertaylor:
@dino-rodriguez 

> I wonder if generating code server side for PayID (if that is the intention) would be more constraining or liberating in trying to define PayID

(Mostly copy / pasted from Slack)
The PayID GET API was pretty bad to begin with. If you look at the diff on this PR you'll see how much simpler the code now is. 

Agree that right now the open api spec isn't particularly descriptive.

I think swagger is worth it for the following reasons:
- The API spec will become descriptive and useful when Travel Rule / Invoices land
- Travel Rule / Invoices complicate the API which makes it worthwhile to use code generation
- Generally expect public GET requests to become more complicated as additional methods / currencies are supported
- Similarly, I think the Admin API is complicated enough that it's worth it to code generate and it will be straightforward to write a descriptive open API spec, so this work is going to get done regardless.

pay-id-api-spec/pay-id.v1.yml
```diff
@@ -0,0 +1,33 @@
...
+        name: path
+        in: path
+        required: true
+    get:
+      tags: []
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
```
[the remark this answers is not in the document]
keefertaylor: I've removed this header and replaced with a TODO to consider what we're doing here.

scripts/pay-id.v1.yml
```diff
@@ -0,0 +1,38 @@
...
+  schemas:
+    Mapping:
+      title: Mapping
+      type: object
+      properties:
+        address:
+          type: string
+      required:
+        - address
+swagger: "2.0"
```
hbergren: What is the difference between `scripts/pay-id.v1.yml` and `pay-id-api-spec/pay-id.v1.yml`?

src/PayID/pay-id-client.ts
```diff
@@ -24,53 +18,41 @@ export default class PayIDClient {
...
-    if (!address) {
-      throw new PayIDError(
-        PayIDErrorType.UnexpectedResponse,
-        'Sucessful response was in an unknown format',
-      )
-    }
-    return address
+    return new Promise((resolve, reject) => {
+      // Ignore unused var in generated code.
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
hbergren: Add this to your ESLint config file?
https://github.com/xpring-eng/xpring-wallet-system/blob/master/system/.eslintrc.js#L73

src/PayID/pay-id-client.ts
```diff
@@ -24,53 +18,41 @@ export default class PayIDClient {
...
+        if (error) {
+          if (error.status === 404) {
+            // Not Found
+            resolve(undefined)
+          } else {
+            const message = `${error.status}: ${error.response.text}`
+            reject(new PayIDError(PayIDErrorType.UnexpectedResponse, message))
+          }
+        } else if (data?.address) {
+          resolve(data.address)
```
hbergren: What does this `resolve(data.address)` call do?

src/PayID/pay-id-client.ts
```diff
@@ -1,5 +1,6 @@
-import axios, { AxiosInstance } from 'axios'
```
hbergren: How do we make HTTP calls now? Is that generated from ApiClient?

scripts/pay-id.v1.yml
```diff
@@ -0,0 +1,38 @@
+info:
+  title: test
+  version: '1.0'
+servers:
+  - url: 'http://localhost:3000'
```
hbergren: This seems a little odd.

For public GET APIs, the "payment pointer server" is whatever the payment pointer is. So `$example.com/hansbergren` would just be a GET request to `https://example.com/hansbergren`.

The Private API has a server, but that's not described here.

scripts/pay-id.v1.yml
keefertaylor: Removed the former, which was a test to try to debug another alternative build strategy.

src/PayID/pay-id-client.ts
keefertaylor: This resolves the promise with the XRP address that came back. Something else you're looking for?

src/PayID/pay-id-client.ts
keefertaylor: Yup. The API client codegenerates the API called.

scripts/pay-id.v1.yml
keefertaylor: I deleted this file. If you look at the right file, you'll see server isn't defined since there is no set endpoint.

src/PayID/pay-id-client.ts
hbergren: No that seems fine. Based on the current public API spec: https://docs.google.com/document/d/1Xc3dgjFhG_XYu_slMQrWQZhJeNzb2hMzhvA-Gw_N9Ug/edit#

This might be `data?.addressDetails?.address` for XRP addresses.

src/PayID/pay-id-client.ts
keefertaylor: Done.

pay-id-api-spec/pay-id.v1.yml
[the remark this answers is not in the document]
hbergren: This response header still seems to exist. Thoughts?

scripts/regenerate_pay_id_api.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Cleaning Up...."
+
+# Copy Source files
+cp -r $GENERATED_SOURCES_FOLDER/* $DESTINATION_FOLDER
+# Remove everythinge else.
+rm -rf scripts/$TMP_SWAGGER_DIR
+
+echo "Done Cleaning Up"
+echo "All Done!"
```
hbergren: Great job writing this build script. Ridiculous that Swagger needs all this.

pay-id-api-spec/pay-id.v1.yml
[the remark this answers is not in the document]
dino-rodriguez: The PayID server may also respond with a 503 explicitly/500 if some service is down.

Or perhaps as most recently mentioned, a 415 in the case of an unsupported MIME type.

Thoughts on how to handle these?

keefertaylor:
> The PayID server may also respond with a 503 explicitly/500 if some service is down.
> Or perhaps as most recently mentioned, a 415 in the case of an unsupported MIME type.
> Thoughts on how to handle these?

I've typed these in the API spec. Internally, I think they should just get mapped to `PayIDErrorType.Unknown`:
- We don't let users send arbitrary mime types so if we get back a 415 then we've messed up in an unknown way
- Server errors are by default unknown - we don't really know what went wrong.

PR #272, opened 2020-03-17
Exports PayIDClient in index.ts

tedkalaw:
This exports PayIDClient and the associated error+type so that we can use it elsewhere.

### Context of Change
We need to export `PayIDClient` so we can use it

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`PayIDClient`, `PayIDError`, and `PayIDErrorType` can all be used by dependencies of `Xpring-JS`

## Test Plan
tslint


src/PayID/pay-id-client.ts
```diff
@@ -74,3 +73,5 @@ export default class PayIDClient {
     return address
   }
 }
+
+export { PayIDErrorType, PayIDError }
```
keefertaylor: These are exported from `pay-id-error.ts`. Why export them from here? is this a common pattern? 

in index.ts you pull them directly from the error file?

tedkalaw: this was a mistake; fixed

PR #274, opened 2020-03-17
Include request headers in PayID

keefertaylor:

Include request headers in PayID.

### Context of Change

This change includes request headers for the network. It exposes an additional enum for the network to hit.  This enum is defaulted to testnet so that this isn't a breaking change and because Testnet feels like a sane default. 

Note that to prove compatibility, we move integration tests from hitting `doug.purdy.im` (which runs a non-compliant PayID service) to dev.payid.xpring.money (a WIP complaint service). 

### Type of Change

## Before / After

Users can hit a reference PayID service

## Test Plan

New integration tests continue to pass on CI.


pay-id-api-spec/pay-id.v1.yml
```diff
@@ -11,12 +11,19 @@ paths:
...
+      #   - application/xrpl-mainnet+json
+      # produces:
+      #   - application/xrpl-mainnet+json
       tags: []
       responses:
         '200':
           description: OK
           content:
-            application/json:
+            application/xrpl-mainnet+json:
```
dino-rodriguez: what about devnet?

src/PayID/pay-id-client.ts
```diff
@@ -30,28 +44,54 @@ export default class PayIDClient {
     // Swagger generates the '/' in the URL by default and the payment pointer's 'path' is prefixed by a '/'. Strip off the leading '/'.
     const path = paymentPointer.path.substring(1)
 
-    // TODO(keefertaylor): Append headers here.
     const client = new ApiClient()
     client.basePath = `https://${paymentPointer.host}`
-    const apiInstance = new DefaultApi(client)
+
+    // Accept only the given netowrk in response.
```
dino-rodriguez: nit: network

src/PayID/pay-id-client.ts
```diff
@@ -30,28 +44,54 @@ export default class PayIDClient {
...
+        accepts,
+        returnType,
+        (error, data, _response) => {
+          if (error) {
+            if (error.status === 404) {
+              // Not Found
+              resolve(undefined)
+            } else {
+              const message = `${error.status}: ${error.response.text}`
+              reject(new PayIDError(PayIDErrorType.UnexpectedResponse, message))
```
dino-rodriguez: Whoa. That's a lot of properties!

pay-id-api-spec/pay-id.v1.yml
keefertaylor: Added.

src/PayID/pay-id-client.ts
keefertaylor: Assume you're talking about the `callApi` method? 

Yeah. The parsing in Swagger wants to use `accept` headers as an array of content types the client may accept (and the server picks one and returns it). This isn't really analogous to the PayID protocol, where we're asking for a specific thing (an address of network / asset) rather than an array of things.

Swagger wants us to declare content types up front and then ranks them in their declared order.  I'm not really sure if Swagger is abusing the semantics of the accept header or if PayID is, but

pay-id-api-spec/pay-id.v1.yml
hbergren: Maybe add `application/xrpl+json` as well (although it's shorthand for `mainnet`.

src/PayID/pay-id-client.ts
```diff
@@ -29,28 +42,54 @@ export default class PayIDClient {
...
+      const postBody = null
+      const pathParams = {
+        path,
+      }
+      const queryParams = {}
+      const headerParams = {}
+      const formParams = {}
+      const authNames = []
+      const contentTypes = []
+      const returnType = Mapping
```
hbergren: Swagger is bad for this problem space haha.

Worth moving this ridiculousness to its own file?

src/PayID/pay-id-client.ts
hbergren: > Yeah. The parsing in Swagger wants to use accept headers as an array of content types the client may accept

That is valid. As Doug pointed out (and I learned today), it's fine to pass an Accept header for `btc+json` and respond with `Content-Type = xrpl+json`. The web is just like that. (Although I don't think that's a great implementation).

It's also reasonable to pass an array of `Accept` headers that you would accept, in order of preference. For example, maybe you have XRP, ETH, and BTC wallets, and you would want to pay someone in that order based on transaction costs. They don't have a XRP wallet, but they have an ETH address, so they respond with ETH.

This is better than responding with `eth+json` when you ONLY asked for `xrpl+json`, because at least `ETH` was in the array of Accept headers you passed.

src/PayID/pay-id-client.ts
keefertaylor: At the moment I was going to just leave all the nastiness encapsulated here. When we implement more complex interactions, I was going to see what refactors we could make.

keefertaylor:
> That is valid. As Doug pointed out (and I learned today), it's fine to pass an Accept header for btc+json and respond with Content-Type = xrpl+json. The web is just like that. (Although I don't think that's a great implementation).

Yup. Swagger actually conforms to this.

> It's also reasonable to pass an array of Accept headers that you would accept, in order of preference. For example, maybe you have XRP, ETH, and BTC wallets, and you would want to pay someone in that order based on transaction costs. They don't have a XRP wallet, but they have an ETH address, so they respond with ETH.

Yup. The problem is that Swagger implicitly orders them and send them all if you use the generated wrapper class. So if it's in the Swagger definition, then our client will take it. That's described here:
https://swagger.io/docs/specification/describing-parameters/

> This is better than responding with eth+json when you ONLY asked for xrpl+json, because at least ETH was in the array of Accept headers you passed.

Agree.

keefertaylor:
> Maybe add application/xrpl+json as well (although it's shorthand for mainnet.

Done

pay-id-api-spec/pay-id.v1.yml
keefertaylor: I walked this back since we removed from spec.

PR #283, opened 2020-03-18
Refactor PayIDClient to throw errors when not found.

keefertaylor:

Refactor PayIDClient to throw an error if the mapping was not found. 

### Context of Change

In conversations earlier today, it became apparent that users of this method should be sure that the payment pointer they are passing is valid. If they pass an invalid payment pointer, then we will throw an error. 

This means that the return type for the method can move from an optional string to a string. This simplifies upstream logic for us.

Refactor tests.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

API changes. This isn't released though so it's going to be a breaking change. 

## Test Plan

Updated tests continue to pass on CI.


pay-id-api-spec/pay-id.v1.yml
```diff
@@ -16,7 +16,16 @@ paths:
         '200':
           description: OK
           content:
-            application/json:
+            application/xrpl+json:
```
hbergren: You can cut this header.

src/PayID/pay-id-client.ts
```diff
@@ -29,28 +42,54 @@ export default class PayIDClient {
...
+      const queryParams = {}
+      const headerParams = {}
+      const formParams = {}
+      const authNames = []
+      const contentTypes = []
+      const returnType = Mapping
+      client.callApi(
+        '/{path}',
+        'GET',
+        pathParams,
```
dino-rodriguez: is there an option for passing these parameters as an object? Wondering if they all need to be defined explicitly as constants above since they are only used as parameters

keefertaylor: There is not, AFAICT. I think a wrapper class might help, but leaving for now to better judge later.

src/PayID/pay-id-error.ts
```diff
@@ -2,10 +2,11 @@
...
 export enum PayIDErrorType {
-  Unimplemented = 0,
-  InvalidPaymentPointer = 1,
-  Unknown = 2,
-  UnexpectedResponse = 3,
+  InvalidPaymentPointer,
+  MappingNotFound,
+  UnexpectedResponse,
+  Unimplemented,
+  Unknown,
```
hbergren: Why did you cut the explicit numbers?

PR #290, opened 2020-03-19
Upgrade to new PayID format

keefertaylor:

PayID protocol implemented breaking changes. Upgrade.

### Context of Change

Upgrade to breaking changes in PayID. 

Note that `addressDetailsType` is actually a discriminator field for the `addressDetails` oneof. I simplified and just hardcoded to CryptoAddressDetails for a few reasons:
- Binds to the "make this work for XRP and all else will follow" ethos
- Tests are currently broken; this unblocks us. 
- The change to support ACH is additive, non-breaking and requires additional logic / testing.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

PayID works with latest spec.

## Test Plan

CI

### DIFF
--- a/test/PayID/pay-id-client-test.ts
+++ b/test/PayID/pay-id-client-test.ts
@@ -40,7 +40,10 @@ describe('Pay ID Client', function(): void {
     nock('https://xpring.money')
       .get('/georgewashington')
       .reply(200, {
-        address: 'X7cBcY4bdTTzk3LHmrKAK6GyrirkXfLHGFxzke5zTmYMfw4',
+        addressDetailsType: 'crypto',
+        addressDetails: {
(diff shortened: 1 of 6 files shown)


test/PayID/pay-id-client-test.ts
```diff
@@ -40,7 +40,10 @@ describe('Pay ID Client', function(): void {
     nock('https://xpring.money')
       .get('/georgewashington')
       .reply(200, {
-        address: 'X7cBcY4bdTTzk3LHmrKAK6GyrirkXfLHGFxzke5zTmYMfw4',
+        addressDetailsType: 'crypto',
```
dino-rodriguez: Think this should be `CryptoAddressDetails` according to @austin-king (went through it on my PR today)

keefertaylor: Done

PR #293, opened 2020-03-20
Implements paymentHistory (1/2)

amiecorso:
Implements `paymentHistory` method in `DefaultXRPClient` and related classes.  Does not yet expose this method in `XRPClient` public API.

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/151

### Context of Change
This PR implements `paymentHistory` and adds additional unit test cases for the logic as well as beginning the process of centralizing fake protobuf and native objects used in testing.

Note this method still does not exist in the public API (will put in separate PR) and thus changelogs etc are not updated and integration tests haven't been written.

### Type of Change

## Test Plan
CircleCI

## Future Tasks

Expose `paymentHistory` in public API, like [this one](https://github.com/xpring-eng/XpringKit/pull/148/files) in Swift SDK


src/default-xrp-client.ts
```diff
@@ -311,6 +316,56 @@ class DefaultXRPClient implements XRPClientDecorator {
       return false
     }
   }
+
+  /**
+   * Return the history of payments for the given account.
+   * Note: This method only works for payment type transactions, see: https://xrpl.org/payment.html
```
keefertaylor: One blank line after the method summary

src/default-xrp-client.ts
```diff
@@ -311,6 +316,56 @@ class DefaultXRPClient implements XRPClientDecorator {
...
+    )
+
+    const getTransactionResponses = transactionHistory.getTransactionsList()
+
+    // Filter transactions to payments only and convert them to XRPTransactions.
+    // If a payment transaction fails conversion, throw an error.
+    const payments: Array<XRPTransaction> = []
+    // eslint-disable-next-line no-restricted-syntax
+    // eslint-disable-next-line no-underscore-dangle
+    for (const _getTransactionResponse of getTransactionResponses) {
```
keefertaylor: don't prefix this with `_` since you use it on 353

src/default-xrp-client.ts
```diff
@@ -311,6 +316,56 @@ class DefaultXRPClient implements XRPClientDecorator {
...
+          const xrpTransaction = XRPTransaction.from(transaction)
+          if (!xrpTransaction) {
+            throw new Error(XRPClientErrorMessages.paymentConversionFailure)
+          } else {
+            payments.push(xrpTransaction)
+          }
+          break
+        }
+        default:
+      }
```
keefertaylor: put a comment here like:
```
default:
   // Intentionally do nothing, non-payment type transactions are ignored.
```

That way future programmers know this was intentional  and you didn't get distracted halfway through a PR.

src/default-xrp-client.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable no-restricted-syntax */
```
keefertaylor: what is the restricted syntax?

src/xrp-client-decorator.ts
```diff
@@ -63,4 +64,15 @@ export interface XRPClientDecorator {
    * @returns A boolean if the account is on the blockchain.
    */
   accountExists(address: string): Promise<boolean>
+
+  /**
+   * Return the history of payments for the given account.
+   * Note: This method only works for payment type transactions, see: https://xrpl.org/payment.html
```
keefertaylor: +1 vertical line after method summary

test/fakes/fake-xrp-protobufs.ts
```diff
@@ -0,0 +1,193 @@
+import {
```
keefertaylor: This is so great.

test/default-xrp-client-test.ts
```diff
@@ -461,4 +469,130 @@ describe('Default Xpring Client', function(): void {
...
+
+  it('Payment History - non-payment transactions', async function(): Promise<
+    void
+  > {
+    // GIVEN an XRPClient client which will return a transaction history which contains non-payment transactions
+
+    // Generate expected transactions from the default response, which only contains payments.
+    const nonPaymentTransactionResponse = new GetTransactionResponse()
+    nonPaymentTransactionResponse.setTransaction(testCheckCashTransaction)
+    const heteregeneousTransactionHistory = testGetAccountTransactionHistoryResponse // here, homogeneous w/ PAYMENT txns
```
keefertaylor: I find these comments. a bit ard to follow. I would just write "// transaction history which contains payment and non-payment transactions." as a one liner above the `const hetergenousTransactionHistory`. 

I think inline is harder to read. YMMV and if you feel strongly you can leave it this way.

test/default-xrp-client-test.ts
```diff
@@ -461,4 +469,130 @@ describe('Default Xpring Client', function(): void {
...
+      heteroHistoryNetworkResponses,
+    )
+
+    const xrpClient = new DefaultXRPClient(heteroHistoryNetworkClient)
+
+    // WHEN the transactionHistory is requested.
+    const transactionHistory = await xrpClient.paymentHistory(testAddress)
+
+    // THEN the returned transactions are conversions of the inputs with non-payment transactions filtered.
+    const expectedTransactionHistory = [
```
keefertaylor: Your input was `testGetAccountTransactionHistoryResponse` so I think a small improvement here could be to loop through that and convert the payments.

test/default-xrp-client-test.ts
```diff
@@ -461,4 +469,130 @@ describe('Default Xpring Client', function(): void {
...
+    // payment transaction array?  Or should I just make my own directly like this: ?
+    //
+    //        const expectedPaymentHistory = [XRPTransaction.from(testPaymentTransaction), XRPTransaction.from(testPaymentTransaction)]
+    // It leaves less room for error in the logic, but it no longer automatically reflects the composition of
+    // the testGetAccountTransactionHistory object
+    const testTransactionResponseList = testGetAccountTransactionHistoryResponse.getTransactionsList()
+    const expectedPaymentHistory: Array<XRPTransaction> = []
+    // eslint-disable-next-line no-restricted-syntax
+    // eslint-disable-next-line no-underscore-dangle
+    for (const _getTransactionResponse of testTransactionResponseList) {
```
keefertaylor: don't underscore here you use this on 491

test/default-xrp-client-test.ts
```diff
@@ -461,4 +469,130 @@ describe('Default Xpring Client', function(): void {
...
   })
+
+  it('Payment History - successful response', async function(): Promise<void> {
+    // GIVEN a DefaultXRPClient.
+    const xrpClient = new DefaultXRPClient(fakeSucceedingNetworkClient)
+
+    // WHEN the payment history for an address is requested.
+    const paymentHistory = await xrpClient.paymentHistory(testAddress)
+
+    // TODO(amiecorso): should I use the same logic that's in DefaultXRPClient to reconstruct the expected
```
keefertaylor: I would use the same logic like you're doing. It's test code so you could probably have dropped the switch and just force unwrapped BUT you already have it so I would leave it.

src/default-xrp-client.ts
amiecorso: It was keeping me from using a for loop instead of array iteration via a map or something.  but I honestly vote for the for loop here.  Hans showed me how I could use flatmap but it was more confusing to look at and not any more performant

amiecorso: (he suggested I just disable the linter for this case)

test/fakes/fake-xrp-protobufs.ts
amiecorso: I think a refactor I might do is to have all the protobuf conversion tests reference objects from this file

test/default-xrp-client-test.ts
amiecorso: my feelings are nonexistent

test/default-xrp-client-test.ts
amiecorso: So I did update this... but that logic for looping through the transactionHistory proto and converting it into an array of XRPTransactions happens in several places here (the `paymentHistory` implementation, and two of the test cases for default-xrp-client), which feels bad.   If I were to factor this into its own method, where would you think that method should belong?  Or should I not, since we don't care about repeating ourselves as much if it's just test code?

test/default-xrp-client-test.ts
amiecorso: I guess this is related to my long comment above about whether this logic belongs in its own method

test/fakes/fake-xrp-protobufs.ts
keefertaylor: Feel free but we can also just do it over time. You could also add this as a bootcamp in Asana if you think it's a good candidate.

src/default-xrp-client.ts
keefertaylor: Makes sense. Maps / flat maps are hard until you've looked at them long enough and I agree it's not exactly self explanatory so I'm happy to use more explicit syntax. 

BTW, I'd generally assume that nearly everything we do in this library is going to be performance bound by network latency so I wouldn't worry too much about performance of actual code unless you're doing something really egregious. Even then we  can always refactor later on.

src/default-xrp-client.ts
```diff
@@ -44,6 +46,9 @@ export class XRPClientErrorMessages {
   public static readonly xAddressRequired =
     'Please use the X-Address format. See: https://xrpaddress.info/.'
   /* eslint-enable @typescript-eslint/indent */
+
+  public static readonly paymentConversionFailure =
+    'Could not convert payment transaction: (transaction). Please file a bug at https://github.com/xpring-eng/xpringkit'
```
keefertaylor: fix link: xpring-js vs xpringkit. 

Also maybe link directly to issues:
https://github.com/xpring-eng/Xpring-JS/issues

test/default-xrp-client-test.ts
keefertaylor: So an alternative here which is really simple is that the constructor for `XRPTransaction` will return undefined if the given input is not a payment. We use a switch in the actual library because we want to catch the case where:
- The transaction is a payment
- Conversion fails

And and throw an error rather than returning an incomplete set of data (which someone may then use to make incorrect choices based on incomplete data). 

You could really write:
```javascript
var xrpTxs = []
for (var i = 0; i < transactions.length; i++) {
  const tx = transactions[i]
  const xrpTX = XRPTransaction.from(tx)
  if (!xrpTx) {
      continue
  }
  xtpTxs.push(xrpTx)
}
```

Your way also works. 

With regards to where to put this code, I wouldn't worry about deduping test code with the actual code base (there's no obvious place to put this logic without making it public which feels like an over optimization). If you're repeating yourself in the tests then you could just write a helper in  this file, or if you need to share it across multiple files you could make a `XRPTestUtils` class in `test/XRP/Helpers`

test/fakes/fake-xrp-protobufs.ts
amiecorso: That's a good idea - I'll do this.

src/legacy/legacy-default-xrp-client.ts
```diff
@@ -22,6 +26,8 @@ export class LegacyXRPClientErrorMessages {
 
   public static readonly xAddressRequired =
     'Please use the X-Address format. See: https://xrpaddress.info/.'
+
```
dino-rodriguez: Seems like an enum would make sense here as well (dealing with a list of constants)

test/default-xrp-client-test.ts
```diff
@@ -461,4 +468,135 @@ describe('Default Xpring Client', function(): void {
...
+
+  it('Payment History - successful response', async function(): Promise<void> {
+    // GIVEN a DefaultXRPClient.
+    const xrpClient = new DefaultXRPClient(fakeSucceedingNetworkClient)
+
+    // WHEN the payment history for an address is requested.
+    const paymentHistory = await xrpClient.paymentHistory(testAddress)
+
+    const expectedPaymentHistory: Array<XRPTransaction> = []
+    for (const getTransactionResponse of testGetAccountTransactionHistoryResponse.getTransactionsList()) {
```
dino-rodriguez: If this is the same loop as above, it could be useful to refactor it into a helper function that can be reused.

src/legacy/legacy-default-xrp-client.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable class-methods-use-this */
+/* eslint-disable @typescript-eslint/require-await */
```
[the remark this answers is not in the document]
amiecorso: good Q.  these are here for the paymentHistory method in the legacy client.  Since the legacy client is no longer being supported, this method will not be implemented here and instead just throws an unimplemented error (but it has to be here to conform to the interface).  I actually moved the disable `no unused vars` down to be right before this method, and got rid of the async keyword entirely (and the corresponding linter disabling) because I realized it's not required.  good catch!

src/legacy/legacy-default-xrp-client.ts
amiecorso: It probably would, but since this is legacy we're not putting any effort into refactoring it.

test/default-xrp-client-test.ts
amiecorso: yeah... related to the comments above between me and keefer.  This lends enough impetus so I just went ahead and refactored it and created a new XRPTestUtils class.  This'll help inspire us to keep things well maintained as the tests continue to expand.

dino-rodriguez: Got it!

PR #328, opened 2020-03-26
Improves accountExists method in XRPClient

amiecorso:
`accountExists` method in `XRPClient` was treating any exception raised as a "false" conclusion (account does not exist). This PR updates the method to conclude false only if a `grpc.status.NOT_FOUND` error code is in the exception. Other codes will cause the exception to be re-thrown. Unit and integration tests also updated to test ability to distinguish error codes.

This PR was originally written here: https://github.com/xpring-eng/Xpring-JS/pull/186 and like a proper newb I merged it into a non-master branch.  Redoing now.

### Type of Change

## Before / After
Before: accountExists could have returned misleading false-negatives.
After: will propagate exceptions that it's not qualified to interpret.

## Test Plan
CI


src/XRP/default-xrp-client.ts
```diff
@@ -295,7 +296,10 @@ class DefaultXRPClient implements XRPClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if (e.code && e.code === grpc.status.NOT_FOUND) {
```
hbergren:
```suggestion
      if (e?.code === grpc.status.NOT_FOUND) {
```
?

src/XRP/default-xrp-client.ts
```diff
@@ -295,7 +296,10 @@ class DefaultXRPClient implements XRPClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
```
hbergren: Do we use `catch (e)` throughout `xpring-js`?

Single character variables seem like a bad practice to me. I personally use `err`, but I've seen `error` as well in `catch` blocks in our codebases.

But, if `catch(e)` is consistent with the rest of `xpring-js`, then maybe we can punt this to a future Find+Replace PR.

src/XRP/legacy/legacy-default-xrp-client.ts
```diff
@@ -250,7 +251,10 @@ class LegacyDefaultXRPClient implements XRPClientDecorator {
       await this.getBalance(address)
       return true
     } catch (e) {
-      return false
+      if (e.code && e.code === grpc.status.NOT_FOUND) {
+        return false
+      }
+      throw e // error code other than NOT_FOUND should re-throw error
```
hbergren: See above comments.

test/XRP/xrp-client-integration-test.ts
```diff
@@ -115,20 +115,35 @@ describe('Xpring JS XRPClient Integration Tests', function(): void {
...
+
     assert.equal(doesExist, true)
   })
 
+  it('Check if Account Exists - false - rippled', async function(): Promise<
+    void
+  > {
+    this.timeout(timeoutMs)
+
+    const coinbaseMainnet = 'XVYUQ3SdUcVnaTNVanDYo1NamrUukPUPeoGMnmvkEExbtrj' // valid address but should NOT show up on testnet
```
hbergren: Nice. Whose address is this?

test/XRP/fakes/fake-grpc-error.ts
```diff
@@ -0,0 +1,19 @@
+import grpc from 'grpc'
+
+/**
+ * Convenience class for creating more specific Error objects to mimick gRPC error responses.
+ */
+export default class FakeGRPCError extends Error {
+  /**
```
nkramer44: I copied this class into /ILP to test my error handling.  Whoever merges second should consider putting this in a common location for all tests to use

src/XRP/default-xrp-client.ts
keefertaylor: +1 I'm sorry if I originally didn't comment. I'd favor `error` because I try to not use abbreviations in code.

test/XRP/fakes/fake-grpc-error.ts
keefertaylor: +1.

Either do this now or add a TODO and assign yourself to do it in a separate PR.

test/XRP/xrp-client-integration-test.ts
amiecorso: it's the coinbase mainnet! so by using it on Testnet we know that it's 1) valid in format and 2) nonexistant

PR #335, opened 2020-03-30
Dynamically create a transaction hash for integration tests

keefertaylor:

Create a transaction hash on the fly for integration checks rather than relying on a static hash.

### Context of Change

The Testnet nodes we use for integration tests have limited history and are reset from time to time. As a result, static hashes age out of the nodes history at least monthly, causing CI to fail.

Instead, send a payment with a known good hash and check that hash. This ensures that our hash will always be up to date. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No more hash refreshing check ins.

## Test Plan

CI


amiecorso:
whoa, genius, this was easier than I thought haha

PR #338, opened 2020-03-31
(1/2) Add a `Network` Parameter to XRPClient

keefertaylor:

Add a parameter to XRPClient that indicates the network.

### Context of Change

In the [Xpring Component](https://github.com/xpring-eng/Xpring-JS/blob/master/src/Xpring/xpring-client.ts#L17) clients are composed together. We should verify that these clients are on the same network (for instance, having a PayID Client talking to testnet and an XRP Client talking to mainnet wouldn't make sense). 

Default this parameter to being optional and add a warning (JS can only have one initializer so this is a hacky way of having multiple deprecated initializers). 

A future PR will add validation in `XpringClient`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

`XRPClient` now has network metadata associated with it.

## Test Plan

CI


keefertaylor:
Pinging reviewers: 
@intelliot  @amiecorso  @mayurbhandary

src/XRP/xrp-client.ts
```diff
@@ -20,14 +24,25 @@ class XRPClient implements XRPClientInterface {
    * The XRPClient will use gRPC to communicate with the given endpoint.
    *
    * @param grpcURL The URL of the gRPC instance to connect to.
+   * @param network The network this XRPClient is connecting to. Defaults to undefined. If set to undefined, Testnet is assumed.
    * @param useNewProtocolBuffers If `true`, then the new protocol buffer implementation from rippled will be used. Defaults to true.
    * @param forceWeb If `true`, then we will use the gRPC-Web client even when on Node. Defaults to false. This is mainly for testing and in the future will be removed when we have browser testing.
    */
   public constructor(
     grpcURL: string,
+    network: XRPLNetwork | undefined = undefined,
```
intelliot, line 33: I think this is a breaking change to the constructor since this changes the 2nd parameter.

Three options:
(1) Given this is a breaking change to the constructor anyway, if we want the `network` parameter to be required, then we could make it required right now.
(2) Move the `network` parameter to be last (after `forceWeb`). That way, existing users will continue to work without changes. (Non-breaking change)
(3) Use an [options object](https://rclayton.silvrback.com/easy-class-api-options-with-typescript-and-joi). This can be a breaking or non-breaking change, depending on whether you replace the existing parameters, or just add a new options object.

keefertaylor: Agree you're right. I'm going to just rip the band aid off. We're going to do a major semver anyway because we're moving to support rippled directly so we may as well just do this. 

Options objects are interesting - I think that this should be a required parameter but I'll consider this in the future if things keep getting more complicated.

PR #339, opened 2020-03-31
Remove TODO

keefertaylor:

After consideration, this TODO isn't actionable and nothing should be changed.

### Context of Change

Swagger generates endpoints in the form of:
`<baseURL><endpoint>`

For instance:
baseURL [host, in Payment Pointer jargon] = `https://xpring.money`
path = `/keefertaylor`
computes to: `https://xpring.money/keefertaylor`

I had wondered if we could encode `keefertaylor` into the `baseURL` (this would help with some escaping logic). If we use path = `/.well_known/pay` then Swagger by default attempts to escape the second `/`. 

Putting the path into the base URL has a snag though. Swagger forces all paths to always begin with `/` so the endpoint is always at least: `/`. 

Thus we would have:
baseURL = `https://xpring.money/keefertaylor`
path = `/`
computes to: `https://xpring.money/keefertaylor/`

The [payment pointer spec](https://interledger.org/rfcs/0026-payment-pointers/#payment-setup-protocol-receiver-endpoint-resolution) doesn't allow trailing slashes. Thus I think this idea is dead in the water. 

### Type of Change

- [x] Documentation Updates
- [ ] Release

## Before / After

No change

## Test Plan

CI (though nothing should change; only an update to a comment)


keefertaylor:
Requesting a large set of reviewers who might have input in case I'm reading the payment pointer spec wrong or there are other creative ideas.

PR #340, opened 2020-03-31
(2/2) Check networks in components of XpringClient

keefertaylor:

Throw an error if XpringClient is initialized with objects that are bound to different networks.

### Context of Change

`XpringClient` composes functionality across components. It would not make sense to have components be operating on different networks as they try to interoperate. Provide a sanity check at construction time for this case.

Note that while the `network` property exists in `PayIDClient` and `XRPClient` they are not exposed in the interface and need to be created on `FakePayIDClient` and `FakeXRPClient`.

Introduce a `XpringError` to handle these cases. Throw a XpringError if this precondition is not met.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Safer code, fewer sharp edges.

## Test Plan

Additional unit test introduced to prove correctness.

### DIFF
--- a/src/XRP/xrp-client-interface.ts
+++ b/src/XRP/xrp-client-interface.ts
@@ -1,11 +1,15 @@
 import { Wallet } from 'xpring-common-js'
 import { BigInteger } from 'big-integer'
 import TransactionStatus from './transaction-status'
+import XRPLNetwork from '../Common/xrpl-network'
 
 /**
  * An interface describing XRPClient.
  */
 export default interface XRPClientInterface {
+  /** The XRPL Network of the node that this client is communicating with. */
+  network: XRPLNetwork
+
   /**
    * Retrieve the balance for the given address.
    *
--- a/src/Xpring/xpring-error.ts
+++ b/src/Xpring/xpring-error.ts
@@ -9,8 +9,11 @@ export enum XpringErrorType {
  * Represents errors thrown by PayID components of the Xpring SDK.
  */
 export default class XpringError extends Error {
-  /**
-   * Default errors.
+  /** 
+   * Input entities given to a Xpring component were attached to different networks. 
+   *
+   * For instance, this error may be thrown if a XpringClient was constructed with 
+   * a PayIDClient attached to Testnet and a XRPClient attached to Mainnet.
    */
   public static mismatchedNetworks = new XpringError(
     XpringErrorType.MismatchedNetworks,
--- a/src/Xpring/xpring-error.ts
+++ b/src/Xpring/xpring-error.ts
@@ -1,0 +1,30 @@
+/**
+ * Types of errors that originate from Xpring.
+ */
+export enum XpringErrorType {
+  MismatchedNetworks,
+}
+
+/**
+ * Represents errors thrown by PayID components of the Xpring SDK.
+ */
+export default class XpringError extends Error {
+  /**
+   * Default errors.
+   */
+  public static mismatchedNetworks = new XpringError(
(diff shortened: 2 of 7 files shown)


keefertaylor:
\- Austin (on vacation)
\+ a new devx reviewer

src/Xpring/xpring-error.ts
```diff
@@ -0,0 +1,30 @@
...
+export enum XpringErrorType {
+  MismatchedNetworks,
+}
+
+/**
+ * Represents errors thrown by PayID components of the Xpring SDK.
+ */
+export default class XpringError extends Error {
+  /**
+   * Default errors.
```
tedkalaw, line 13: doc `mismatchedNetworks`?

src/XRP/xrp-client-interface.ts
```diff
@@ -1,11 +1,15 @@
 import { Wallet } from 'xpring-common-js'
 import { BigInteger } from 'big-integer'
 import TransactionStatus from './transaction-status'
+import XRPLNetwork from '../Common/xrpl-network'
 
 /**
  * An interface describing XRPClient.
  */
 export default interface XRPClientInterface {
+  /** The XRPL Network of the node that this client is communicating with. */
```
amiecorso, line 10: should this be `@param` like it is in `PayIDClient`? (or however they should be consistent)

src/Xpring/xpring-error.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Represents errors thrown by PayID components of the Xpring SDK.
+ */
+export default class XpringError extends Error {
+  /** 
+   * Input entities given to a Xpring component were attached to different networks. 
+   *
+   * For instance, this error may be thrown if a XpringClient was constructed with 
+   * a PayIDClient attached to Testnet and a XRPClient attached to Mainnet.
```
amiecorso, line 16: I think "`XRP`" should always be preceded by "an" instead of "a" so we don't sound dumb in the mind's ear of other devs.

test/Xpring/xpring-client-test.ts
```diff
@@ -132,4 +135,30 @@ describe('Xpring Client', function(): void {
...
+      XRPLNetwork.Test,
+    )
+    const payIDClient = new FakePayIDClient(payIDError, XRPLNetwork.Main)
+
+    // WHEN a XpringClient is constructed THEN a mismatched network XpringError is thrown.
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const xpringClient = new XpringClient(payIDClient, xrpClient)
+      assert.fail(
+        `Should not have been able to construct a but instead created: ${xpringClient}`,
```
amiecorso, line 158: construct a but

src/Xpring/xpring-error.ts
amiecorso: Though if we pronounce this "zerp" I guess we're already good

keefertaylor: Done.

src/XRP/xrp-client-interface.ts
keefertaylor: In payID client it is an actual parameter to the constructor. We use some syntactic sugar to synthesize a property. 

So when I write:
```
constructor(public readonly network: XRPLNetwork) {}
```

This expands to:
```
public readonly network: XRPLNetwork

constructor(network: XPRLNetwork) {
  this.network = network
}
```

The iVar declaration is what this interface is covering. So no need to @param this since it is just a variable that needs to exist.

PR #345, opened 2020-04-03
Changes postInvoice return type to SignatureWrapperInvoice

tedkalaw:
@dino-rodriguez and I were trying to get the wallet using `PayIDClient::postInvoice`, but the return type seemed to be a `SignatureWrapperCompliance` instead of a `SignatureWrapperInvoice`. Per @keefertaylor (described in comment below), the compliance/invoice types were swapped for request/response. 

### Context of Change
In PayID, the response body of a `postInvoice` call is another Invoice, usually with some updated data, not a Compliance.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
The yml file has the write response type for a 200 from postInvoice

## Test Plan
`npm run test`
`pay-id-client-test` should pass


tedkalaw:
i don't know how to make this create that `generated` folder in `PayID`, but in order to make this work without updating Xpring-JS, we went into (i shit you not) `system/node_modules/xpring-js/build/PayID/generated/api/DefaultApi.js` and made the change below, from Compliance to Invoice. i tried running `build` and the script for generating PayID but it didn't seem to go

```
postPathInvoice(opts, callback) {
        let postBody = opts['body'];
        let pathParams = {};
        let queryParams = {};
        let headerParams = {};
        let formParams = {};
        let authNames = [];
        let contentTypes = ['application/json'];
        let accepts = ['application/json', 'application/xrpl-mainnet+json', 'application/xrpl-testnet+json'];
        **I swear to you that we had to go in here and change this by hand lol**
        // let returnType = SignatureWrapperCompliance_1.default;        
        let returnType = SignatureWrapperInvoice_1.default;
        return this.apiClient.callApi('/invoice', 'POST', pathParams, queryParams, headerParams, formParams, postBody, authNames, contentTypes, accepts, returnType, callback);
    }
````

keefertaylor:
> I don't know, helpppp

The PayID API is generated from an OpenAPI Spec, which lives here:
https://github.com/xpring-eng/Xpring-JS/blob/master/pay-id-api-spec/pay-id.v1.yml#L36. Ideally, this would live in a submodule across JS / Java / Swift but right now it's just checked in in 3 places. Also ideally, our server side solution would consume this file too, that way we all implement the same API and we would suss these things out earlier (I don't have purview on the server architecture, but CC: @hbergren  and @dino-rodriguez. I think Hans had [negative?] opinions on this so they should do what they think is best for their project).

The yml makes some javascript through vaguely hacky build system (mostly because I haven't had time to figure out the ideal way) but in general, [regenerate_pay_id_api.sh](https://github.com/xpring-eng/Xpring-JS/blob/master/scripts/regenerate_pay_id_api.sh) is invoked to generate JS files from the YML file. You will also note that the `package.json` for node has this [script wired up to the build / clean / test phases](https://github.com/xpring-eng/Xpring-JS/blob/master/package.json#L19).

*All of this is to say, modify `pay-id.v1.yml` and run `npm build` or `npm test` and things will just work.*

The definitions for the return type of the [`/invoice` endpoint](https://github.com/xpring-eng/Xpring-JS/blob/master/pay-id-api-spec/pay-id.v1.yml#L36) for [POST requests](https://github.com/xpring-eng/Xpring-JS/blob/master/pay-id-api-spec/pay-id.v1.yml#L67) can return a [200 response](https://github.com/xpring-eng/Xpring-JS/blob/master/pay-id-api-spec/pay-id.v1.yml#L71) with a [`SignatureWrapperCompliance` type](https://github.com/xpring-eng/Xpring-JS/blob/master/pay-id-api-spec/pay-id.v1.yml#L76).

It looks like the request / response types got flipped here. You'll probably want to modify the request type for this to be `SignatureWrapper_Compliance` rather than `SignatureWrapper_Invoice`.

> Please advise

If I were you, I would swap the request / response types in the yml. Then I would go look at the following tests:
- [PayIDClient Test](https://github.com/xpring-eng/Xpring-JS/blob/master/src/PayID/pay-id-client.ts#L211) Hermetic tests that mock networking
- [PayIDClient Integration Test](https://github.com/xpring-eng/Xpring-JS/blob/master/test/PayID/pay-id-integration-test.ts): Non-hermetic tests that literally hit whatever backend that Dino / Hans have set up. 

If we have a working backend, then your best sanity check would be updating the [postInvoice integration test](https://github.com/xpring-eng/Xpring-JS/blob/master/test/PayID/pay-id-integration-test.ts#L81) to fail before your changes, then make your changes and make sure your tests pass. 

If there isn't a working backend, then you can try mucking around with mocking the network layer in the hermetic tests. 

> Interesting, the function itself has the right signature:

Ah, bummer. But good news for you because it means few code changes. The Swagger API doesn't generate typescript so you can basically throw anything at these functions. In return, I had a mocked backend which always returned a response so the bad type probably wasn't caught.


> we went into (i shit you not) system/node_modules/xpring-js/build/PayID/generated/api/DefaultApi.js and made the change below,

The code in any `Generated` folder is ignored by a .gitignore and so it isn't checked in. CI will regenerate the code transiently when it runs, so any modifications you make to the code won't reflect in our build.

> i tried running build and the script for generating PayID but it didn't seem to go

Not positive what this means :) Let me know if the above gets you started.

tedkalaw:
for posterity - keefer was exactly right about swapping the compliance/invoice req/res types. i followed up with him offline and he helped me get going

tedkalaw:
thanks for looking keefer!

PR #346, opened 2020-04-04
Updates index files in XRP submodule to export model classes

amiecorso:
The recently created protobuf wrapper classes need to be exported so they can be used by consumers of the SDK.  This PR creates those exports with an `index.ts` file in the `/XRP/model` directory, adds an `index.ts` to the `/XRP` directory, and updates the top-level `index` to be more organized as the submodules in the SDK continue to grow.

### Context of Change
Hooking up the xpring-wallet to use the SDK for `paymentHistory` requires knowing about the `XRPTransaction` class.  Having forgotten to export earlier, we'll need to cut another release of Xpring-JS before the wallet can be refactored to use `paymentHistory`.

### Type of Change

## Before / After
XRP protobuf wrapper objects are now importable from the SDK.

## Test Plan
Circle


keefertaylor:
Merging on behalf of Amie, with her permission.

PR #356, opened 2020-04-06
(2/2) Check networks in components of XpringClient

keefertaylor:
Redo of #340, merged into the wrong branch. 

----

Throw an error if XpringClient is initialized with objects that are bound to different networks.

### Context of Change

`XpringClient` composes functionality across components. It would not make sense to have components be operating on different networks as they try to interoperate. Provide a sanity check at construction time for this case.

Note that while the `network` property exists in `PayIDClient` and `XRPClient` they are not exposed in the interface and need to be created on `FakePayIDClient` and `FakeXRPClient`.

Introduce a `XpringError` to handle these cases. Throw a XpringError if this precondition is not met.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Safer code, fewer sharp edges.

## Test Plan

Additional unit test introduced to prove correctness.


CHANGELOG.md
```diff
@@ -6,11 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ## Changed
+- `XRPClient` requires a new parameter in it's constructor that identifies the network it is attached to.
```
amiecorso, line 11: its not it's

PR #367, opened 2020-04-09
Refactor Gnarly PayID code

keefertaylor:

Refactor PayID code to hide complexity of Swagger. 

### Context of Change

There are two bits of complexity:

1) Swagger methods use callbacks instead of promises
Resolve by introducing a `SwaggerResult` interface that contains properties of the call and wrapping calls to Swagger for type `T` in a `Promise<SwaggerResult<T>>`.

2) We access bare metal components of Swagger.
Refactor into a helper method that hides access to parameters which we don't need to configure. Apply the same promisification from step one to this function. 

I think this change is easiest to make atomically, but I can split it out if reviewers desire.

### Type of Change

- [ ] Documentation Updates
IO- [ ] Release

## Before / After

No change; code is more sensical and maintainable.

## Test Plan

CI


src/PayID/pay-id-client.ts
```diff
@@ -246,27 +192,31 @@ export default class PayIDClient implements PayIDClientInterface {
...
+            response,
+          })
+        },
+      )
     })
+
+    // TODO(keefertaylor): Provide more granular error handling.
+    if (error) {
+      const message = `${error.status}: ${error.response?.text}`
+      throw new PayIDError(PayIDErrorType.UnexpectedResponse, message)
```
dino-rodriguez, line 214: Wondering -- why not return the error and error type returned by the PayID server on all of these?

I would think that the services using the SDK would want to receive the error messages from the PayID server, but was curious as to what your thinking was around this.

keefertaylor: I generally favor using enum based errors because it limits the ranges of errors a client has to handle and makes code more predictable. It's possible (and maybe we should) to include the underlying error here if we really wanted to. 

I suspect / know that error handling is going to change a fair deal so I've just done this for now because it works for the moment.

PR #368, opened 2020-04-09
Improve Swagger codegen process

keefertaylor:

Build Swagger with a Maven installed plugin rather than a checked in JAR file. 

### Context of Change

This change significantly improves our build process:
- No janky installation of JDK
- No checked in JAR file
- Use a `pom.xml` and `maven` invocation to have the same effect

### Type of Change

- [ ] Documentation Updates
- [ ] Release


keefertaylor:
> Does Swagger run on Java? I'm surprised to see Maven/pom.xml files in xpring-js.

yes, Swagger is just a JAR file. You may notice that we were checking in a JAR. This improves that process by making it configureable via XML

PR #372, opened 2020-04-10
New PayID Format Support

keefertaylor:
Blocked:
- merge https://github.com/xpring-eng/xpring-common-js/pull/236
- publish new version of xpring-common-js

Migrate PayID format from:
`$xpring.money/georgewashington`
to
`georgewashington$xpring.money`

### Context of Change

Product requested change in format. 

Remove references to 'payment pointer' and rename a few classes / methods to reflect that we're not longer compliant with the payment pointer spec.

This is a breaking change, but PayID is experimental so I don't consider it a big deal.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

New spec implemented.

## Test Plan

New tests provided for CI. 

## Future Tasks
Uptake parsing logic and test cases in client libraries (JS / Swift / Java)


test/PayID/pay-id-integration-test.ts
```diff
@@ -109,7 +109,7 @@ describe('PayID Integration Tests', function (): void {
 
   it('receipt', async function (): Promise<void> {
     // GIVEN a Pay ID.
-    const payID = '$travel.payid.xpring.money/dino'
+    const payID = 'dino$travel.payid.xpring.money'
```
dino-rodriguez, line 112: @austin-king will this break after your PR on the refactor is merged? Would these have to be changed to `bob` etc. ?

austin-king: No, it just won't work in general as we never had this value seeded in the DB. We can add one if need be.

tedkalaw:
charging-approaching-photographer

test/PayID/pay-id-client-test.ts
```diff
@@ -30,37 +30,32 @@ const beneficiaryName = 'xpring'
...
+      assert.equal((error as PayIDError).errorType, PayIDErrorType.InvalidPayID)
       done()
     })
   })
 
-  it('xrpAddressForPayID - successful response - match found', async function() {
+  it('xrpAddressForPayID - successful response - match found', async function () {
     // GIVEN a PayID client, valid PayID and mocked networking to return a match for the PayID.
-    const payID = '$xpring.money/georgewashington'
+    const payID = 'georgewashington$xpring.money'
```
amiecorso, line 53: Just a thought: would it make sense to create a global variable like `VALID_PAY_ID` and use it everywhere that we're using a valid payID?  Shortens the code a little and makes it obvious what is and isn't a valid PayID.  Otherwise looks great to me.

PR #376, opened 2020-04-11
Adds documentation to protobuf wrapper classes (XRP component)

amiecorso:
Adds documentation based on the information in xrpl.org docs to the fields and constructors in native JS protobuf wrapper classes. 

### Context of Change
The Java SDK is well documented in this capacity - JS and Swift need to be made consistent. 

### Type of Change

- [x] Documentation Updates

## Test Plan
None

## Future Tasks

Do in Swift SDK

### DIFF
--- a/src/XRP/model/xrp-issued-currency.ts
+++ b/src/XRP/model/xrp-issued-currency.ts
@@ -7,6 +7,14 @@ import XRPCurrency from './xrp-currency'
  * @see: https://xrpl.org/basic-data-types.html#specifying-currency-amounts
  */
 export default class XRPIssuedCurrency {
+  /**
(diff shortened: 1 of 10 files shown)


src/XRP/model/xrp-issued-currency.ts
```diff
@@ -27,6 +35,12 @@ export default class XRPIssuedCurrency {
     return undefined
   }
 
+  /**
+   *
```
intelliot, line 39: No description for this comment?

amiecorso: Nope, you'll notice that none of the private constructors have any comments apart from documenting the individual fields.  These classes expose a different constructor which wraps the private constructor, and that's the method that gets fully documented.

PR #383, opened 2020-04-14
Adds nonce to PayIDClient::postInvoice

tedkalaw:
This PR updates `PayIDClient::postInvoice` to take a `nonce`. A nonce is required to generate an invoice: 
https://github.com/xpring-eng/payid/blob/8cdbfb3d5e846430b5f9b7582cbc57297fa93281/src/middlewares/invoices.ts#L31

Note that, just like in `getInvoice`, the nonce **is a query param**; that is, we're trying to hit `xpring.money/tedkalaw?nonce=123456`. This required me to update `pay-id-api-spec/pay-id.v1.yml` to include a new `query` param. As such, I looked at what Keefer did before for `getInvoice` and sorta copypastad.

I then had to update the tests. Since `xpring-js` is well tested, updating the mock to make sure it is hit with a `query` param was all I needed to do to verify that things were working. The integration test also continues to work.

Note that this is a breaking change - this changes the interface of `postInvoice`. Not sure how this works for versions.

### Context of Change
The PayID server expects a `nonce` parameter; because the SDK doesn't send it, we have to make changes to `master` on `payid` to to use the SDK.

### Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Updates
- [ ] Release

## Before / After
`PayIDClient::postInvoice` requires a `nonce`

## Test Plan
`npm run test`
Everything should pass


tedkalaw:
added @hbergren + @dino-rodriguez since they have context PayID side and to double check i'm actually doing the right thing :)

pay-id-api-spec/pay-id.v1.yml
```diff
@@ -99,6 +99,15 @@ paths:
             schema:
               $ref: '#/components/schemas/SignatureWrapper_Compliance'
       operationId: post-path-invoice
+      parameters:
+        - schema:
+            # Typed as a string to get around swagger-codegen issues.
+            # TODO(keefertaylor): Fix after https://github.com/swagger-api/swagger-codegen/issues/10105 is resolved
+            type: string
```
hbergren, line 106: If you want to be really specific, you could add `format: uuid` to this payload.

test/PayID/pay-id-client-test.ts
```diff
@@ -7,7 +7,7 @@ import XRPLNetwork from '../../src/Common/xrpl-network'
 import SignatureWrapperInvoice from '../../src/PayID/Generated/model/SignatureWrapperInvoice'
 import ComplianceType from '../../src/PayID/compliance-type'
 
-// Parameters for getInvoice
+// Nonce for getInvoice and postInvoice
 const nonce = '123456'
```
hbergren, line 11: We should probably update this to be a UUIDv4

test/PayID/xrp-pay-id-integration-test.ts
```diff
@@ -88,6 +88,7 @@ describe('PayID Integration Tests', function(): void {
     // WHEN the Pay ID receipt endpoint is hit
     const invoice = await payIDClient.postInvoice(
       payID,
+      '123456',
```
hbergren, line 91: UUIDv4?

test/PayID/pay-id-client-test.ts
```diff
@@ -304,11 +304,13 @@ describe('Pay ID Client', function(): void {
 
     nock('https://xpring.money')
       .post(`/georgewashington/invoice`)
+      .query({ nonce })
```
keefertaylor, line 307: Oh this is awesome. I've been just hardcoding these into URLs

tedkalaw:
@hbergren - i added https://app.asana.com/0/1145620725368172/1172270951438721 to track switching stuff to UUIDv4 but i'm going to be paranoid here and wait to do that

PR #384, opened 2020-04-14
Updates .gitignore to ignore auto-generated files in VSCode

amiecorso:
Something is automatically producing (and re-generating) the following files in VSCode:
```
scripts/.settings/*
scripts/.classpath
scripts/.project
```
and this update prevents them from getting pushed to master.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
.gitignore up to date for anyone with the same files present.

## Test Plan
CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ build/*
 
 # Ignore files from VSCode.
 *.code-workspace
+scripts/.settings/*
+scripts/.classpath
+scripts/.project
 
 # Created by https://www.gitignore.io/api/osx,node
 # Edit at https://www.gitignore.io/?templates=osx,node


amiecorso:
Noah I put you on here because these files actually seem like an eclipse/Java thing and maybe you know why they appeared in this project (which I'm looking at in VSCode).  If not, no worries.

tedkalaw:
have you found any docs on these files? any hints on docs that might suggest what conditions cause them to be generated?

nkramer44:
.project and .classpath look like eclipse files to me, but maybe VS code uses the same files types?

Whatever their purpose, they definitely shouldn't be in the git repo

tedkalaw:
still curious if we have any leads on why this is happening...is this vscode? the only reason i ask is that it seems odd to grow `.gitignore` with files unique to a person's env - you could imagine this file being arbitrarily long

PR #401, opened 2020-04-23
Uses grc-web.StatusCode instead of grpc.status

tedkalaw:
The SDK doesn't work in the wallet right now because the web version is pulling in `grpc`, which is `node` only.

I did a bisect and ended up at https://github.com/xpring-eng/Xpring-JS/pull/328.

As such, I replaced all uses of `grpc.status` with `grpcStatus`. The values for it come from https://github.com/grpc/grpc/blob/master/doc/statuscodes.md - shouts out to `UNAUTHENTICATED` for being interleaved instead of where it would go monotonically.

Also note that `ilp-error` still includes `ServiceError`, which might seem kind of weird. It is ONLY being used as a type, which is compiled away, which is why it works. We might want to consider making a type just for `ServiceError`, since it's just a bag of values.

I tested this by running tests and building the wallet against my `xpring-js` copy locally. 

### Context of Change
The `web` version of `xpring-js` shouldn't use `grpc` because it's for `node`. This was resulting in build errors in `xpring-wallet-system`'s `ui/`, since it is browser-based. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`xpring-js` `web` doesn't use `grpc` any more
as the first commit that the wallet can't build against.

## Test Plan
CI
Run tests

Use `npm link` and build against a web app; I used the wallet.


tedkalaw:
going to use `grpc-web`'s status codes, brb

test/XRP/fakes/fake-grpc-error.ts
```diff
@@ -9,12 +7,9 @@ export default class FakeGRPCError extends Error {
...
    *
    * @param message The text details of the error.
-   * @param code The grpc.status code.
+   * @param code The gRPC status code.
    */
-  constructor(
-    public readonly message: string,
-    public readonly code: grpc.status,
-  ) {
+  constructor(public readonly message: string, public readonly code: number) {
```
tedkalaw, line 12: typed as `number` because `grpc-web.StatusCode` is a `namespace`, which doesn't generate a type

src/ILP/ilp-error.ts
```diff
@@ -1,5 +1,5 @@
-import { ServiceError, status } from 'grpc'
-import { Error as grpcWebError } from 'grpc-web'
+import { ServiceError } from 'grpc'
```
Stormtv, line 1: Does importing ServiceError here still break the web system? We should avoid importing anything from GRPC in files where the browser will be consuming.

Stormtv, line 1:
```suggestion
```

src/ILP/ilp-error.ts
```diff
@@ -77,13 +77,13 @@ export default class IlpError extends Error {
   public static from(error: ServiceError | grpcWebError | IlpError): IlpError {
```
Stormtv, line 77:
```suggestion
  public static from(error: grpcWebError | IlpError): IlpError {
```

src/ILP/ilp-error.ts
tedkalaw: per offline discussion - this doesn't break it since it's a TS construct and gets compiled out, but we decided to cut it since we're not really...using it here. also as mentioned, this avoids a `grpc` important

PR #402, opened 2020-04-23
Expose CryptoAddressDetails in PayIDClient

keefertaylor:

Modify `PayIDClient` to return the full details of addresses rather than just a string.

### Context of Change

If someone put a classic address and tag into PayID they would only get the classic address back, which is problematic. 

A future PR will enforce that `XRPPayIDClient` always returns X-Addresses. 

This is a breaking change but PayID is still unreleased so I don't think it's a big deal. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Full details exposed. 

## Test Plan

Tests updated for CI.

### DIFF
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -11,6 +11,7 @@ import Compliance from './Generated/model/Compliance'
 import TravelRule from './Generated/model/TravelRule'
 import DefaultApi from './Generated/api/DefaultApi'
 import ComplianceType from './compliance-type'
+import { CryptoAddressDetails } from './Generated'
 
 interface PayIDComponents {
   host: string
@@ -59,7 +60,7 @@ export default class PayIDClient {
    * @param payID The payID to resolve for an address.
    * @returns An address representing the given PayID.
    */
-  async addressForPayID(payID: string): Promise<string> {
+  async addressForPayID(payID: string): Promise<CryptoAddressDetails> {
     const payIDComponents = PayIDClient.parsePayID(payID)
     const basePath = `https://${payIDComponents.host}`
     const { path } = payIDComponents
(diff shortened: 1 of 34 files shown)


src/PayID/pay-id-client.ts
```diff
@@ -59,7 +60,7 @@ export default class PayIDClient {
    * @param payID The payID to resolve for an address.
    * @returns An address representing the given PayID.
    */
-  async addressForPayID(payID: string): Promise<string> {
+  async addressForPayID(payID: string): Promise<CryptoAddressDetails> {
```
hbergren, line 63: So `Xpring-SDK` only supports PayID for `CryptoAddressDetails`?

How would you retrieve `AchAddressDetails`?

keefertaylor: For now yes. At some point in the future we'll have to change this and return the complete wrapper and let the client fend for themselves. That feels like a worse API and we are already regularly making breaking changes here so it doesn't feel worth future proofing.

hbergren: I bet there are a limited number of `AddressDetailType`s that we'll support.

I wonder if eventually, we can just have a `addressForPayID` that has helper functions like `cryptoAddressForPayID` and `achAddressForPayID`, and it looks at the `addressDetailType` in the response payload to determine which helper method to use.

src/PayID/xrp-pay-id-client.ts
```diff
@@ -11,4 +11,18 @@ export default class XRPPayIDClient extends PayIDClient {
...
+   *
+   * Note: Addresses are always in the X-Address format.
+   * @see https://xrpaddress.info/
+   *
+   * @param payID The payID to resolve for an address.
+   * @returns An address representing the given PayID.
+   */
+  async xrpAddressForPayID(payID: string): Promise<string> {
+    const result = await super.addressForPayID(payID)
+    return result.address
```
amiecorso, line 26: Does this not also just return the classic address?  like even though `addressForPayID` returns the full address details?  Oh wait, this is what the future PR is going to do... it's going to make sure that this address that you're digging out of the details IS an X-address and therefore incorporates the tag... Just lemme know if I'm wrong about this otherwise approving.

keefertaylor: No that's exactly the case. I was just doing it in two PRs to make the deltas clear for reviewers.

PR #413, opened 2020-04-27
Encode XRP Addresses to X-Address Format in PayID

keefertaylor:

Always return an X-Address from PayID.

### Context of Change

The current implementation of PayID would silently drop a `tag` if the result was a classic address. We recommend folks use x-addresses only, but inevitably someone will put a classic address in PayID and then something bad will happen. 

We generally only return X-Addresses from our APIs and I'd like to do the same here. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Classic addresses properly supported in PayID. 

## Test Plan

New test cases provided for CI.

### DIFF
--- a/src/PayID/xrp-pay-id-client.ts
+++ b/src/PayID/xrp-pay-id-client.ts
@@ -28,10 +28,10 @@ export default class XRPPayIDClient extends PayIDClient {
     const result = await super.addressForPayID(payID)
 
     const { address } = result
-    if (Utils.isValidXAddress(result.address)) {
+    if (Utils.isValidXAddress(address)) {
       return address
     }
-    const isTest = !(this.network === XRPLNetwork.Main)
+    const isTest = this.network !== XRPLNetwork.Main
     const encodedXAddress = Utils.encodeXAddress(address, result.tag, isTest)
     if (!encodedXAddress) {
       throw new PayIDError(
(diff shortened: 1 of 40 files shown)


src/PayID/xrp-pay-id-client.ts
```diff
@@ -23,6 +26,19 @@ export default class XRPPayIDClient extends PayIDClient {
    */
   async xrpAddressForPayID(payID: string): Promise<string> {
     const result = await super.addressForPayID(payID)
-    return result.address
+
+    const { address } = result
+    if (Utils.isValidXAddress(result.address)) {
```
tedkalaw, line 31: you can just use `address` here since you destructured it already

src/PayID/xrp-pay-id-client.ts
```diff
@@ -23,6 +26,19 @@ export default class XRPPayIDClient extends PayIDClient {
...
+      return address
+    }
+    const isTest = !(this.network === XRPLNetwork.Main)
```
tedkalaw, line 34: seems reasonable - do you typically prefer to negate the whole statement instead of testing for inequality?

src/PayID/xrp-pay-id-client.ts
keefertaylor: ty

src/PayID/xrp-pay-id-client.ts
keefertaylor: nah, you're probably right and my brain is just too tired to have done this right the first time around. 

fixed.

PR #421, opened 2020-04-30
Replace grpc with grpc-js

Stormtv:
Replace `grpc` with the new `@grpc/grpc-js` https://grpc.io/blog/grpc-js-1.0/

@grpc/grpc-js is full typescript and is the upgrade to the deprecated grpc lib.

This is a breaking change as `@grpc/grpc-js only works on Node ^8.13.0 || >=10.10.0` so this would deprecate old versions of node.js.

### Context of Change
Updated the code generation for the new protobufs.
Updated the constructors for network services to work with grpc-js.
Replaced some node imports with web imports in some places where we should always default to web.
Replaced the `toProto()` for SendPaymentResponse as it was always using the web version of send_payment_request_pb which caused issues with the new grpc-js generated files.

### Type of Change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Before / After
Nothing should really change except we are using the newer lib.

## Test Plan
CI Passes


src/XRP/grpc-xrp-network-client.ts
```diff
@@ -1,44 +1,44 @@
...
   public constructor(grpcURL: string) {
-    if (isNode()) {
-      this.grpcClient = new XRPLedgerAPIServiceClient(
-        grpcURL,
-        credentials.createInsecure(),
-      )
-    } else {
-      throw new Error('Use gRPC-Web Network Client on the browser!')
-    }
+    const XRPLedgerAPIServiceClient = grpc.makeClientConstructor(
```
keefertaylor, line 34: I think we can also remove the protocol buffer getters at the end of this file?

Stormtv:
> This is awesome. A few questions:
> 
> * Have you tested this works in the browser? Can we `npm link` it into the web wallet?
> * Do we still need to hit envoy or does this library work over HTTP2?
> 
> Thanks for the answers, it's not totally clear from the blog post and I assume you know them off the top of your head rather than me trundling through the code.

Yes this works in the browser, we do still require the envoy to my knowledge. I'll see if I can play around to figure out if there is a way to get rid of the envoy.

Stormtv:
@keefertaylor Got around to testing this and it did not work with the browser, so I reverted this PR back to just upgrade grpc with grpc-js and keep the shim

Stormtv:
Still looking into why tests are failing...

Stormtv:
Tests are finally passing now, one of the issues was the `toProto` method in the ILP payment request was using the web version even on node.

keefertaylor:
\+ @nkramer44  as FYI since this might affect Hermes

Stormtv:
Just did a big merge conflict fix it should work but now with v1.1.1 of the new grpc-js there is a better way to make the client constructors so I should update to that.

Stormtv:
Okay the service client should be typed now

Stormtv:
Linked this into Xpring Wallet System and build / tested this based on the branch here https://github.com/xpring-eng/xpring-wallet-system/pull/853. Everything works as expected.

keefertaylor:
Thanks for verifying. Approved.

PR #423, opened 2020-04-30
Enforce Ledger Specifiers as to not use Open Ledgers

Stormtv:
Currently when requesting account information and transaction history we are not setting the ledger specifier to validated ledgers only. This means all the balances we return and potentially the transaction histories can include unconfirmed values. This could lead to issues where a user thinks they have received a payment when in reality that payment could be fraudulent.

I noticed this in the XpringWallet as the balance was consistently updating in under 500ms after being funded from the faucet, this should of course take 4s on average.

### Context of Change
Add ledger specifier to the default xrp client.

### Type of Change

## Before / After
No more unconfirmed transactions impacting the user's balance.

## Test Plan
CI Passes

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -29,6 +29,7 @@ import GRPCNetworkClientWeb from './grpc-xrp-network-client.web'
 import { XRPNetworkClient } from './xrp-network-client'
 import isNode from '../Common/utils'
 import XRPError from './xrp-error'
+import { LedgerSpecifier } from './Generated/web/org/xrpl/rpc/v1/ledger_pb'
 
 /** A margin to pad the current ledger sequence with when submitting transactions. */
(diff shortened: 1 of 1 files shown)


src/XRP/default-xrp-client.ts
```diff
@@ -81,6 +82,10 @@ class DefaultXRPClient implements XRPClientDecorator {
     const request = this.networkClient.GetAccountInfoRequest()
     request.setAccount(account)
 
+    const ledger = new LedgerSpecifier()
```
keefertaylor, line 85: I actually think you could just refactor this to be:
```javascript
const accountData = this.getAccountData(address)
```

And that would simplify this method and reduce the boiler plate where you need to set this.

Stormtv: @keefertaylor Not quite sure I follow on this comment. Is that in the scope of the PR?

PR #428, opened 2020-05-05
Fix Build

keefertaylor:

Swagger generates JS files with incorrect annotations and `tsc` compiles them. This fails builds upstream. 

### Context of Change

Swagger's JS is a bit rough around the edges and it causes typescript declaration file generation to output faulty files. 

Dodge this problem by not generating .d.ts files for .js files. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

npm package actually works. 

## Test Plan

I pulled down the wallet on branch [ac/utility-refactor](https://github.com/xpring-eng/xpring-wallet-system/pull/619) and ran:
```shell
$ cd system
$ npm i 
$ tsc -d
```

I observed a typescript error. 

Then I ran `npm link` to link this build in instead and ran `tsc -d` again. No errors!

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build/**/*"
   ],
   "scripts": {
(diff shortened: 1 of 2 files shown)


Stormtv:
The original reason for AllowJS was due to us having the generated javascript. If this works great but I am surprised this didn't throw any errors.

keefertaylor:
Yeah, I just re-worked this as it broke our tests.

PR #429, opened 2020-05-05
Fix Constants in POST Invoice

keefertaylor:

Fix some minor constants to put us inline with PayID spec. Thanks to @tedkalaw  who pointed this out. 

### Context of Change

Fix constants in PayID. 

Also remove ComplianceType as a configurable parameter and hardcode to travel rule, as other values don't make sense (the rest of the inputs to this function are for travel rule). 

Breaking change, but PayID is experimental.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before: Not compliant
After: Compliant

## Test Plan

CI

### DIFF
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -198,7 +198,7 @@ export default class PayIDClient {
     const compliance = new Compliance(ComplianceType.TravelRule, travelRule)
 
     const signatureWrapper = new SignatureWrapperCompliance(
-      'compliance',
+      'Compliance',
       compliance,
       publicKeyType,
       publicKeyData,
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -10,8 +10,8 @@ import SignatureWrapperCompliance from './Generated/model/SignatureWrapperCompli
 import Compliance from './Generated/model/Compliance'
 import TravelRule from './Generated/model/TravelRule'
 import DefaultApi from './Generated/api/DefaultApi'
-import ComplianceType from './compliance-type'
 import { CryptoAddressDetails } from './Generated'
+import ComplianceType from './compliance-type'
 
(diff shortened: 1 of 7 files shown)


src/PayID/pay-id-client.ts
```diff
@@ -197,10 +195,10 @@ export default class PayIDClient {
 
     const travelRule = new TravelRule(originator, beneficiary)
 
-    const compliance = new Compliance(complianceType, travelRule)
+    const compliance = new Compliance(ComplianceType.TravelRule, travelRule)
 
     const signatureWrapper = new SignatureWrapperCompliance(
-      complianceType,
+      'compliance',
```
dino-rodriguez, line 201: These strings are uppercase in the WP

Not sure if that really affects functionality, but I would say it would be best to keep those the same

dino-rodriguez: To be clear, not all uppercase, but like: `Compliance`

src/PayID/pay-id-client.ts
```diff
@@ -197,10 +195,10 @@ export default class PayIDClient {
 
     const travelRule = new TravelRule(originator, beneficiary)
 
-    const compliance = new Compliance(complianceType, travelRule)
+    const compliance = new Compliance(ComplianceType.TravelRule, travelRule)
 
     const signatureWrapper = new SignatureWrapperCompliance(
-      complianceType,
+      'Compliance',
```
hbergren, line 201: Should this not be 'TravelRule'?

keefertaylor: I'm told by @dino-rodriguez  and @tedkalaw that it should be TravelRule.

hbergren: I did not look carefully enough at this and realize we were building a `SignatureWrapper`.

PR #432, opened 2020-05-06
Fix parameter name

keefertaylor:

The server returns `addressDetailType` not `addressDetailsType`.

### Context of Change

This feels backwards to me because the other field is called `addressDetails`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Compliant to server

## Test Plan

CI


keefertaylor:
Without a definitive guide, I'm going to merge this and adopt the change across Xpring SDK.

hbergren:
I think it might make more sense to update the PayID server to return `addressDetailsType`.

I know that's a breaking change, but we probably have one coming up anyways with Dino's return all addresses feature, and I want to make _another_ breaking change to the private API, to switch the `pay_id` property to `payId` to be consistent with the public API.

keefertaylor:
Can we file an Asana issue to track this @hbergren / @dino-rodriguez ?  I'm happy to leave this as is AS LONG as we're tracking why it's not compatible with the server version.

hbergren:
https://app.asana.com/0/1143722888069470/1175597112527169

(This is a subtask of that. I just grouped all the potential Breaking Changes things I can think of).

keefertaylor:
TY Hans!

PR #438, opened 2020-05-11
Removes LedgerSpecifier from GetAccountTransactionHistoryRequest

amiecorso:
A `GetAccountTransactionHistory` gRPC request only returns validated transaction data by default.  Adding a `LedgerSpecifer` to the request is likely to result in incomplete data, so this PR removes the `LedgerSpecifer` from `GetAccountTransactionHistoryRequest`.

For more details on when to use `LedgerSpecifier`s see the notes in these PRs:
- https://github.com/xpring-eng/xpring4j/pull/218
- https://github.com/xpring-eng/XpringKit/pull/204

### Type of Change

## Before / After
Before: Transaction history likely to be incomplete unless transactions were *just* submitted.
After: transaction history should be correct.

## Test Plan
CI / integration tests pass

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -316,10 +316,6 @@ class DefaultXRPClient implements XRPClientDecorator {
     const request = this.networkClient.GetAccountTransactionHistoryRequest()
     request.setAccount(account)
 
-    const ledger = new LedgerSpecifier()
-    ledger.setShortcut(LedgerSpecifier.Shortcut.SHORTCUT_VALIDATED)
-    request.setLedgerSpecifier(ledger)
-
     const transactionHistory = await this.networkClient.getTransactionHistory(
       request,
     )


intelliot:
To add some clarification here:

- If you specify the "validated" ledger, then you are only looking at the most recent ledger that has been validated by the whole network. That will typically contain only transactions validated in that 3-5-second window.
- A ledger contains the current account object and some transactions, but if you want _historical_ transactions, you need to consider all of the past ledgers that also contained transactions affecting your account.
- It is probably useful to retrieve/show unvalidated transactions and they should be clearly indicated as "pending" when they are shown in a UI.
  - The main consideration/question here is the user experience.
- With `account_tx` you would check the `validated` field for each tx in the response. https://xrpl.org/account_tx.html

cjcobb23:
>     * With `account_tx` you would check the `validated` field for each tx in the response. https://xrpl.org/account_tx.html
@intelliot  nice clarification. `account_tx` only ever returns validated data though, so you don't need to check the `validated` field. You do need to check `validated` when using `tx` though.

PR #439, opened 2020-05-11
Add configuration for Patch

keefertaylor:

The `project` tag applies the setting to a project, while the `patch` tag is needed to run on the PR's coverage. 

### Context of Change

Now that we have a base code coverage submitted, I am experimenting and ensuring it works. Not providing a patch configuration was one place I messed up. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Configuration runs on patch. 

## Test Plan

CI

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,11 @@
 coverage:
   status:
+    patch:
+      default:
+        target: 80%
+        threshold: 1%
+        base: auto 
+        if_ci_failed: error
     project:
       default:
         target: 80%


codecov.yml
```diff
@@ -1,5 +1,11 @@
 coverage:
   status:
+    patch:
+      default:
+        target: 80%
+        threshold: 1%
+        base: auto 
+        if_ci_failed: error
```
hbergren, line 8: Solid. So just to be clear,

The `project` tag below asserts that the project stays at 80% coverage or above, and _this_ tag asserts that any PR changes are at least 80% covered by tests. Is that right?

keefertaylor: Correct.

PR #440, opened 2020-05-11
Keeps xrpToDrops and dropsToXrp utilities in correct namespace

amiecorso:
This PR removes some exports from `index.ts` files to maintain better namespacing as the SDK grows and diversifies.  Note that this a breaking change for imports from the SDK.   The Xpring Wallet will have to be updated when this change is released with the import statement:

```import { xrpToDrops, dropsToXrp } from 'xpring-js/src/XRP/xrp-utils'```

### Context of Change
Keep the SDKs organized as we go!

### Type of Change

## Before / After
Before: `xrpToDrops` and `dropsToXrp` in `xpring-js` namespace.
After:  `xrpToDrops` and `dropsToXrp` in 'xpring-js/src/XRP/xrp-utils'

## Test Plan
CI passes

## Future Tasks

Fix Wallet import when the new release of `xpring-js` is cut.

### DIFF
--- a/src/XRP/index.ts
+++ b/src/XRP/index.ts
@@ -13,4 +13,3 @@ export {
   XRPTransactionType,
   XRPTransaction,
 } from './model'
-export { xrpToDrops, dropsToXrp } from './xrp-utils'
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,13 +15,7 @@ export { default as XRPError, XRPErrorType } from './XRP/xrp-error'
 export { default as IlpError, IlpErrorType } from './ILP/ilp-error'
 export { default as XRPPayIDClient } from './PayID/xrp-pay-id-client'
 
-export {
-  RippledFlags,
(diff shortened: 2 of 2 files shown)


keefertaylor:
+ Hans who had opinions about this (I think this is what he would want but want to make sure)

amiecorso:
> Seems reasonable! I'm still coming to terms with the best way to architect a directory structure and various exports.
> 
> How do users of this library discover what functionality is publicly available and where to import it from?

I asked Keefer about this, and the consensus is that apart from reading the code structure itself, we need more extensive documentation that will ideally be auto-generated from the code documentation.  Currently, functionality has just been demo'ed/documented in the README.  We're getting to a point of maturity where we've been starting to think about more formal documentation, but we're just not quite there yet.

In the meantime, as I should have noted in the original comment, this PR makes the structure of `xpring-js` XRP utilities consistent with the other two SDKs.

PR #446, opened 2020-05-12
Updates XRP Utils to use XRPError instead of generic Error

amiecorso:
This PR updates utilities `xrpToDrops` and `dropsToXrp` to throw an `XRPError` with enum type `XRPErrorType.InvalidInputs` instead of a generic `Error`.  This is in keeping with the implementations of these utilities in the other two SDKs. 

### Context of Change
Maintain closest possible consistency across SDK flavors.

### Type of Change

## Before / After
`xrpToDrops` and `dropsToXrp` used to throw an `Error`, and now throw an `XRPError`.

## Test Plan
CI passes.


src/XRP/xrp-error.ts
```diff
@@ -2,6 +2,7 @@
  * Types of errors that originate from interacting with XRPL.
  */
 export enum XRPErrorType {
+  InvalidInputs,
```
keefertaylor, line 5: I would make this singular, unless you really feel like it needs to be plural. 

I've generally seen this be the case in other codebases, and it seems like if you Google it most people agree to use singular: https://stackoverflow.com/questions/15755955/naming-of-enums-in-java-singular-or-plural

PR #455, opened 2020-05-14
(1/n) Rename function to indicate it does not return the latest validated ledger

keefertaylor:

The fee RPC returns a ledger sequence that is open, not the most recently validated sequence. Rename methods and iVars throughout the codebase to indicate this.

### Context of Change

Modified classes and interfaces were never exposed so this is a non-breaking change. 

This affects our reliable transaction submission logic - future PRs will address this.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Code is named after its behavior

## Test Plan

CI

### DIFF
--- a/src/XRP/reliable-submission-xrp-client.ts
+++ b/src/XRP/reliable-submission-xrp-client.ts
@@ -54,7 +54,7 @@ class ReliableSubmissionXRPClient implements XRPClientDecorator {
     }
 
     // Retrieve the latest ledger index.
-    let latestLedgerSequence = await this.getLastValidatedLedgerSequence()
+    let latestLedgerSequence = await this.getCurrentOpenLedgerSequence()
 
     // Poll until the transaction is validated, or until the lastLedgerSequence has been passed.
     /*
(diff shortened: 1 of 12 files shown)


src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -54,7 +54,7 @@ class ReliableSubmissionXRPClient implements XRPClientDecorator {
     }
 
     // Retrieve the latest ledger index.
-    let latestLedgerSequence = await this.getLastValidatedLedgerSequence()
+    let latestLedgerSequence = await this.getCurrentOpenLedgerSequence()
```
tedkalaw, line 57: random q, as i don't know enough about xrpl - is this necessarily the most recent ledger sequence overall? mostly curious of the var should be renamed as well

keefertaylor: Nah, we should probably rename this. Technically this code is just plain wrong now, and it's corrected in #459.

PR #456, opened 2020-05-14
(2/n) Create an isLedgerSequenceValidated() function

keefertaylor:

Create a new function with unit and integration tests.

### Context of Change

This is a necessary building block to fix a bug in our reliable transaction submission logic. `DefaultXpringClient` is not exposed to the world.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A
## Test Plan

New tests provided for CI

### DIFF
--- a/test/XRP/default-xrp-client-test.ts
+++ b/test/XRP/default-xrp-client-test.ts
@@ -609,7 +609,7 @@ describe('Default Xpring Client', function (): void {
   it('isLedgerSequenceValidated - Validated Ledger', async function (): Promise<
     void
   > {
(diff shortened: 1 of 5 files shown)


test/XRP/default-xrp-client-test.ts
```diff
@@ -604,4 +605,106 @@ describe('Default Xpring Client', function (): void {
       done()
     })
   })
+
+  it('isLedgerSequenceValidated - Validated Ledger', async function (): Promise<
+    void
+  > {
+    // GIVEN an XRPClient client which will return an account info object indicating a validated ledger.
```
amiecorso, line 612: can omit redundant "client" here and throughout file?

PR #465, opened 2020-05-15
Adds `validated` and `ledgerIndex` fields to `XRPTransaction`

amiecorso:
This PR adds the `validated` and `ledgerIndex` fields to the `XRPTransaction` object. 
These fields come from a `GetTransactionResponse` protobuf object, which is returned by a call to `getTransaction`. 

### Context of Change
This change is made in preparation for implementing a `getTransaction` method in the SDKs. The original implementation of the `XRPTransaction` class in the SDK only included the fields present in the `Transaction` protocol buffer object (which is wrapped by a `GetTransactionResponse` also containing other metadata about the transaction's outcome.) Certain of these other metadata fields have been added as-needed to the `XRPTransaction` object, and this PR takes care of the remaining fields available in `GetTransactionResponse`.

Our current approach is to combine the distinct `GetTransactionResponse` and `Transaction` protocol buffer objects into one type (`XRPTransaction`) since the metadata about the transaction's outcome is of equal importance to the specific transaction data, and while the type distinction makes sense in the context of authoring rippled, it would add unnecessary complexity in the SDK. 

### Type of Change

## Before / After
Two additional fields available on an `XRPTransaction` object.

## Test Plan
Added new test values to fake protobuf objects for conversion tests.
CI passes.

## Future Tasks

Update Java and Swift SDKs with the same changes.


src/XRP/model/xrp-transaction.ts
```diff
@@ -160,7 +166,12 @@ export default class XRPTransaction {
    * @param type The type of transaction.
    * @param paymentFields An XRPPayment object representing the additional fields present in a PAYMENT transaction.
    *                      see "https://xrpl.org/payment.html#payment-fields"
+   * @param deliveredAmount The actual amount delivered by this transaction, in the case of a partial payment.
```
amiecorso, line 169: I noticed this documentation was missing so just sneaking it in here.

PR #470, opened 2020-05-20
Implements `getPayment` in `XRPClient`

amiecorso:
This PR implements `getTransaction` in `XRPClient` and all related classes, which queries the XRP ledger for a transaction by hash, and if found, returns an `XRPTransaction` object representing the transaction.  Unit and integration tests added.

Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/230

### Context of Change
Mayur mentioned that a partner (presumably Forte) was looking for this functionality in the SDK. 

### Type of Change

## Before / After
`XRPClient` now exposes `getTransaction`

## Test Plan
New unit and integration tests pass on CI. 

## Future Tasks

Implement same functionality in Swift SDK

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -388,7 +388,7 @@ class DefaultXRPClient implements XRPClientDecorator {
    *       See the `validated` field to make this distinction.
    *
    * @param transactionHash The hash of the transaction to retrieve.
-   * @throws Error if the transaction hash was invalid.
+   * @throws An error if the transaction hash was invalid.
    * @return An XRPTransaction object representing an XRP Ledger transaction.
    */
   public async getPayment(
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -336,7 +336,7 @@ class DefaultXRPClient implements XRPClientDecorator {
    * Note: This method only works for payment type transactions, see: https://xrpl.org/payment.html
    * Note: This method only returns the history that is contained on the remote node, which may not contain a full history of the network.
    *
-   * @param address: The address (account) for which to retrive payment history.
+   * @param address: The address (account) for which to retrieve payment history.
    * @throws: An error if there was a problem communicating with the XRP Ledger.
    * @return: An array of transactions associated with the account.
    */
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -389,7 +389,7 @@ class DefaultXRPClient implements XRPClientDecorator {
    *
    * @param transactionHash The hash of the transaction to retrieve.
    * @throws An error if the transaction hash was invalid.
-   * @return An XRPTransaction object representing an XRP Ledger transaction.
+   * @return An {@link XRPTransaction} object representing an XRP Ledger transaction.
    */
   public async getPayment(
     transactionHash: string,
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -388,7 +388,7 @@ class DefaultXRPClient implements XRPClientDecorator {
    *       See the `validated` field to make this distinction.
    *
    * @param transactionHash The hash of the transaction to retrieve.
-   * @throws XRPException If the transaction hash was invalid.
+   * @throws Error if the transaction hash was invalid.
    * @return An XRPTransaction object representing an XRP Ledger transaction.
    */
   public async getPayment(
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -380,6 +380,28 @@ class DefaultXRPClient implements XRPClientDecorator {
     } // end for
     return payments
   }
+
+  /**
+   * Retrieve the payment transaction corresponding to the given transaction hash.
+   *
+   * Note: This method can return transactions that are not included in a fully validated ledger.
+   *       See the `validated` field to make this distinction.
+   *
+   * @param transactionHash The hash of the transaction to retrieve.
+   * @throws XRPException If the transaction hash was invalid.
+   * @return An XRPTransaction object representing an XRP Ledger transaction.
+   */
+  public async getPayment(
+    transactionHash: string,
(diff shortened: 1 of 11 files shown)


src/XRP/default-xrp-client.ts
```diff
@@ -336,7 +336,7 @@ class DefaultXRPClient implements XRPClientDecorator {
    * Note: This method only works for payment type transactions, see: https://xrpl.org/payment.html
    * Note: This method only returns the history that is contained on the remote node, which may not contain a full history of the network.
    *
-   * @param address: The address (account) for which to retrive payment history.
+   * @param address: The address (account) for which to retrieve payment history.
```
hbergren, line 339: 👍 Good attention to detail!

src/XRP/default-xrp-client.ts
```diff
@@ -380,6 +380,28 @@ class DefaultXRPClient implements XRPClientDecorator {
     } // end for
     return payments
   }
+
+  /**
+   * Retrieve the payment transaction corresponding to the given transaction hash.
+   *
+   * Note: This method can return transactions that are not included in a fully validated ledger.
+   *       See the `validated` field to make this distinction.
```
hbergren, line 388: This is the validated field in the retrieved transaction?

Worth adding a `@see/@link` tag to link to XRPL documentation?

src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -123,6 +123,12 @@ class ReliableSubmissionXRPClient implements XRPClientDecorator {
   public async paymentHistory(address: string): Promise<Array<XRPTransaction>> {
     return this.decoratedClient.paymentHistory(address)
   }
+
+  public async getPayment(
+    transactionHash: string,
+  ): Promise<XRPTransaction | undefined> {
+    return this.decoratedClient.getPayment(transactionHash)
```
hbergren, line 130: Total aside: What is the `this.decoratedClient` pattern going on here?

src/XRP/xrp-client-decorator.ts
```diff
@@ -10,7 +10,7 @@ export interface XRPClientDecorator {
    * Retrieve the balance for the given address.
    *
    * @param address The X-Address to retrieve a balance for.
-   * @returns A `BigInteger` representing the number of drops of XRP in the account.
+   * @return A `BigInteger` representing the number of drops of XRP in the account.
```
hbergren, line 13: I would use `@returns` over `@return`.

Reasoning:
1. It reads more like a sentence ("Returns a BigInteger...")
2. [`@returns` is the standard for TSDoc](https://github.com/microsoft/tsdoc/blob/master/tsdoc/src/details/StandardTags.ts#L421), which does not yet support tag synonyms
3. [`@returns` is the one actually documented in JSDoc, with `@return` used as a potential synonym](https://jsdoc.app/tags-returns.html)

src/XRP/xrp-client-decorator.ts
```diff
@@ -81,9 +81,21 @@ export interface XRPClientDecorator {
...
+   * Retrieve the payment transaction corresponding to the given transaction hash.
+   *
+   * Note: This method can return transactions that are not included in a fully validated ledger.
+   *       See the `validated` field to make this distinction.
+   *
+   * @param transactionHash The hash of the transaction to retrieve.
+   * @throws An error if the transaction hash was invalid.
+   * @return An XRPTransaction object representing an XRP Ledger transaction.
+   */
+  getPayment(transactionHash: string): Promise<XRPTransaction | undefined>
```
hbergren, line 100: Totally random question, why is this an interface as opposed to an `abstract class`?

(I'm not sure how this is actually used, so that may answer my question).

It seems weird that we have to JSDoc this same function in so many places. I wonder if there's a clean way to not do that. (Might be out of scope for this PR though).

test/XRP/default-xrp-client-test.ts
```diff
@@ -604,4 +607,88 @@ describe('Default Xpring Client', function (): void {
...
+    )
+    const fakeNetworkClient = new FakeXRPNetworkClient(fakeNetworkResponses)
+    const xrpClient = new DefaultXRPClient(fakeNetworkClient)
+
+    // WHEN a transaction is requested.
+    const transaction = await xrpClient.getPayment(transactionHash)
+
+    // THEN the result is undefined
+    assert.isUndefined(transaction)
+  })
```
hbergren, line 693: Is it worth adding an integration test that isn't mocked? I'm not sure how you guys make the decision for what to mock vs not mock?

src/XRP/xrp-client-decorator.ts
keefertaylor: It hasn't been clear to me how to handle this problem either. I've been JSDoc'ing everything, since ultimately I'd like to generate documentation automatically from source. 

I'm curious what benefits you think an abstract class would have here?

src/XRP/default-xrp-client.ts
amiecorso: So my only hesitation to linking to the XRPL documentation is that `validated` is actually a derived field that is calculated when a response to the `tx` JSON API call is assembled.  (OR when a response to the gRPC equivalent is assembled, which follows the JSON as closely as possible).  There is xrpl.org documentation for the `tx` command, but it might be confusing to be linking to a different API from the SDK. Instead, adding an `@link XRPTransaction` in the return, and the documentation there links to the ledger status documentation https://xrpl.org/ledgers.html#open-closed-and-validated-ledgers

src/XRP/xrp-client-decorator.ts
amiecorso: great, I was wondering about this - thanks for clarifying.  Fixed them in the relevant files here, and made a note to myself to apply to the entire SDK for consistency.

test/XRP/default-xrp-client-test.ts
amiecorso: So everything in this file is mocked and I believe considered a unit test?  The integration tests for `DefaultXRPClient` are essentially the integration tests for `XRPClient` because the `XRPClient` decorates an underlying `DefaultXRPClient`.

PR #473, opened 2020-05-21
Properly namespace `XRPUtils` for export/import hygiene

amiecorso:
This PR fixes the namespacing problems WRT `xrpToDrops` and `dropsToXrp` utility functions.  Initially, I had these exported naked at the top level of `xpring-js`.  In an effort to keep things organized, these should be namespaced as `XRPUtils`.  However, I did that wrong by removing the top-level exports and believing I could import based on the directory structure in a similar way to other languages' package style (Java Python etc.), but that doesn't work and led to an import statement all the way from `node_modules` in the wallet.   This PR should achieve the goal.

This will be a breaking change for the Wallet, which should change its import to:
`import {...., XRPUtils} from 'xpring-js'`
And should then resolve scope with:
`  ... XRPUtils.xrpToDrops(...) .. ` etc. 

@keefertaylor : There are utilities in `xpring-common-js` that are currently being exported as `Utils` from `xpring-js`.  Maybe we should audit this situation and decide of any of those should be considered `XRPUtils`, and the rest we could consider exporting as `XpringUtils` or something... just because `Utils` is such a pollutant.

### Context of Change
Keep the SDKs organized.

### Type of Change

## Before / After
`xrpToDrops` and `dropsToXrp` available in `XRPUtils`

## Test Plan
CI
Also tested wallet imports via `npm link`

### DIFF
--- a/src/XRP/xrp-utils.ts
+++ b/src/XRP/xrp-utils.ts
@@ -1,117 +1,113 @@
-/* eslint-disable no-inner-declarations */
-/* eslint-disable @typescript-eslint/no-namespace */
 import BigNumber from 'bignumber.js'
 import { XRPErrorType, XRPError } from '..'
 
-namespace XRPUtils {
-  export function dropsToXrp(drops: BigNumber.Value): string {
-    const dropsRegEx = RegExp(/^-?[0-9]*\.?[0-9]*$/)
+function dropsToXrp(drops: BigNumber.Value): string {
+  const dropsRegEx = RegExp(/^-?[0-9]*\.?[0-9]*$/)
 
-    if (typeof drops === 'string') {
-      if (!dropsRegEx.exec(drops)) {
-        throw new XRPError(
-          XRPErrorType.InvalidInput,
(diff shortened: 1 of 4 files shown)


src/XRP/xrp-utils.ts
```diff
@@ -110,4 +110,5 @@ function xrpToDrops(xrp: BigNumber.Value): string {
     .toString(10)
 }
 
-export { xrpToDrops, dropsToXrp }
+const XRPUtils = { dropsToXrp, xrpToDrops }
+export default XRPUtils
```
hbergren, line 114: 👍

hbergren: Maybe this is the pattern I should use in `xpring-common-js` for this same problem?

(As opposed to:
```ts
const XRPUtils = {
  dropsToXrp(drops) {...},
  ...
}
```

keefertaylor: This does seem cleaner to me but I ultimately don't care as long as we're consistent.

PR #474, opened 2020-05-22
Updates in response to breaking PayID changes - Public API

amiecorso:
This PR updates the SDK to handle breaking changes made to PayID.  Specifically, the `PaymentInformation` type was changed to return addresses in the form of an array of `Address` (new type) objects. 

Note: I think Codecov is failing because there are so many changes to the .yml?  I think this is ok?

@keefertaylor check out my TODO in `PayIDClient`.

---------------------------------------------------------------
New object structure in payid:

Screen Shot 2020-05-21 at 6 42 45 PM

### Type of Change

## Before / After
`PayIDClient` unit and integration tests pass again.

## Test Plan
CI

## Future Tasks

Update other SDKs

### DIFF
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -89,8 +89,8 @@ export default class PayIDClient {
         throw new PayIDError(PayIDErrorType.UnexpectedResponse, message)
       }
       // TODO(keefertaylor): make sure the header matches the request.
-    } else if (data?.addressDetails) {
-      return data.addressDetails
+    } else if (data?.addresses) {
+      return data.addresses[0].addressDetails
     } else {
       throw new PayIDError(PayIDErrorType.UnexpectedResponse)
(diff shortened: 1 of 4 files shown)


src/PayID/pay-id-client.ts
```diff
@@ -89,8 +89,9 @@ export default class PayIDClient {
         throw new PayIDError(PayIDErrorType.UnexpectedResponse, message)
       }
       // TODO(keefertaylor): make sure the header matches the request.
-    } else if (data?.addressDetails) {
-      return data.addressDetails
+    } else if (data?.addresses) {
+      // TODO(amiecorso): what if addresses is empty or contains more than one CryptoAddressDetails?
```
keefertaylor, line 93: Given that this client only ever makes a request for one type of address (right @dino-rodriguez ?) I think we should throw an error here.

PR #482, opened 2020-05-27
Adds error condition and unit test for multiple returned addresses for PayID

amiecorso:
This PR follows up a [TODO here](https://github.com/xpring-eng/Xpring-JS/pull/474), throwing an error if a PayID lookup returns a number of addresses other than 1, and introducing a unit test to check this behavior.
This functionality is already implemented in the other two SDKs as part of the original response to the breaking PublicAPI changes.

### Context of Change
The SDKs will likely support multiple address lookup and content negotiation in the future.  However, the current implementation of a `PayIDClient` or `XRPPayIDClient` is tied to a specific network, and therefore should only ever be retrieving a single address for a PayID.

### Type of Change

## Before / After
Error now thrown if the number of addresses returned by a PayID lookup is different from 1.

## Test Plan
New unit test introduced. CI passes.

### DIFF
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -90,8 +90,8 @@ export default class PayIDClient {
       }
       // TODO(keefertaylor): make sure the header matches the request.
     } else if (data?.addresses) {
+      // With a specific network, exactly one address should be returned by a PayId lookup.
       if (data.addresses.length === 1) {
-        // With a specific network, exactly one address should be returned by a PayId lookup.
         return data.addresses[0].addressDetails
       }
       throw new PayIDError(PayIDErrorType.UnexpectedResponse)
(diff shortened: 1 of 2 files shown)


src/PayID/pay-id-client.ts
```diff
@@ -90,8 +90,11 @@ export default class PayIDClient {
       }
       // TODO(keefertaylor): make sure the header matches the request.
     } else if (data?.addresses) {
-      // TODO(amiecorso): what if addresses is empty or contains more than one CryptoAddressDetails?
-      return data.addresses[0].addressDetails
+      // With a specific network, exactly one address should be returned by a PayId lookup.
```
hbergren, line 93:
```suggestion
      // With a specific network, exactly one address should be returned by a PayID lookup.
```

src/PayID/pay-id-client.ts
```diff
@@ -90,8 +90,11 @@ export default class PayIDClient {
       }
       // TODO(keefertaylor): make sure the header matches the request.
     } else if (data?.addresses) {
```
hbergren, line 92: Can we compress this?

```suggestion
    } else if (data?.addresses?.length === 1) {
```

PR #487, opened 2020-05-29
Adds X-Addresses to `XRPTransaction` and `XRPPayment` model objects

amiecorso:
This PR adds X-address to the `XRPPayment` and `XRPTransaction` model objects.  `XRPPayment` contains a `destination` and a `destinationTag` property, and now also contains a `destinationXAddress` property that is the X-address encoded version of this information. `XRPTransaction` contains an `account` property, and now also contains an `accountXAddress` property, which is the X-address encoded version of the originating account with no destination tag.

In order to create the correct X-address encoding, we need to know whether we're on Testnet or Mainnet.  Therefore the `.from` constructors for `XRPPayment` and `XRPTransaction` now have an additional `XRPLNetwork` parameter.  Tests and client code have been updated accordingly. 

### Type of Change

## Before / After
`XRPTransaction` and `XRPPayment` now contain X-address versions of address and tag properties. 

## Test Plan
Unit tests for protocol buffer conversion into these objects have been updated to verify the X-address properties.
CI stil passes.

## Future Tasks

Questions for @keefertaylor 
- do we need to deprecate the classic address/tag fields, or just leave everything side-by-side?
- should a note about this go in the CHANGELOG?  Technically the `XRPTransaction` and `XRPPayment` objects experience a breaking change because their constructors changed.
- should other classes that contain XRP addresses (such as `XRPPathElement` or `XRPIssuedCurrency`) also contain X-address versions of these?

### DIFF
--- a/src/XRP/model/xrp-payment.ts
+++ b/src/XRP/model/xrp-payment.ts
@@ -1,3 +1,4 @@
+import { Utils } from 'xpring-common-js'
 import { Payment } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
 import XRPCurrencyAmount from './xrp-currency-amount'
 import XRPPath from './xrp-path'
@@ -13,10 +14,11 @@ export default class XRPPayment {
    *
    * @param payment a Payment (protobuf object) whose field values will be used
    *                to construct an XRPPayment
+   * @param isTest Whether this Payment object came from the XRPL Testnet, defaults to `false`.
    * @return an XRPPayment with its fields set via the analogous protobuf fields.
    * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L224
    */
-  public static from(payment: Payment): XRPPayment | undefined {
+  public static from(payment: Payment, isTest = false): XRPPayment | undefined {
     const paymentAmountValue = payment.getAmount()?.getValue()
     const amount =
       paymentAmountValue && XRPCurrencyAmount.from(paymentAmountValue)
@@ -30,6 +32,12 @@ export default class XRPPayment {
     }
 
     const destinationTag = payment.getDestinationTag()?.getValue()
+
+    const destinationXAddress = Utils.encodeXAddress(
+      destination,
+      destinationTag,
+      isTest,
+    )
 
     // If the deliverMin field is set, it must be able to be transformed into a XRPCurrencyAmount.
     const paymentDeliverMinValue = payment.getDeliverMin()?.getValue()
@@ -58,6 +66,7 @@ export default class XRPPayment {
       amount,
       destination,
       destinationTag,
+      destinationXAddress,
       deliverMin,
       invoiceID,
       paths,
@@ -70,6 +79,8 @@ export default class XRPPayment {
    * @param amount The amount of currency to deliver.
    * @param destination The unique address of the account receiving the payment.
    * @param destinationTag (Optional) Arbitrary tag that identifies a hosted recipient to pay, or the reason for the payment.
+   * @param destinationXAddress The address and (optional) destination tag of the account receiving the payment,
+   *                             encoded in X-Address format. See https://xrpaddress.info/.
    * @param deliverMin (Optional) Minimum amount of destination currency this transaction should deliver.
    * @param invoiceID (Optional) Arbitrary 256-bit hash representing a specific reason or identifier for this payment.
    * @param paths Array of payment paths to be used for this transaction.  Must be omitted for XRP-to-XRP transactions.
@@ -79,6 +90,7 @@ export default class XRPPayment {
     readonly amount?: XRPCurrencyAmount,
     readonly destination?: string,
     readonly destinationTag?: number,
+    readonly destinationXAddress?: string,
     readonly deliverMin?: XRPCurrencyAmount,
     readonly invoiceID?: Uint8Array,
     readonly paths?: Array<XRPPath | undefined>,
--- a/src/XRP/model/xrp-payment.ts
+++ b/src/XRP/model/xrp-payment.ts
@@ -2,6 +2,7 @@ import { Utils } from 'xpring-common-js'
 import { Payment } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
 import XRPCurrencyAmount from './xrp-currency-amount'
 import XRPPath from './xrp-path'
+import XRPLNetwork from '../../Common/xrpl-network'
 
 /*
  * Represents a payment on the XRP Ledger.
@@ -14,11 +15,14 @@ export default class XRPPayment {
    *
    * @param payment a Payment (protobuf object) whose field values will be used
    *                to construct an XRPPayment
-   * @param isTest Whether this Payment object came from the XRPL Testnet, defaults to `false`.
+   * @param xrplNetwork The XRPL network from which this object was retrieved, defaults to XRPLNetwork.Main (Mainnet).
    * @return an XRPPayment with its fields set via the analogous protobuf fields.
    * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L224
    */
-  public static from(payment: Payment, isTest = false): XRPPayment | undefined {
+  public static from(
+    payment: Payment,
+    xrplNetwork = XRPLNetwork.Main,
+  ): XRPPayment | undefined {
     const paymentAmountValue = payment.getAmount()?.getValue()
     const amount =
       paymentAmountValue && XRPCurrencyAmount.from(paymentAmountValue)
@@ -36,7 +40,7 @@ export default class XRPPayment {
     const destinationXAddress = Utils.encodeXAddress(
       destination,
       destinationTag,
-      isTest,
+      xrplNetwork === XRPLNetwork.Test,
     )
 
     // If the deliverMin field is set, it must be able to be transformed into a XRPCurrencyAmount.
--- a/src/XRP/model/xrp-transaction.ts
+++ b/src/XRP/model/xrp-transaction.ts
@@ -19,11 +19,13 @@ export default class XRPTransaction {
    *
    * @param transaction a Transaction (protobuf object) whose field values will be used
    *                    to construct an XRPTransaction
+   * @param isTest Whether this Transaction object came from the XRPL Testnet, defaults to `false`.
    * @returns an XRPTransaction with its fields set via the analogous protobuf fields.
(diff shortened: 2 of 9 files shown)


src/XRP/model/xrp-payment.ts
```diff
@@ -1,6 +1,8 @@
+import { Utils } from 'xpring-common-js'
 import { Payment } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
 import XRPCurrencyAmount from './xrp-currency-amount'
 import XRPPath from './xrp-path'
+import XRPLNetwork from '../../Common/xrpl-network'
```
keefertaylor, line 5: You can remove the TODO on line 11

src/XRP/model/xrp-payment.ts
```diff
@@ -79,6 +94,7 @@ export default class XRPPayment {
     readonly amount?: XRPCurrencyAmount,
     readonly destination?: string,
     readonly destinationTag?: number,
+    readonly destinationXAddress?: string,
```
keefertaylor, line 97: Can we mark destinationAddress and destinationTag as deprecated?

src/XRP/model/xrp-transaction.ts
```diff
@@ -176,6 +190,7 @@ export default class XRPTransaction {
   private constructor(
     readonly hash: string,
     readonly account?: string,
+    readonly accountXAddress?: string,
```
keefertaylor, line 193: Same comment on deriving this internally.

keefertaylor, line 193: Curious on your thoughts: We have `account` and `sourceTag`. Should we also make a sourceXAddress? 

I think yes.

src/XRP/model/xrp-payment.ts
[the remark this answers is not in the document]
amiecorso: so if I remove `destinationXAddress` as a parameter to the `XRPPayment` constructor, then I have to add `xrplNetwork` as a parameter to the `XRPPayment` constructor in order to use it inside the constructor while encoding the address - do you think that still makes sense?

amiecorso: yep, and just to clarify - *don't* mark `account` and/or `sourceTag` deprecated in `XRPTransaction`?  Or should those be marked deprecated too?  I'm trying to understand whether there might be use cases for these objects in which the classic address format might be useful or necessary?

src/XRP/model/xrp-transaction.ts
amiecorso: derived this internally too

src/XRP/model/xrp-payment.ts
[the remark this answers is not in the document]
amiecorso: I implemented it this way - let me know if it looks good / is documented correctly.

keefertaylor:
We should probably update CHANGELOG.md as well.

src/XRP/model/xrp-payment.ts
```diff
@@ -1,22 +1,33 @@
...
+   * See https://xrpaddress.info/.
+   */
+  destinationXAddress: string | undefined
+
   /**
    * Constructs an XRPPayment from a Payment.
    *
    * @param payment a Payment (protobuf object) whose field values will be used
    *                to construct an XRPPayment
+   * @param xrplNetwork The XRPL network from which this object was retrieved, defaults to XRPLNetwork.Main (Mainnet).
```
keefertaylor, line 23: Is there a reason to default here? Why not force users to explicitly say where this came from?

src/XRP/model/xrp-payment.ts
```diff
@@ -67,21 +79,31 @@ export default class XRPPayment {
...
     readonly amount?: XRPCurrencyAmount,
     readonly destination?: string,
     readonly destinationTag?: number,
     readonly deliverMin?: XRPCurrencyAmount,
     readonly invoiceID?: Uint8Array,
     readonly paths?: Array<XRPPath | undefined>,
     readonly sendMax?: XRPCurrencyAmount,
-  ) {}
+  ) {
+    this.destinationXAddress = destination
```
keefertaylor, line 101: It's kinda ugly that we're going to make a public parameter be an `string | undefined` rather than a `string`, especially since this should *never* fail (destinationAddress and destinationTag should always be able to be encoded). 

I don't think we should ship this to every client of the class by having a property that is `string | undefined`. There's two options here:
1) Have this initializer throw an `unknown` exception if this ever fails. Catch exceptions in the `from` method and return `undefined` from there. 
2) We could revert to what you had before - which is maybe a bit cleaner? 

Open to suggestions here and I'm sorry for the churn.

src/XRP/model/xrp-payment.ts
amiecorso: That's a good point... `xrplNetwork` defaults in the client classes, which makes sense, but since it's the client classes that feed their value to these objects, we may as well not default.  One reason to default is to avoid having to update all of the existing constructor calls, but that's not a good reason.

amiecorso: It'll also make our Java code cleaner, which doesn't deal well with default arguments.  I'll update this.

src/XRP/model/xrp-payment.ts
```diff
@@ -31,6 +36,16 @@ export default class XRPPayment {
...
 
+    const destinationXAddress = Utils.encodeXAddress(
+      destination,
+      destinationTag,
+      xrplNetwork === XRPLNetwork.Test,
+    )
+    // An X-address should always be able to be encoded.
+    if (!destinationXAddress) {
+      return undefined
+    }
```
hbergren, line 47: Instead of returning `undefined`, should we just use a `!` not-null operator:
```ts
// eslint-disable-next-line ...
const xAddress = Utils.encodeXAddress(...)!
```

src/XRP/model/xrp-payment.ts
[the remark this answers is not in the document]
hbergren: Ah, I see what you mean. I'm not sure what the best solution is, maybe @keefertaylor has an opinion?

src/XRP/model/xrp-payment.ts
amiecorso: So I'm inclined to continue returning `undefined` here because that is how we're treating other error states that should never happen in all of these wrapper classes.  The X-address *should* always be encodable, but if it isn't, I do want the whole object to be rotten.

amiecorso:
So I reverted this to encoding the X-addresses in the `.from` constructor along with all of the other fields because I think it's much cleaner, and it also causes all of the field documentation to stay in the same place, and gets rid of that extra layer of `XRPLNetwork` argument passing.  Currently, all of these protobuf wrapper classes are extremely minimal in terms of sanity checks and error messages because the idea is that we trust the server not to give us bogus data.  Something on my radar to discuss further with Keefer is whether we should go through and systematically implement some better checking/error infrastructure... because while I trust the rippled server, it is still possible that there are errors in the gRPC implementation and we should probably be throwing a fit if we encounter them instead of failing silently.

src/XRP/model/xrp-payment.ts
amiecorso: went with 2

amiecorso:
> Looks reasonable.
> 
> What is `xrp-offer-cancel.ts`?

I'm not sure how this got in there... it's one of the new proto wrapper classes I'm working on but looks like an empty file so I deleted.

src/XRP/model/xrp-transaction.ts
```diff
@@ -6,31 +6,42 @@ import XRPTransactionType from './xrp-transaction-type'
...
     }
 
     const account = transaction.getAccount()?.getValue()?.getAddress()
+    if (!account) {
+      return undefined
+    }
+
+    const accountXAddress = Utils.encodeXAddress(
+      account,
+      undefined,
```
keefertaylor, line 42: shouldn't we pass sourceAddress here?

src/XRP/model/xrp-transaction.ts
```diff
@@ -142,12 +161,13 @@ export default class XRPTransaction {
   }
 
   /**
-   *
    * @param hash The identifying hash of the transaction.
    * @param account The unique address of the account that initiated the transaction.
```
keefertaylor, line 165: @deprecated

src/XRP/model/xrp-transaction.ts
```diff
@@ -162,20 +182,23 @@ export default class XRPTransaction {
    * @param sourceTag (Optional) Arbitrary integer used to identify the reason for this payment or a sender on whose behalf this
```
keefertaylor, line 182: @deprecated

src/XRP/model/xrp-payment.ts
keefertaylor: +1 to Amie's point here - we shouldn't use force unwraps in non-test code IMO. 

I realize this is a logical fallacy right now, but there could be **some** point in the future where preconditions change and this is no longer true. Then force unwraps will fail and things will break in unexpected ways at runtime

src/XRP/model/xrp-payment.ts
keefertaylor: Yup I would mark these as deprecated too. Commented as such.

src/XRP/model/xrp-transaction.ts
keefertaylor: EDIT: I see what's happening. 

I think there should just be a `sourceXAddress` which is the union of `account` and `sourceTag`. Instead of having two source X-Addresses. 

[The doc for `sourceTag`](https://xrpl.org/transaction-common-fields.html) says:
"""
Arbitrary integer used to identify the reason for this payment, or a sender on whose behalf this transaction is made. Conventionally, a refund should specify the initial payment's SourceTag as the refund payment's DestinationTag.
"""

So you could imagine that you want to reverse a payment, you'd just use:
```
const paymentToReverse = ...
const destinationForReverse  = paymentToReverse.sourceXAddress
```

And things just work. Is there a reason you think you'd need both addresses?

src/XRP/model/xrp-transaction.ts
```diff
@@ -162,12 +155,10 @@ export default class XRPTransaction {
 
   /**
    * @param hash The identifying hash of the transaction.
-   * @param account The unique address of the account that initiated the transaction.
+   * @deprecated @param account Deprecated, please use sourceXAddress. The unique address of the account that initiated the transaction.
```
hbergren, line 158: Misc: This is why when you start to get over ~3 parameters, it often makes sense to switch the call signature to a single parameter that is an object of properties.

Because right now, you have a deprecated parameter in the constructor function that _must_ be provided, because the order of parameters matters. If we used an object here, we could just omit the `account` property.

PR #493, opened 2020-05-31
(6/17) Creates `XRPDepositPreauth` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPDepositPreauth` - a native Typescript class to wrap the [`DepositPreauth` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L159).  See also https://xrpl.org/depositpreauth.html

### Context of Change

### Type of Change

## Before / After
`XRPDepositPreauth` class and conversion unit tests now exist.

## Test Plan

## Future Tasks


src/XRP/model/xrp-deposit-preauth.ts
```diff
@@ -0,0 +1,38 @@
+import { DepositPreauth } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
+
+/*
+ * Represents a DepositPreauth transaction on the XRP Ledger.
+ * A DepositPreauth transaction gives another account pre-approval to deliver payments to the sender of this transaction.
```
keefertaylor, line 5: +1 line

test/XRP/xrp-transaction-type-proto-conversion-test.ts
```diff
@@ -236,4 +239,38 @@ describe('Protobuf Conversions - Transaction Types', function (): void {
     // THEN the result is undefined.
     assert.isUndefined(checkCreate)
   })
+
+  it('Convert DepositPreauth protobuf to XRPDepositPreauth object - authorize set', function (): void {
+    // GIVEN a DespoitPreauth protocol buffer with authorize field set.
```
keefertaylor, line 244: s/DespoitPreauth/DepositPreauth

keefertaylor:
Technically, both of these fields can't be set at the same time. You could add a test for this, but I'm not sure we've enforced that elsewhere and at some point we should trust the remote server to not be giving us bogus data. 

Open to whether you think this is an actual problem or not.

PR #501, opened 2020-06-01
is/transaction memos

theotherian:

Add support for transaction memos

### Context of Change

We actually had an object in place already for the memos, it just wasn't being wired in. Also needed to add some stuff due to the restriction of using hex strings since it seems unlikely that passing that burden on to the client would be welcome.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

- Before: memos weren't addable to the transaction
- After: memos can be added and created with hex and non-hex strings

## Test Plan

Added unit tests for the new functionality; may explore trying to add an integration test as well.


src/Common/utils.ts
```diff
@@ -1,5 +1,38 @@
+import { Utils } from 'xpring-common-js/build/src'
```
keefertaylor, line 1: `import { Utils } from 'xpring-common-js`

No need to specify the exact path because this is exported from 'index.ts'

src/Common/utils.ts
```diff
@@ -1,5 +1,38 @@
...
 
+/**
+ * Converts a string such that each of its characters are represented as hex.
+ *
+ * @param value - the string to convert to hex.
+ * @returns the hex encoded version of the string.
+ */
+export const stringToHex = (value: string): string => {
+  let result = ''
+  // eslint-disable-next-line no-plusplus
```
keefertaylor, line 15: @hbergren To explain if this is correct or not?

keefertaylor: I suspect there's an easy way to use `Buffer` to do this without the ned for a loop?

src/Common/utils.ts
```diff
@@ -1,5 +1,38 @@
...
+}
+
+/**
+ * Converts a string that is optionally in hex format into a Uint8Array.
+ *
+ * @param value - the string to convert to a Uint8Array.
+ * @param isHex - flag to indicate it's a hex string or not.
+ * @returns the Uint8Array value.
+ */
+export const stringToUint8Array = (
```
keefertaylor, line 29: There's already a method called toBytes() in `Utils`? 

Maybe just `asciiToBytes` and you can drop the `isHex` param?

src/XRP/default-xrp-client.ts
```diff
@@ -136,12 +141,14 @@ class DefaultXRPClient implements XRPClientDecorator {
...
    * @param destinationAddress A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @param memos An optional list of memos to add to the transaction.
    * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   public async send(
     drops: BigInteger | number | string,
     destinationAddress: string,
     sender: Wallet,
+    memos?: XRPMemo[],
```
keefertaylor, line 151: nit: we've generally preferred the syntax:
```
memos?: Array<XRPMemo>
```

keefertaylor, line 151: I understand the XRP API is complex. Lets assume that XpringSDK is going to (imminently) expose a call that will let power users craft and send custom transactions. 

Would it make sense to make this API simpler? Most users will want a single memo as a string? Why not expose this parameter as `memo?: string?` rather than forcing external clients to understand what `XRPMemo` is? 

(If there's a legit answwer to this question I'm happy to hear it)

src/XRP/model/xrp-memo.ts
```diff
@@ -1,4 +1,10 @@
 import { Memo } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
+import { stringToUint8Array } from '../../Common/utils'
+
+export interface MemoField {
```
keefertaylor, line 4: can you doc this? It's not immediately obvious what the use is.

src/XRP/model/xrp-memo.ts
```diff
@@ -20,6 +26,48 @@ export default class XRPMemo {
...
+  /**
+   * Constructs an XRPMemo from raw values that represent the fields of a memo.
+   *
+   * @param data an optional Uint8Array for the data in the memo.
+   * @param format an optional Uint8Array for the format in the memo.
+   * @param type an optional Uint8Array for the type in the memo.
+   * @returns an XRPMemo with its fields set via the arguments matching the names of said fields.
+   * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L80
+   */
+  public static fromRaw(
```
keefertaylor, line 38: I think you can just use the `constructor` on line 79? No reason that we can't make that public.

test/XRP/default-xrp-client-test.ts
```diff
@@ -274,6 +274,31 @@ describe('Default XRP Client', function (): void {
...
+    // WHEN the account makes a transaction.
+    const transactionHash = await xrpClient.send(
+      amount,
+      destinationAddress,
+      wallet,
+      memos,
+    )
+
+    // THEN the transaction hash exists and is the expected hash
+    const expectedTransactionHash = Utils.toHex(
```
keefertaylor, line 294: I'd throw the expected vars under `GIVEN`, unless you have strong opinions about it being here. 

That way the test reads as:
GIVEN x
and an expected output from X

WHEN foo

THEN bar

src/Common/utils.ts
hbergren: Yup, we should be able to do something like:

```ts
Buffer.from(value).toString('hex')
```

But if it's that easy, I wonder if we need the `stringToHex` helper at all.

src/Common/utils.ts
theotherian: Ah, sorry, I think the IDE did this on its own? I didn't type that out myself; will fix.

src/Common/utils.ts
theotherian: I had followed an example on stackoverflow that was probably out of date. I can remove the helper if you prefer.

src/Common/utils.ts
theotherian: Maybe this is just a point of confusion for me, but the fact that the memo only allows for hex strings was something that I thought made the case for more flexibility in the API since it's not obvious if the string that was passed in is already a hex string. Maybe I'm just being dumb though.

src/XRP/model/xrp-memo.ts
theotherian: Sho thang.

src/XRP/default-xrp-client.ts
theotherian: The `Memo` is more than one field though (it's data, format and type) and it supports an arbitrarily sized array of memos as well. Maybe only attaching one memo is a reasonable assertion, but I honestly don't have enough context to know if saying that only specifying the `data` field is; I know we plan on specifying the format so that we know what to look for with our Open Payments ledger scanning. 

I think it maybe bite us to try to simplify something aggressively when we already have a use case that requires more fields than what you're suggesting, and I don't think we want to pass in multiple fields to describe one memo either.

src/XRP/model/xrp-memo.ts
theotherian: I just figured if we were already following a factory pattern for construction it was worth keeping. Given the relative incompatible yet similar arguments to construct an instance, my gut was that a factory method was more descriptive about the intention than a raw constructor.

src/XRP/default-xrp-client.ts
theotherian: Addressed this for the next commit.

src/XRP/xrp-client.ts
```diff
@@ -64,17 +65,19 @@ class XRPClient implements XRPClientInterface {
...
    * @param destination A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @param memos An optional list of memos to add to the transaction.
    * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   public async send(
     amount: BigInteger | number | string,
     destination: string,
     sender: Wallet,
+    memos?: Array<XRPMemo>,
```
Stormtv, line 78: same question about formatting as above

src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -31,6 +32,7 @@ class ReliableSubmissionXRPClient implements XRPClientDecorator {
     amount: string | number | BigInteger,
     destination: string,
     sender: Wallet,
+    memos?: Array<XRPMemo>,
```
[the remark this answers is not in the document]
amiecorso: It's my understanding that we do prefer `Array<...>`, @keefertaylor can correct me if not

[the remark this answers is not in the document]
theotherian: https://github.com/xpring-eng/Xpring-JS/pull/501#discussion_r434063176

src/XRP/xrp-client-decorator.ts
```diff
@@ -29,15 +30,17 @@ export interface XRPClientDecorator {
...
    * @param destination A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @param memos An optional list of memos to add to the transaction.
    * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   send(
     amount: BigInteger | number | string,
     destination: string,
     sender: Wallet,
+    memos?: Array<XRPMemo>,
```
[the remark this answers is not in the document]
theotherian: https://github.com/xpring-eng/Xpring-JS/pull/501#discussion_r434063176

src/XRP/xrp-client-interface.ts
```diff
@@ -33,15 +34,17 @@ export default interface XRPClientInterface {
...
    * @param destination A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
+   * @param memos An optional list of memos to add to the transaction.
    * @returns A promise which resolves to a string representing the hash of the submitted transaction.
    */
   send(
     amount: BigInteger | number | string,
     destination: string,
     sender: Wallet,
+    memos?: Array<XRPMemo>,
```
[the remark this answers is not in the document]
theotherian: https://github.com/xpring-eng/Xpring-JS/pull/501#discussion_r434063176

src/XRP/xrp-client.ts
theotherian: https://github.com/xpring-eng/Xpring-JS/pull/501#discussion_r434063176

src/XRP/model/xrp-memo.ts
theotherian: I went back and removed this actually.

theotherian:
@Stormtv I think there's supposed to be a restriction of 1024 characters; I'll go back and check.

theotherian:
Actually @Stormtv, in looking at the docs, this is the limit:

https://xrpl.org/transaction-common-fields.html#memos-field

> The Memos field is limited to no more than 1KB in size (when serialized in binary format).

It's not quote as straightforward as just a length check as you have to check all the fields combined as a serialized value. I'd probably rely on the underlying ledger to appropriately reject here personally since we're plugged into the gRPC stack at that point.

Stormtv:
@theotherian I do think we should codify the character constraint in here but I understand the complexity of checking the size of the serialized memos fields.

`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,;=%`

keefertaylor:
@theotherian are talking about API changes offline - please LMK if reviewers want to be involved.

src/Common/utils.ts
keefertaylor: SGTM

keefertaylor:
After talking with Ian I am happy letting this go in. I think it's pragmatic for us to allow fokls to have this feature and we can figure out if we want a better API after we use it a bit.  The optional parameter makes this a non-breaking change.

theotherian:
@keefertaylor changelog updated, @LoisRP would you mind reviewing for tech doc stuff on the changelog file?

PR #502, opened 2020-06-01
Adds PayID-Version header to PayIDClient

tedkalaw:
This hardcodes the `PayID-Version` header to be `1.0`, per the breaking changes in https://docs.google.com/document/d/1EnOJfpxr4PyM1dMKfYpTeCph2om3F8vZYc4S_PqjNSI/edit#. As is, `xpring-js` does not work for `PayID` `1.0.0` (because it doesn't pass `PayID-Version`) and it does not work for `PayID` `0.1.0` (because it expects to get back an array of `addresses`).

I can update this to take a version if we want but I wanted to go for the smallest change so we can cut a release.

### Context of Change
https://docs.google.com/document/d/1EnOJfpxr4PyM1dMKfYpTeCph2om3F8vZYc4S_PqjNSI/edit#

https://github.com/xpring-eng/Xpring-JS/pull/474/ updates `PayIDClient` to handle the new payload. We still need to support versioning

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`PayIDClient.addressForPayID` works against `PayID 1.0.0`

## Test Plan
CI

### DIFF
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -306,6 +306,7 @@ export default class PayIDClient {
   private static makeOptionsWithAcceptTypes(acceptType: string): object {
     const headers = {
       Accept: acceptType,
+      'PayID-Version': 1.0,
     }
 
     const options = {
--- a/src/PayID/pay-id-client.ts
+++ b/src/PayID/pay-id-client.ts
@@ -13,6 +13,8 @@ import {
   Compliance,
 } from './Generated/api'
 
+const PAYID_API_VERSION = '1.0'
+
 /* eslint-disable */
 
 interface PayIDComponents {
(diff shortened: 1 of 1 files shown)


tedkalaw:
check out the circle CI runs for before and after. that wasn't on purpose, but it shows the change. i'm guessing all xpring-js builds will fail for now

src/PayID/pay-id-client.ts
```diff
@@ -306,6 +306,7 @@ export default class PayIDClient {
   private static makeOptionsWithAcceptTypes(acceptType: string): object {
     const headers = {
       Accept: acceptType,
+      'PayID-Version': '1.0',
```
keefertaylor, line 309: Does the Key need to be quoted? 

Can you refactor 1.0 to a top level variable?

hbergren: > Can you refactor 1.0 to a top level variable?

Why would that be better?

tedkalaw: yeah, keys of js objects that have `-` need to be in quotes. if they have `_`, on the other hand, they don't

i'll do the top level var

PR #505, opened 2020-06-03
(10/17) Creates `XRPOfferCancel` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPOfferCancel` - a native Typescript class to wrap the [`OfferCancel` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L206).  See also https://xrpl.org/offercancel.html

### Context of Change

### Type of Change

## Before / After
`XRPOfferCancel` class and conversion unit tests now exist.

## Test Plan

## Future Tasks


src/XRP/model/xrp-offer-cancel.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+    return new XRPOfferCancel(offerSequence)
+  }
+
+  /**
+   * @param offerSequence The sequence number of a previous OfferCreate transaction. If specified,
+   *                      cancel any offer object in the ledger that was created by that transaction.
+   *                      It is not considered an error if the offer specified does not exist.
+   */
+  private constructor(readonly offerSequence?: number) {}
```
keefertaylor, line 31: same comment about this being required

PR #507, opened 2020-06-04
(11/17) Creates `XRPOfferCreate` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPOfferCancel` - a native Typescript class to wrap the [`OfferCreate` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L212).  See also https://xrpl.org/offercreate.html

### Context of Change

### Type of Change

## Before / After
`XRPOfferCreate` class and conversion unit tests now exist.

## Test Plan

## Future Tasks


src/XRP/model/xrp-offer-create.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * @param expiration (Optional) Time after which the offer is no longer active, in seconds since the Ripple Epoch.
+   * @param offerSequence (Optional) An offer to delete first, specified in the same way as OfferCancel.
+   * @param takerGets The amount and type of currency being provided by the offer creator.
+   * @param takerPays The amount and type of currency being requested by the offer creator.
+   */
+  private constructor(
+    readonly expiration?: number,
+    readonly offerSequence?: number,
+    readonly takerGets?: XRPCurrencyAmount,
```
keefertaylor, line 57: If these are required they shouldn't be optional

keefertaylor:
I think this looks reasonably tested (and so optional) but if you're curious why only 66% of your diff was hit you can see the missed lines here: 
https://codecov.io/gh/xpring-eng/Xpring-JS/compare/e3031a4c77b5cac56c87c2f233f1cb7bdfa9e0e8...26853c18ee0d68d4f7f89b53c98c0d23c7f33cd7/diff

PR #519, opened 2020-06-09
Xpring eslint prereqs

hbergren:

Various pre-reqs required to actually start using `@xpring-eslint`.

This basically just upgrades our linting dependencies to the latest versions, and either disables rules that are now newly enabled, or fixes the linting violations associated with upgrading the dependencies.

### Context of Change

I was hacking on linting the tests here using `@xpring-eslint`, when [my build failed...](https://app.circleci.com/pipelines/github/xpring-eng/Xpring-JS/2564/workflows/c41d39ec-a20f-4c0b-b292-5b3d733669bc/jobs/2645/steps) because ESLint v6 doesn't know about `/* eslint disable -- directive comments */`, which is lame.

So I just upgraded everything.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

We now have new versions of all our linting dependencies.

## Test Plan

`npm run test` is green so I can't have broken anything too seriously.


.eslintrc.js
```diff
@@ -86,6 +86,15 @@ module.exports = {
...
         ],
+
+        // These rules are deprecated, but we have an old config that enables it
+        '@typescript-eslint/camelcase': 'off',
+        '@typescript-eslint/ban-ts-ignore': 'off',
+
+        // These rules are actually disabled in @xpring-eng/eslint-config-base/loose at the moment
+        '@typescript-eslint/no-unsafe-call': 'off',
+        '@typescript-eslint/no-unsafe-member-access': 'off',
+        '@typescript-eslint/no-unsafe-assignment': 'off',
```
hbergren, line 97: These are rules that are now enabled by default in `@typescript-eslint >= 3.0.0`, but cause various warnings. I don't even have them enabled in the default `@xpring-eslint/loose` configuration yet (but might do before I tag 1.0).

package.json
```diff
@@ -56,7 +57,7 @@
     "tslint-config-prettier": "^1.18.0",
     "tslint-config-standard": "^9.0.0",
     "tslint-eslint-rules": "^5.4.0",
-    "typescript": "^3.9.2",
+    "typescript": "^3.9.5",
```
hbergren, line 60: I needed to upgrade all the dependencies because my actual linting PR was failing CI because ESLint v6 doesn't know about `/* eslint-disable -- directive comments */`

src/XRP/raw-transaction-status.ts
```diff
@@ -30,7 +30,10 @@ export default class RawTransactionStatus {
     return new RawTransactionStatus(
       getTransactionResponse.getValidated(),
       getTransactionResponse.getMeta()?.getTransactionResult()?.getResult(),
-      getTransactionResponse.getTransaction()?.getLastLedgerSequence(),
+      getTransactionResponse
+        .getTransaction()
+        ?.getLastLedgerSequence()
+        ?.getValue(),
```
hbergren, line 36: I _think_ this is the right thing to do here?

src/XRP/raw-transaction-status.ts
```diff
@@ -39,9 +42,9 @@ export default class RawTransactionStatus {
    * Note: This constructor is exposed for testing purposes. Clients of this code should favor using a static factory method.
    */
   constructor(
-    public isValidated,
-    public transactionStatusCode,
-    public lastLedgerSequence,
-    public isFullPayment,
+    public isValidated: boolean,
+    public transactionStatusCode: string | undefined,
+    public lastLedgerSequence: number | undefined,
```
hbergren, line 47: I wasn't sure about the typing on this one. It was either `number | undefined` or `LastLedgerSequence | undefined`.

If I made it `LastLedgerSequence`, then all the tests where we pass an actual number start failing due to type incompatibility (and some places in source code where we do a numeric comparison).

Thoughts?

src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -57,7 +57,7 @@ class ReliableSubmissionXRPClient implements XRPClientDecorator {
       transactionHash,
     )
     const { lastLedgerSequence } = rawTransactionStatus
-    if (lastLedgerSequence === 0) {
+    if (!lastLedgerSequence) {
```
hbergren, line 60: This is `number | undefined`, so I switched the check so it also type guards against `undefined`.

test/XRP/fakes/fake-xrp-protobufs.ts
```diff
@@ -405,9 +405,7 @@ testInvalidGetAccountTransactionHistoryResponse.setTransactionsList(
 // XRP OBJECTS ===================================================
 
 // test XRPTransaction
-const testXRPTransaction = XRPTransaction.from(
-  testGetTransactionResponseProto,
-)!!
+const testXRPTransaction = XRPTransaction.from(testGetTransactionResponseProto)!
```
hbergren, line 408: Remove an unnecessary extra non-null assertion.

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -1,3 +1,8 @@
+/* eslint-disable @typescript-eslint/no-non-null-asserted-optional-chain --
+ * This rule should never be disabled in production, but for tests,
+ * having a x?.y?.z! chain is a succinct way of getting the typing correct.
+ */
```
hbergren, line 4: I really don't know about disabling this rule. I wonder if some constructors need to have their types updated?

An example from line 211 would be:
```ts
XRPCurrencyAmount.from(
  testPaymentProtoAllFieldsSet.getAmount()?.getValue()!,	        
)
```

package.json
```diff
@@ -20,52 +20,42 @@
...
-    "typescript-eslint-parser": "22.0.0"
+    "typescript": "^3.9.3"
   },
   "dependencies": {
     "axios": "^0.19.2",
     "big-integer": "^1.6.48",
     "bignumber.js": "^9.0.0",
     "grpc": "^1.24.2",
     "grpc-web": "1.1.0",
-    "superagent": "^5.2.2",
```
hbergren, line 68: I just went ahead and removed all the dependencies that weren't actually being used.

src/XRP/raw-transaction-status.ts
keefertaylor: Yup, it is.

src/XRP/raw-transaction-status.ts
keefertaylor: It should be a number.

keefertaylor: TBH we might be able to remove this class. I believe it is just a bridge that was used for the legacy protocol buffers. 

Can you add a TODO for me at the class level to remove this?

test/XRP/xrp-protocol-buffer-conversion.test.ts
keefertaylor: I think this is reasonable to do. It's only test code and the protocol buffers tend to be pretty verbose if you do it the 'right' way.

src/XRP/raw-transaction-status.ts
hbergren: It was already there (but I changed the syntax a little so it turns highlighted yellow for me using this [TODO Highlight extension](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight).

PR #520, opened 2020-06-09
Rename PayID Code Idiomatically

keefertaylor:

JavaScript prefers to use `PayId` to `PayID`. This is a large one shot PR to fix that. 

### Context of Change

This is a large in place refactor. We rely on our tests to keep us honest that functionality has not changed. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Closer to idiomatic and consistent naming. 

## Test Plan

CI


src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+    payID: string,
+    nonce: string,
+  ): Promise<SignatureWrapperInvoice> {
+    return this.wrappedPayIdClient.getInvoice(payID, nonce)
+  }
+
+  /**
+   * Post an invoice with compliance data.
+   *
+   * @param payID The Pay ID compliance data is associated with.
```
LoisRP, line 85: Pay ID ->PayID

src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+   */
+  async getInvoice(
+    payID: string,
+    nonce: string,
+  ): Promise<SignatureWrapperInvoice> {
+    return this.wrappedPayIdClient.getInvoice(payID, nonce)
+  }
+
+  /**
+   * Post an invoice with compliance data.
```
LoisRP, line 83: Are we still using invoice now?

src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+   * @returns An address representing the given PayID.
+   */
+  async addressForPayID(payID: string): Promise<CryptoAddressDetails> {
+    return this.wrappedPayIdClient.addressForPayId(payID)
+  }
+
+  /**
+   * Generate a new invoice with compliance requests.
+   *
+   * @param payID The Pay ID to request an invoice for.
```
LoisRP, line 72: Pay ID -> PayID (throughout)

src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+   * @param nonce A randomly selected nonce that is unique to this invoice request.
```
LoisRP, line 73: For some reason, I thought nonce was no longer going to be used? I realize that's outside the scope of this review.

src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+   * @param originatorUserLegalName The legal name of the originator.
+   * @param originatorAccountID The account ID of the originator.
+   * @param originatorUserPhysicalAddress The physical address of the originator.
+   * @param originatorInstitutionName The institution name of the originator.
+   * @param valueAmount The value being transferred.
+   * @param valueScale The scale of the value.
+   * @param timestamp The time that the operation occurred.
+   * @param beneficiaryInstitutionName The beneficiary insitution name.
+   * @param beneficiaryUserLegalName The legal name of the receiver at the beneficiary institution. Optional, defaults to undefined.
+   * @param beneficiaryUserPhysicalAddress The phsyical address of the receiver at the beneficiary institution. Optional, defaults to undefined.
```
LoisRP, line 99: physical - typo

src/PayID/pay-id-client.ts
```diff
@@ -43,6 +52,140 @@ export default class PayIDClient {
...
+   * @param publicKeyType The type of public key.
+   * @param publicKeyData An array of public keys which lead back to the root trust certificate.
+   * @param publicKey The public key.
+   * @param signature The signature of the operation.
+   * @param originatorUserLegalName The legal name of the originator.
+   * @param originatorAccountID The account ID of the originator.
+   * @param originatorUserPhysicalAddress The physical address of the originator.
+   * @param originatorInstitutionName The institution name of the originator.
+   * @param valueAmount The value being transferred.
+   * @param valueScale The scale of the value.
```
LoisRP, line 95: The scale of the value. - Is it clear what this is, or what the value means?

keefertaylor: Ah, I think I should say `valueAmount`? 

Happy to hear other revisions to this documentation to. 

The idea is that $1 USD at scale 2 is 100. Or $1.23USD is 123 at scale 2. Scale is the number of significant decimal places? 

This is terminology that is borrowed from ILP. All their APIs take a value and a scale. Maybe they've found a better way to describe it?

src/PayID/pay-id-client.ts
keefertaylor: Same comment here. Queried the PayID folks.

src/PayID/pay-id-client.ts
keefertaylor: AFAIK we are but I could be out of the loop. 

I queried the PayID folks and I'll leave this PR open until I get a definitive answer.

hbergren: This is `paymentSetupDetails` now... but Travel Rule is pretty faked and not real, so do what you will.

src/PayID/pay-id-client.ts
hbergren: Yeah, we're not using `nonce` anymore. I think the field we use now is `txId`, which gets provided by the PayID server on a `GET /payment-setup-details`, and which the client should provide for a `POST /payment-setup-details` and `POST /payment-proof`.

src/PayID/pay-id-client.ts
hbergren: I think XRPL uses the same notion of an amount and scale.

PR #529, opened 2020-06-10
Rename internal XRP components

keefertaylor:

Rename non-exported XRP Components idiomatically. JavaScript prefers `Xrp` to `XRP`. 

### Context of Change

Since these are all internal components, we can make this change without a deprecation cycle to keep developers lives easier. 

CHANGELOG.md is updated as a matter or record, though these shouldn't be breaking changes. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

More idiomatic code is used. 

## Test Plan

CI


src/PayID/xrp-pay-id-client-interface.ts
```diff
@@ -19,3 +21,20 @@ export default interface XRPPayIDClientInterface {
...
+   * @param network The network that addresses will be resolved on.
+   */
+  xrplNetwork: XrplNetwork
+
+  /**
+   * Retrieve the XRP Address associated with a PayID.
+   *
+   * @note The returned value will always be in an X-Address format.
+   *
+   * @param payId The payID to resolve for an address.
```
LoisRP, line 36: Obviously the parameter `payId` uses camelcase. Do you want to duplicate that with "The payId to ...", or should it be "The PayID to ..."?

src/PayID/xrp-pay-id-client.ts
```diff
@@ -22,35 +26,85 @@ export default class XRPPayIDClient extends PayIDClient
...
+    }
+  }
+
+  /**
+   * Retrieve the XRP address associated with a PayID.
+   *
+   * Note: Addresses are always in the X-Address format.
+   * @see https://xrpaddress.info/
+   *
+   * @param payID The payID to resolve for an address.
```
LoisRP, line 104: What casing are you using for the param and the description?

src/PayID/pay-id-client.ts
```diff
@@ -96,7 +96,11 @@ export class PayIDClient {
    * @param timestamp The time that the operation occurred.
    * @param beneficiaryInstitutionName The beneficiary insitution name.
    * @param beneficiaryUserLegalName The legal name of the receiver at the beneficiary institution. Optional, defaults to undefined.
+<<<<<<< HEAD
```
amiecorso, line 99: unresolved merge situation?

src/PayID/xrp-pay-id-client-interface.ts
keefertaylor: Yeah, sorry this is a mess. We wrote a *lot* of code before this was all agreed upon. 

In code, we want to camel case things, so we'll always use "PayId" for consistency. This matches up with our other code. 

In docs, we should always call it a "PayID" since that is the noun we've landed on. I'll make this change.

PR #534, opened 2020-06-11
Rename XRP/Models

keefertaylor:

Rename classes in `XRP/models`. JavaScript prefers `Xrp` to `XRP`. 

### Context of Change

Deprecate the existing classes and add a new idiomatically cased classes. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

More idiomatic code is used. 

## Test Plan

CI


keefertaylor:
Apologies for the bad code coverage score on this PR. 

My sense here is that since we'll remove this in N+2 releases (1 month) it isn't worth duplicating all the tests to game a coverage score. LMK if reviewers feel strongly otherwise.

src/XRP/model/xrp-path.ts
```diff
@@ -31,3 +33,32 @@ export default class XRPPath {
...
+export default class XrpPath {
+  /**
+   * Constructs an XrpPath from a Path.
+   *
+   * @param path a Path (protobuf object) whose field values will be used
+   *             to construct an XrpPath
+   * @returns an XrpPath with its fields set via the analogous protobuf fields.
+   * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L237
+   */
+  public static from(path: Path): XRPPath {
```
amiecorso, line 50: this return type should be `XrpPath` ?

src/XRP/model/xrp-path-element.ts
```diff
@@ -43,3 +45,44 @@ export default class XRPPathElement {
...
+   *
+   * @param pathElement a PathElement (protobuf object) whose field values will be used
+   *                    to construct an XrpPathElement
+   * @returns an XrpPathElement with its fields set via the analogous protobuf fields.
+   * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L227
+   */
+  public static from(pathElement: PathElement): XrpPathElement {
+    const account = pathElement.getAccount()?.getAddress()
+    const currency = pathElement.getCurrency()
+    const xrpCurrency = currency && XRPCurrency.from(currency)
```
amiecorso, line 65: should the idiomatically-cased classes be sure to use idiomatically-cased members?  i.e. this should be `XrpCurrency` ?

amiecorso: I suppose we'll catch any of these stragglers when we remove the deprecated code and some things are broken

src/XRP/model/xrp-payment.ts
```diff
@@ -103,3 +105,103 @@ export default class XRPPayment {
...
+      paymentDeliverMinValue && XrpCurrencyAmount.from(paymentDeliverMinValue)
+    if (paymentDeliverMinValue && !deliverMin) {
+      return undefined
+    }
+
+    const invoiceID = payment.getInvoiceId()?.getValue_asU8()
+
+    const paths =
+      payment.getPathsList()?.length > 0
+        ? payment.getPathsList().map((path) => XRPPath.from(path))
```
amiecorso, line 163:
```suggestion
        ? payment.getPathsList().map((path) => XrpPath.from(path))
```

src/XRP/model/xrp-transaction.ts
```diff
@@ -208,3 +210,204 @@ export default class XRPTransaction {
...
+      case Transaction.TransactionDataCase.PAYMENT: {
+        const payment = transaction.getPayment()
+        if (!payment) {
+          return undefined
+        }
+        paymentFields = payment && XrpPayment.from(payment, xrplNetwork)
+        if (!paymentFields) {
+          return undefined
+        }
+        type = payment && XRPTransactionType.Payment
```
amiecorso, line 293:
```suggestion
        type = payment && XrpTransactionType.Payment
```

src/XRP/xrp-client-decorator.ts
```diff
@@ -2,7 +2,7 @@ import { Wallet } from 'xpring-common-js'
 import { BigInteger } from 'big-integer'
 import TransactionStatus from './transaction-status'
 import RawTransactionStatus from './raw-transaction-status'
-import XRPTransaction from './model/xrp-transaction'
+import { XRPTransaction } from './model/xrp-transaction'
```
amiecorso, line 5: Are we just going to wait until deleting the old code to replace the references to the old class names?

src/XRP/model/xrp-path.ts
keefertaylor: I can't believe that compiles...

src/XRP/xrp-client-decorator.ts
keefertaylor: See #539. 

I think I have to deprecate / migrate some more XRP classes than initially intended.

PR #535, opened 2020-06-11
(1/x) Create `SingleNetworkPayIdClient`

keefertaylor:

Rename `PayIdClient` to `SingleNetworkPayIdClient` in a non-breaking way.

### Context of Change

When we started developing PayID it was only possible to receive one address for a given request. Now, it is possible that you can retrieve all addresses. 

This is the first step in a migration to make that possible. 

`SingeAddressPayIdClient` -> The new `PayIdClient`. Gets a single address for a single network
`PayIdClient` -> Changes to a private `PayIdClientImpl` and returns arrays of addresses
`AllAddressesPayIdClient` -> Returns all addresses for a PayID

Both `SingleAddresspayIdClient` and `AllAddressesPayIdClient` wrap a `PayIdClientImpl` and interpret results. 

I dislike these names but don't have many better ideas. Open to suggestions from reviewers. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

CI. I duplicated the tests to make sure we keep test coverage.


keefertaylor:
@amiecorso 

> why not have a sort of generic PayIdClient that has several different methods (i.e. for finding a single address for a specific network like, and another for finding all addresses associated with a PayID)

I guess there's no reason. It feels odd to me to have a client bound to a network and then for that class to also be able to resolve on all networks. 

That said, the reason `PayIdClient` is bound to a network was so that it worked with `XpringClient`. Nowadays, `XrpClient` works with `XrpPayIdClient` so I guess there's no reason that `PayIdClient` needs to be bound to a network. 

I think this architecture is building under the assumption that we wanted to maintain a 1:1 mapping. If that isn't the case (and there's no reason it should be) then we could also use your proposed architecture. 

>  - where you can then also have network-specific payID clients like XrpPayIdClient. What's the benefit of having an entire separate client class for the find-all functionality?

None, really. I think it's kinda confusing TBH.

src/PayID/single-network-pay-id-client.ts
```diff
@@ -0,0 +1,341 @@
...
+   * Networks in this constructor take the form of an asset and an optional network (<asset>-<network>), for instance:
+   * - xrpl-testnet
+   * - xrpl-mainnet
+   * - eth-rinkeby
+   * - ach
+   *
+   * TODO(keefertaylor): Link a canonical list at payid.org when available.
+   *
+   * @param network The network that addresses will be resolved on.
+   * @param useHttps Whether to cuse HTTPS when making PayID requests. Most users should set this to 'true' to avoid
```
austin-king, line 48: appreciate the helpful docs

src/PayID/single-network-pay-id-client.ts
```diff
@@ -0,0 +1,341 @@
...
+   * @param payId The payID to resolve for an address.
+   * @returns An address representing the given PayID.
+   */
+  async addressForPayId(payId: string): Promise<CryptoAddressDetails> {
+    const payIdComponents = SingleNetworkPayIdClient.parsePayId(payId)
+    const basePath = this.useHttps
+      ? `https://${payIdComponents.host}`
+      : `http://${payIdComponents.host}`
+
+    // Swagger API adds the leading '/' in path automatically because it is part of the endpoint.
```
austin-king, line 68: also v helpful comment

src/PayID/single-network-pay-id-client.ts
```diff
@@ -0,0 +1,341 @@
...
+        )
+      }
+    }
+  }
+
+  /**
+   * Generate a new invoice with compliance requests.
+   *
+   * @param payId The Pay ID to request an invoice for.
+   * @param nonce A randomly selected nonce that is unique to this invoice request.
```
austin-king, line 121: PayID no longer accepts a nonce from the client, the server returns a `txId` instead. This is because we were essentially giving the client authority to determine the identifier for something the server would need to store. 

On top of that this all might get tossed out the window in favor of the TRISA stuff lol

test/PayID/single-network-pay-id-client.test.ts
```diff
@@ -0,0 +1,434 @@
...
+    const singleNetworkPayIdClient = new SingleNetworkPayIdClient(
+      XrplNetwork.Test,
+    )
+
+    const payIDComponents = PayIdUtils.parsePayID(payId)
+    if (!payIDComponents) {
+      throw new Error('Test precondition failed: Could not generate a Pay ID')
+    }
+    nock('https://xpring.money')
+      .get('/georgewashington')
```
austin-king, line 63: my man george making another appearance

keefertaylor:
Closing in favor of #540

PR #540, opened 2020-06-12
Add Fetch All Addresses Feature and Unbound `PayIdClient` from a specific network

keefertaylor:

Adds new functionality for retrieving all addresses. 

### Context of Change

This change adds a new method to retrieve all addresses for a PayID. Unit / integration tests are also provided. 

There's a larger change here where `PayIdClients` are no longer bound to a network. This is because it doesn't make logical sense for a `PayIDClient` bound to a BTC network to be able to read XRP addresses. As such, there's a few deprecations that occur in `PayIdClient`. I can unbundle these if reviewers would like. This change was discussed with the PayID-Eng slack channel. 

Admittedly, we're pretty messy with deprecations for PayID at the moment, but we should be able to clear this up in a few releases. 

Update CHANGELOG and README.md. Note that #546 will provide the majority of README.md and this will just be an incremental change when they get merged together. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No breaking changes - but we're definitely changing course a bit before launch. 

## Test Plan

CI - new tests are provided (unit + integration)


keefertaylor:
It’s odd that this decreased coverage so noticeably. Need to take a brief commute but then I’ll investigate.

keefertaylor:
CodeCov looks like a false positive. Reporting large changes in:
- xrp-account-set.ts
- xrp-check-cancel.ts
- xrp-account-delete.ts

All of which were recently added and I think master hasn't caught up yet because the build flaked.

src/PayID/pay-id-client.ts
```diff
@@ -182,22 +183,87 @@ export default class PayIdClient {
...
+        new PayIdError(
+          PayIdErrorType.UnexpectedResponse,
+          'Received more addresses than expected',
+        ),
+      )
+    }
+  }
+
+  /**
+   * Retrieve the all addresses associated with a PayId.
```
LoisRP, line 249: Retrieve all

README.md
```diff
@@ -248,91 +280,118 @@ const senderWallet = generationResult.wallet;
...
+    balance.netBalance +
+    ' with asset scale ' +
+    balance.assetScale,
+)
 ```
 
 #### Sending a Payment
+
 An `IlpClient` can send an ILP payment to another ILP address by supplying a [Payment Pointer](https://github.com/interledger/rfcs/blob/master/0026-payment-pointers/0026-payment-pointers.md)
 and a sender's account ID
```
LoisRP, line 396: Add period.

test/PayID/pay-id-client.test.ts
```diff
@@ -191,6 +191,80 @@ describe('PayIdClient', function (): void {
     })
   })
 
+  it('allAddressesForPayId - invalid Pay ID', function (done): void {
+    // GIVEN an PayIdClient and an invalid PayID.
+    const invalidPayID = 'xpring.money/georgewashington' // Does not start with '$'
```
amiecorso, line 196: is it that it doesn't start with a '$'... or that it doesn't contain one?

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `XRPTransactionType` now includes an `AccountSetValue`.
 - `XrpTransactionType` replaces the now deprecated `XRPTransactionType`.
 - `XrpTransaction` replaces the now deprecated `XRPTransaction`.
+- A new method, `cryptoAddressForPayId`, replaces the now deprecated `addressForPayId` method in `PayIdClient`.
```
intelliot, line 35: What is wrong with the name `addressForPayId`? What's the benefit of adding `crypto` to the name of the method?

hbergren: Does it maybe assume a `CryptoAddressDetails` object? Or is this just so we can "deprecate" `addressForPayId`, without making a breaking change to an existing function?

Would we theoretically have `achAddressForPayId`, `sepaAddressForPayId`, `creditCardDetailsForPayId`?

README.md
```diff
@@ -65,107 +71,128 @@ and send payments over ILP.
...
 ##### Hierarchical Deterministic Wallets
+
 A hierarchical deterministic wallet is created using a mnemonic and a derivation path. Simply pass the mnemonic and derivation path to the wallet generation function. Note that you can omit the derivation path and have a default path be used instead.
 
 ```javascript
-const { Wallet } = require("xpring-js");
+const { Wallet } = require('xpring-js')
 
-const mnemonic = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
+const mnemonic =
```
hbergren, line 88: Did you run this through Prettier or something?

CHANGELOG.md
keefertaylor: It does assume crypto details. Also it takes a parameter. 

This PR is fundamentally changing this method.

README.md
keefertaylor: I think it just autosaved in VSCode and got formatted nicely.

PR #543, opened 2020-06-16
Renames `XRPAccountSet` to `XrpAccountSet`

amiecorso:
I merged the `XRPAccountSet` wrapper before systematically renaming all the other transaction type wrapper classes.  This PR just makes sure that `AccountSet` / `XrpAccountSet` conforms to the same convention.

### Type of Change

## Before / After
Correct casing in `XrpAccountSet` and references

## Test Plan
CI still passes


keefertaylor:
Technically, this is 'breaking' but we haven't cut a release since the original PR was merged.

PR #581, opened 2020-07-07
Merges `FakeGRPCError` into one class

amiecorso:
`FakeGRPCError` existed almost identically in two separate fakes folders under `XRP` and `ILP` components.  This PR merges `FakeGRPCError` into one class under `test/Common/Fakes` and makes necessary small refactors. 

### Context of Change
Post-launch tech debt cleanups/TODOs.

### Type of Change

## Before / After
One version of `FakeGRPCError`

## Test Plan
CI still passes


test/Common/Fakes/fake-grpc-error.ts
```diff
@@ -1,15 +1,17 @@
...
   /**
    * Construct a new FakeServiceError.
    *
    * @param message The text details of the error.
    * @param code The gRPC status code.
    */
-  constructor(public readonly message: string, public readonly code: number) {
+  constructor(
+    public readonly code: number,
+    public readonly message: string = '',
```
keefertaylor, line 13: This doesn't really matter, but I believe message is a property on the super class so when you make this `public readonly` it actually overrides that member. 

It's probably best to remove `public readonly` just in case it causes unexpected behavior in the future.

PR #584, opened 2020-07-09
Refactors Transaction Prep/Submission for Re-use

amiecorso:
In preparation to include additional transaction types in the `XrpClient`, this PR refactors the code for preparing and submitting transactions into private reusable helper functions.  

Specifically, 

`prepareBaseTransaction` constructs and returns a `Transaction` protobuf object that has been populated with the required and recommended common fields (https://xrpl.org/transaction-common-fields.html) necessary for all transaction types.  

`submitTransaction` takes a completed `Transaction` protobuf object and handles the signing and submitting process.

The logic for constructing and submitting an arbitrary transaction type would now be:
- prepare a `Transaction` using `prepareBaseTransaction`
- construct the transaction-specific details and assign to the `Transaction` object (https://xrpl.org/transaction-types.html)
- populate any desired, optional common fields such as memos.
- submit transaction using `submitTransaction`

### Type of Change

## Before / After
Common transaction submission steps are now reusable.

## Test Plan
CI still passes


src/XRP/default-xrp-client.ts
```diff
@@ -466,4 +425,70 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+   *
+   * @param wallet The wallet that will sign and submit this transaction.
+   * @returns A promise which resolves to a Transaction protobuf with the required common fields populated.
+   */
+  private async prepareBaseTransaction(wallet: Wallet): Promise<Transaction> {
+    const classicAddress = Utils.decodeXAddress(wallet.getAddress())
+    if (!classicAddress) {
+      throw XrpError.xAddressRequired
+    }
+    const fee = await this.getMinimumFee()
```
keefertaylor, line 443: Are fees on the ledger always the same?

keefertaylor, line 443: nit: 1 vertical line break after :442

src/XRP/default-xrp-client.ts
```diff
@@ -466,4 +425,70 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+    return transaction
+  }
+
+  /**
+   * Signs the provided Transaction object using the provided Wallet and submits to the XRPL network.
+   *
+   * @param wallet The wallet that will sign and submit this transaction.
+   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
+   */
+  private async submitTransaction(
```
keefertaylor, line 476: Optional: I know this sounds stupidly verbose, but `signAndSubmitTransaction` is maybe a better name?

src/XRP/default-xrp-client.ts
```diff
@@ -466,4 +425,70 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+    wallet: Wallet,
+    transaction: Transaction,
+  ): Promise<string> {
+    const signedTransaction = Signer.signTransaction(transaction, wallet)
+    if (!signedTransaction) {
+      throw XrpError.malformedResponse
```
keefertaylor, line 482: Not your fault but I think this should be `XrpError.signingError`

src/XRP/default-xrp-client.ts
```diff
@@ -466,4 +425,70 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+    transaction.setLastLedgerSequence(lastLedgerSequence)
+    transaction.setSigningPublicKey(signingPublicKey)
+
+    return transaction
+  }
+
+  /**
+   * Signs the provided Transaction object using the provided Wallet and submits to the XRPL network.
+   *
+   * @param wallet The wallet that will sign and submit this transaction.
```
keefertaylor, line 473: please doc `transaction`

src/XRP/default-xrp-client.ts
```diff
@@ -466,4 +425,70 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
+   */
+  private async submitTransaction(
+    wallet: Wallet,
+    transaction: Transaction,
```
keefertaylor, line 478: Optional / This is a very subjective comment: I'd consider reversing this parameter order so that it reads more logically:
"submit transaction `transaction` (from) `wallet`"

rather than:
"submit transaction (with) `wallet` (signing) `transaction`"

amiecorso: I agree with this

src/XRP/default-xrp-client.ts
amiecorso: doc'ed

src/XRP/default-xrp-client.ts
amiecorso: I had to add `signingError` to the `XrpError` class (it was already an option in the enum but needed to become a static member of the class).  Let me know if you agree with that and/or if you have any suggestions for improving the wording.

src/XRP/default-xrp-client.ts
amiecorso: This is definitely an important question... great catch.  I don't think they are.  Going to do some more investigating.

amiecorso: Per our offline discussion - this is the correct fee for the next transaction type we're adding (`AccountSet`).  From my notes: "there are four types of transactions that have different fee requirements: multi-signed transactions (which I suppose could be any sort of transaction because signers is a common field), EscrowFinish transactions (where the fee is proportional to the size of the fulfillment condition in bytes), key reset transaction (0 fee), and an AccountDelete transaction (flat fee of 5,000,000 drops)" -- we can leave fee population in the `prepareBaseTransaction` method and then overwrite it in the select few cases that require this (if we end up implementing them)

PR #585, opened 2020-07-09
(2/2) Implements `enableDepositAuth` on `XRPClient`

amiecorso:
This PR adds the method `enableDepositAuth` to the `XrpClient`, which enables [Deposit Authorization](https://xrpl.org/depositauth.html) on the specified account.  This is achieved by submitting an `AccountSet` transaction with a set flag set to 9 (see https://xrpl.org/accountset.html#accountset-flags).  An API for submitting generic `AccountSet` transactions is not yet exposed - this exposes only a specific piece of functionality.

The return type of `enableDepositAuth` is a `Promise<TransactionResult>`, where `TransactionResult` is a new class containing a `hash` and `status` field.  The implementation in `DefaultXrpClient` may return `TransactionStatus`es that are `.pending`, `.succeeded`, `.failed`, whereas the implementation in `ReliableSubmissionXrpClient` may return `TransactionStatus`es that are `.succeeded` or `.failed` but not `.pending`, as it awaits finality of transaction result (inclusion in a validated ledger OR passing of `lastLedgerSequence`) before returning. 

Note: I merged https://github.com/xpring-eng/Xpring-JS/pull/591 (reliable submission for `enableDepositAuth`) into this branch to create an atomic PR for `enableDepositAuth`, which is easier to understand.  Apologies for the large PR size... in the other SDKs I'm doing the flag refactors separately which makes them much smaller.

### Context of Change

The Charity App needs to be able to enable deposit authorization, and we would like it to rely on the SDK.  This is also serving as a preliminary attempt at reasoning about inclusion of new transaction types in a user-friendly way.

### Type of Change

- [x] Documentation Updates

## Before / After
`enableDepositAuth` exposed on `XrpClient`

## Test Plan
CI still passes.  New tests written for `enableDepositAuth`

## Future Tasks

Other two SDKs


src/XRP/default-xrp-client.ts
```diff
@@ -426,6 +428,27 @@ export default class DefaultXrpClient implements XrpClientDecorator {
     return XrpTransaction.from(getTransactionResponse, this.network)
   }
 
+  /**
+   * Enable Deposit Authorization for this XRPL account.
+   * @see https://xrpl.org/depositauth.html
```
keefertaylor, line 433: nit: 1 line break after summary

src/XRP/default-xrp-client.ts
```diff
@@ -426,6 +428,27 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+   *
+   * @param wallet The wallet associated with the XRPL account enabling Deposit Authorization and that will sign the request.
+   * @returns A promise which resolves to a string representing the hash of the submitted AccountSet transaction.
+   */
+  public async enableDepositAuth(wallet: Wallet): Promise<string> {
+    // 9 is the flag for enabling Deposit Auth: https://xrpl.org/accountset.html#accountset-flags
+    const setFlag = new SetFlag()
```
keefertaylor, line 440: Note: We have [`RippledFlags`](https://github.com/xpring-eng/Xpring-JS/blob/master/src/XRP/rippled-flags.ts) which are a set of transaction flags. 

We could maybe create a wrapper for this flag as well? Maybe `RippledFlags` should be renamed `TransactionFlags` to disambiguate? 

Open to this being a TODO with a follow on.

src/XRP/default-xrp-client.ts
keefertaylor: I basically follow this guide: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/

test/XRP/default-xrp-client.test.ts
```diff
@@ -786,4 +786,48 @@ describe('Default XRP Client', function (): void {
...
+  > {
+    // GIVEN a DefaultXrpClient with mocked networking that will return a successful hash for submitTransaction
+    const xrpClient = new DefaultXrpClient(
+      fakeSucceedingNetworkClient,
+      XrplNetwork.Test,
+    )
+
+    const { wallet } = Wallet.generateRandomWallet()!
+
+    // WHEN enableDepositauth is called
```
keefertaylor, line 801:
```suggestion
    // WHEN enableDepositAuth is called
```

test/XRP/fakes/fake-xrp-client.ts
```diff
@@ -67,6 +67,14 @@ class FakeXrpClient implements XrpClientDecorator {
     return FakeXrpClient.returnOrThrow(this.getPaymentValue)
   }
 
+  async enableDepositAuth(_wallet: Wallet): Promise<string> {
+    // NOTE: all we need to do here is return a valid transaction hash, which we already have in the field this.sendValue
```
keefertaylor, line 71: I'd go with the latter - a generic transaction hash. 

If we ever need something more advanced we can always easily refactor this API with breaking changes since it is only test code.

src/XRP/default-xrp-client.ts
amiecorso: Creating a separate, custom `AccountSetFlag` enum

keefertaylor:
\+ stephen to make sure this is actually what we want to do.

amiecorso:
> I know that we don't have an easy way to test this programmatically but can we verify that the integration test actually applied the change on that account correctly?

The integration test retrieves the account info after enabling of deposit auth and check's the account's flag bitmap to ensure the deposit auth flag has been correctly set.

src/XRP/model/account-set-flag.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Flags used in AccountSet transactions.
+ *
+ * @see https://xrpl.org/accountset.html#accountset-flags
+ */
+export enum AccountSetFlag {
```
sarahkevinking, line 6: Small note, but I really appreciate how well commented these enums are in Xpring JS and that we are referencing the Xrpl docs. I wish the table in the docs displayed these flags in order too but thats outside of our scope.

src/XRP/model/payment-flags.ts
```diff
@@ -1,11 +1,13 @@
...
+ * Flags used in payment transactions.
+ *
+ * @note These are only flags which are utilized in Xpring SDK.
+ * For the full list of payment flags, @see https://xrpl.org/payment.html#payment-flags
  *
- * @note These are only flags which are utilized in Xpring SDK. For a complete list of flags, see: https://xrpl.org/transaction-common-fields.html#flags-field.
  */
-class RippledFlags {
+class PaymentFlags {
   static TF_PARTIAL_PAYMENT = 1 << 17
```
sarahkevinking, line 11: Any reason we have to encode these constants in bit shift operations rathering than the hex code used in the documentation?

src/XRP/xrp-client.ts
```diff
@@ -139,4 +140,16 @@ export default class XrpClient implements XrpClientInterface {
   ): Promise<XrpTransaction | undefined> {
     return this.decoratedClient.getPayment(transactionHash)
   }
+
+  /**
```
sarahkevinking, line 144: A pattern I really admired from Go is the ability to define an interface once and require it by other types. I notice that xrp-client.ts, xrp-client-interface, and xrp-client-decorator share the same API. Are there ways to only comment a single time?

src/XRP/model/transaction-result.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Represents the outcome of submitting an XRPL transaction.
+ */
+export default class TransactionResult {
+  hash: string
+  status: TransactionStatus
+  validated: boolean
+
+  /**
+   *
```
keefertaylor, line 12: nit: extra line here

src/XRP/model/transaction-result.ts
```diff
@@ -0,0 +1,27 @@
+import TransactionStatus from '../transaction-status'
+
+/**
+ * Represents the outcome of submitting an XRPL transaction.
+ */
+export default class TransactionResult {
+  hash: string
```
keefertaylor, line 7: please mark as `public readonly`. 

Immutability constraints are generally helpful when we can apply them because it makes client code easier to reason about.

src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -48,12 +49,72 @@ export default class ReliableSubmissionXrpClient implements XrpClientDecorator {
...
+  public async enableDepositAuth(wallet: Wallet): Promise<TransactionResult> {
+    const result = await this.decoratedClient.enableDepositAuth(wallet)
+    const transactionHash = result.hash
+
+    const rawTransactionStatus = await this.awaitFinalTransactionResult(
+      transactionHash,
+      wallet,
+    )
+
+    // Return pending if the transaction is not validated.
```
keefertaylor, line 96: Sorry if I missed this in our other libs but should we refactor this logic to a helper (`determineFinalResult(rawTransactionStatus)`)

src/XRP/xrp-client.ts
keefertaylor: Not that I know of, *BUT* I am not a typescript expert. 

We've opted to over document rather than under document. At some point we'd like to generate docs from these comments and I assume that that point will be when we'll investigate the "right" way to do this. 

Open to feedback if you know the correct answer.

PR #591, opened 2020-07-13
(3/3) Reliable submission logic for `enableDepositAuth`

amiecorso:
The basic functionality for `enableDepositAuth` is implemented [here](https://github.com/xpring-eng/Xpring-JS/pull/585).  This PR implements `enableDepositAuth` in the `ReliableSubmissionXrpClient` (the decorated client actually used by publicly exposed`XrpClient`), and also refactors the reliable submission logic into a private method for re-use.

The return value of `enableDepositAuth` is changed from `Promise<string>` (a transaction hash) to `Promise<TransactionResult>`, where `TransactionResult` is a new class containing a `hash` and `status` field.  The implementation in `DefaultXrpClient` may return `TransactionStatus`es that are `.pending`, `.succeeded`, `.failed`, whereas the implementation in `ReliableSubmissionXrpClient` may return `TransactionStatus`es that are `.succeeded` or `.failed` but not `.pending`, as it awaits finality of transaction result (inclusion in a validated ledger OR passing of `lastLedgerSequence`) before returning. 

TODO: 
- web client integration tests in keeping with the other patterns?
- note new functionality in CHANGELOG
- demo new functionality in README??

### Context of Change
We would like users of the SDK to be able to call `enableDepositAuth` and immediately understand whether the operation was successful or not, without having to follow up on their own to understand whether the transaction has been included in a validated ledger yet, and if so whether it actually modified the Account or not. 

### Type of Change

## Before / After
- `enableDepositAuth` now returns a `Promise<[string, TransactionStatus]>` (which would be breaking but `enableDepositAuth` has not been released yet)
- there is a reliable submission implementation for `enableDepositAuth` 

## Test Plan
CI still passes.  New tests written for `ReliableSubmissionXrpClient`


src/XRP/reliable-submission-xrp-client.ts
```diff
@@ -48,12 +48,69 @@ export default class ReliableSubmissionXrpClient implements XrpClientDecorator {
...
+  ): Promise<[string, TransactionStatus]> {
+    const result = await this.decoratedClient.enableDepositAuth(wallet)
+    const transactionHash = result[0]
+
+    const rawTransactionStatus = await this.awaitFinalTransactionResult(
+      transactionHash,
+      wallet,
+    )
+
+    let transactionStatus: TransactionStatus
```
keefertaylor, line 97: nit: mildly cleaner to write this as:
```js
if (!rawTransactionStatus.isValidated) {
  return [hash, TransactionStatus.pending]
} else {
  const transactionStatus = ...
  return [hash, transactionStatus]
}
```

That way you don't have a dangling `let`.

src/XRP/default-xrp-client.ts
```diff
@@ -435,9 +435,12 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
    *
    * @param wallet The wallet associated with the XRPL account enabling Deposit Authorization and that will sign the request.
-   * @returns A promise which resolves to a string representing the hash of the submitted AccountSet transaction.
+   * @returns A promise which resolves to a Tuple<string, TransactionStatus> representing the hash of the submitted AccountSet transaction
+   *          and the status of the transaction.
    */
-  public async enableDepositAuth(wallet: Wallet): Promise<string> {
+  public async enableDepositAuth(
+    wallet: Wallet,
+  ): Promise<[string, TransactionStatus]> {
```
keefertaylor, line 443: I'd prefer this API to be an object (`Transaction Result`?) rather than an array. 

As a non-JS native programmer, arrays of heterogenous objects don't make a ton of sense to me 
 and objects can have names that confer meaning rather than just indices. It's easier to say something like:

```js
const result = enableDepositAuth(...) 
console.log(result.status)
```

than
```js
console.log(result[1])
```

src/XRP/rippled-flags.ts
```diff
@@ -8,6 +8,8 @@
 class RippledFlags {
   static TF_PARTIAL_PAYMENT = 1 << 17
 
+  // https://xrpl.org/accountroot.html#accountroot-flags
```
keefertaylor, line 11: I think that because this is a separate flag sets it should live in a different file

src/XRP/default-xrp-client.ts
amiecorso: cool, definitely agree with this and was considering creating an object

PR #594, opened 2020-07-15
(1/3) Updates to use newest version of `xpring-common-js`

amiecorso:
The newest version of `xpring-common-js` includes breaking changes around the namespacing of `Utils` and `XrpUtils` functionalities, which have been officially split into the different namespaces.  This PR updates `xpring-js` to use the correct namespace everywhere necessary so that it will be ready to incorporate the new release.

NOTE: There is also an `XrpUtils` that is native to `xpring-js` and this is creating a name collision in `index.js`.  My temporary solve was to export the native `XrpUtils` as `XrpXrpUtils` (since it's in the XRP directory), but I need to figure out what the correct way to handle this is, since both of the `XrpUtils` need to be exported at the top level for consumers of the library.

Also TODO: actually bump the version.  This currently passes CI without the new version because the current version has all functionality in both namespaces (the newest version removes the deprecated methods).  Based these changes on a local `npm link` of current master branch of x-c-js.

### Context of Change
New `xpring-common-js` release w/breaking changes.

### Type of Change

## Test Plan
CI still passes.


keefertaylor:
This is super ugly. 

One way may be to skip this update of x-c-js for the moment and push this code down into xpring-common-js and try again next time? I realize that we specifically want some serialization code though. 

Otherwise.. we could just [merge the two utils](https://www.javascripttutorial.net/object/javascript-merge-objects/) and clean up asynchronously? We would need to test that this actually works.

keefertaylor:
We should merge this after #597 gets merged.

amiecorso:
Great, thanks for taking care of the merge situation.  I've updated all the imports to reference `xrp-utils.ts` and gotten rid of the horrible `XrpXrpUtils`

PR #597, opened 2020-07-17
Merge Xpring-JS XrpUtils with Xpring-Common-JS XrpUtils

keefertaylor:

Export the union of `XrpUtils` functionality in Xpring-JS and Xpring-common-JS.

### Context of Change

Xpring-JS provides some functionality in `XrpUtils` which is language specific. Xpring-common-JS provides some generic functionality. Merge and export these two sets.

I considered pushing this functionality down into xpring-common-JS as well. Since these functions throw, we would also need to push down `XrpError`. This is heavyweight and isn't used by any client besides Xpring-JS (XpringKit/Xpring4j both implement XrpUtils functionality natively). 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

`XrpUtils` is the one stop shop for XRP related functionality (rather than `XrpUtils` and `Utils`)

## Test Plan

CI


keefertaylor:
> This seems great. Does that mean we should refactor the imports throughout the codebase that explicitly import { XrpUtils } from 'xpring-common-js' ? Or does it not matter internally?

Since our exported `XrpUtils` is a super set of the functionality I think we can / should just use the local version.

PR #600, opened 2020-07-17
Update README.md

LoisRP:
typo fix

### Type of Change

- [x] Documentation Updates
- [ ] Release


LoisRP:
I see "A wallet is a fundamental model object in XpringKit", which seems like a copy and paste error. Is that what we want to say here?

PR #607, opened 2020-07-21
Revert "Optimize GitLab CI"

keefertaylor:
Reverts xpring-eng/Xpring-JS#606


keefertaylor:
Unfortunately,  codecov support doesn't work so rolling this back to try again.

PR #648, opened 2020-08-11
Implements `randomWalletFromFaucet` for Improved `XrpClient` Integration Tests

amiecorso:
This PR implements the test utility `randomWalletFromFaucet`, which generates a random wallet and then requests funds via the XRPL Testnet faucet (https://xrpl.org/xrp-testnet-faucet.html).  Integration tests are refactored to use a new funded address each time, which should eliminate the failures that occur due to sequence number clashes when different clients (via different CI runs) send transactions on behalf of the same account.

### Type of Change

## Before / After
New addresses are used for integration tests where relevant.

## Test Plan
Refactor integration tests where necessary.  CI still passes.


test/XRP/helpers/xrp-test-utils.ts
```diff
@@ -29,6 +37,70 @@ export default class XRPTestUtils {
     }
     return paymentXrpTransactions
   }
+
+  /**
+   * Generates a random wallet and funds using the XRPL Testnet faucet.
+   */
+  static async randomWalletFromFaucet(): Promise<Wallet | undefined> {
```
keefertaylor, line 44: since we can't do anything with `undefined` can this function just throw if something goes awry?

amiecorso: throws

test/XRP/helpers/xrp-test-utils.ts
```diff
@@ -1,12 +1,20 @@
...
+import XrpClient from '../../../src/XRP/xrp-client'
+import axios from 'axios'
 
 /**
  * Convenience class for utility functions used in test cases for XrpClient infrastructure.
  */
 export default class XRPTestUtils {
+  static sleep = async (milliseconds: number): Promise<void> => {
+    return new Promise((resolve) => setTimeout(resolve, milliseconds))
+  }
```
tedkalaw, line 16: nit: you don't need an expression+return here; ie

`...Promise<void> => new Promise(....`

test/XRP/helpers/xrp-test-utils.ts
tedkalaw: since it throws and we can't do anything with `undefined`, can we remove it from the return type?

test/XRP/helpers/xrp-test-utils.ts
```diff
@@ -29,6 +37,76 @@ export default class XRPTestUtils {
...
+      } catch {
+        currentBalance = bigInt('0')
+      }
+      // If our current balance has changed then return
+      if (startingBalance.notEquals(currentBalance)) {
+        return wallet
+      }
+
+      // In the future if we had a tx hash from the faucet
+      // We should check the status of the tx which would be more accurate
```
tedkalaw, line 102: is this worth tracking in a task?

amiecorso: probably not - for our purposes here, we're generating a branch new wallet before requesting funds.  I think the only reason this would matter is if there were another client potentially manipulating the same address (or simultaneously sending funds to the same address) while we're tapping the faucet.  Just copied this comment from the wallet implementation of this same logic (where things might matter more)

PR #715, opened 2020-09-21
Adds New `XrpError` Type Specific to Protocol Buffer Conversions

amiecorso:
This PR adds a new type to the `XrpError` class in preparation for throwing descriptive errors from the Protobuf wrapper classes.  Note that while a pre-packaged `XrpError.malformedProtobuf` object exists as a static attribute with default error message on `XrpError`, we encourage the use of custom instances of `XrpError` with type `XrpTransactionType.MalformedProtobuf` and a more descriptive error message. 

### Context of Change
We are about to audit our Protobuf wrapper classes in `xpring-js` for more robust logic checks, more useful throwing behavior, and more descriptive errors.

### Type of Change

## Before / After
`XrpErrorType` has a new enum member, `XrpError` has a new static attribute.

## Test Plan
CI passes

### DIFF
--- a/src/XRP/xrp-error.ts
+++ b/src/XRP/xrp-error.ts
@@ -33,10 +33,11 @@ export default class XrpError extends Error {
 
   /**
    * Encountered a protocol buffer formatted in contradiction to the logic of the XRPL.
+   * @see xrpl.org for XRPL documentation.
    */
   static malformedProtobuf = new XrpError(
     XrpErrorType.MalformedProtobuf,
-    'Encountered a protocol buffer formatted in contradiction to the XRPL business logic',
+    'Encountered a protocol buffer in unexpected format.  See xrpl.org for XRPL documentation.',
   )
 
   /**
--- a/src/XRP/xrp-error.ts
+++ b/src/XRP/xrp-error.ts
@@ -4,6 +4,7 @@
 export enum XrpErrorType {
   InvalidInput,
   PaymentConversionFailure,
+  MalformedProtobuf,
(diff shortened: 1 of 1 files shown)


mvadari (approved):

src/XRP/xrp-error.ts
```diff
@@ -30,6 +31,14 @@ export default class XrpError extends Error {
     'Could not convert payment transaction: (transaction). Please file a bug at https://github.com/xpring-eng/Xpring-JS/issues',
   )
 
+  /**
+   * Encountered a protocol buffer formatted in contradiction to the logic of the XRPL.
```
tedkalaw, line 35: can we link to the docs?

src/XRP/xrp-error.ts
```diff
@@ -30,6 +31,14 @@ export default class XrpError extends Error {
...
+   */
+  static malformedProtobuf = new XrpError(
+    XrpErrorType.MalformedProtobuf,
+    'Encountered a protocol buffer formatted in contradiction to the XRPL business logic',
```
tedkalaw, line 39: perhaps "Encountered a protocol buffer with unexpected format" or something similar? i feel this error message could be simpler

tedkalaw (approved):
seems reasonable, one comment inline

src/XRP/xrp-error.ts
amiecorso: good idea.  these'll have to be the generic xrpl.org docs

mvadari (approved):

PR #716, opened 2020-09-21
Audits `XrpCurrency` Protobuf Wrapper

amiecorso:
This PR adds additional logic validations and descriptive error throwing to the class `XrpCurrency`.

Notes (mostly for Mayukha):
- Note that I changed the type of the `code` field to a `string` from an `UInt8Array` ...  The rationale here is that these wrapper classes are ultimately human-facing... they provide an abstraction on top fo the networking layer, and we'll probably want to eventually help them print pretty, etc.  Strings are a more human-readable format than byte arrays.  As you encounter byte arrays, let's discuss what type we think they should be.
- I probably also could have added checks for whether the currency name is exactly 3 characters and the code is exactly 20 bytes... But for now, I'm just going to worry about tackling the big stuff.  You can use your own discretion or ask us if you're questioning what to examine and what to leave.
- Your test efforts are basically going to involve creating a fake protocol buffer object with the desired properties (such as, for example, containing a code but missing a name), and then either writing fresh tests (as I did here) or altering existing tests (as you will frequently do for the higher-level types).  
- Tests which previously checked for an `undefined` return value in the case of a malformed protobuf will probably have to be updated to check for the properly thrown error once the class is modified.

### Context of Change
One component of our attempt to make protobuf wrapper classes as safe and informative as possible.  An SDK safety net. 

### Type of Change

## Before / After
`XrpCurrency` performs logic/error checking.

## Test Plan
New tests added to verify errors are correctly thrown.  CI passes.

## Future Tasks

Link to Asana task with all related subtasks.
https://app.asana.com/0/1188384817180008/1193021408531788/f

### DIFF
--- a/src/XRP/model/xrp-currency.ts
+++ b/src/XRP/model/xrp-currency.ts
@@ -1,3 +1,4 @@
+import { XrpError, XrpErrorType } from '..'
 import { Currency } from '../Generated/web/org/xrpl/rpc/v1/amount_pb'
 
 /**
@@ -14,12 +15,26 @@ export default class XrpCurrency {
    * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41
    */
   public static from(currency: Currency): XrpCurrency {
-    return new XrpCurrency(currency.getName(), currency.getCode_asU8())
+    const currencyName = currency.getName()
+    const currencyCode = currency.getCode_asB64()
+    if (!currencyName) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Currency protobuf missing required field `name`.',
+      )
+    }
+    if (!currencyCode) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Currency protobuf missing required field `code`.',
+      )
+    }
+    return new XrpCurrency(currencyName, currencyCode)
   }
 
   /**
    * @param name 3 character ASCII code
    * @param code 160 bit currency code. 20 bytes
    */
-  private constructor(readonly name: string, readonly code: Uint8Array) {}
+  private constructor(readonly name: string, readonly code: string) {}
 }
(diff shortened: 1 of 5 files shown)


mvadari (approved):

src/XRP/model/xrp-currency.ts
```diff
@@ -14,12 +15,26 @@ export default class XrpCurrency {
    * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41
    */
   public static from(currency: Currency): XrpCurrency {
-    return new XrpCurrency(currency.getName(), currency.getCode_asU8())
+    const currencyName = currency.getName()
+    const currencyCode = currency.getCode_asB64()
+    if (!currencyName) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Currency protobuf missing required field `name`.',
```
tedkalaw, line 23: if we use this many more times, i think it's probably worth adding a simple formatting function to reuse instead of typing this this out everytime, ie

`missingFieldMessage(fieldName: string)`

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -74,6 +76,30 @@ describe('Protocol Buffer Conversion', function (): void {
     assert.deepEqual(currency.name, testCurrencyProto.getName())
   })
 
+  it('Convert Currency protobuf missing required field name', function (): void {
+    // GIVEN a Currency protocol buffer missing a name.
+    // WHEN the protocol buffer is converted to a native Typescript type THEN an error is thrown.
```
tedkalaw, line 81: is it a common pattern to have THEN on the same line as WHEN if it fits? on first glance i thought it was missing

tedkalaw (approved):
> I probably also could have added checks for whether the currency name is exactly 3 characters and the code is exactly 20 bytes...
worth tracking in a task?

seems reasonable to me, couple thoughts inline

test/XRP/xrp-protocol-buffer-conversion.test.ts
mvadari: I've seen it in a lot of other tests - I don't think I've seen them on separate lines when they're right next to each other

tedkalaw: dope, thanks!

PR #719, opened 2020-09-22
Audits `XrpIssuedCurrency` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpIssuedCurrency`.

### Context of Change

### Type of Change

## Before / After

`XrpIssuedCurrency` performs logic/error checking, based on the business logic in the docs.

## Test Plan

New tests added to confirm that errors are correctly thrown. Old tests edited so that they correctly respond to the new errors. CI passes.

## Future Tasks

### DIFF
--- a/src/XRP/model/xrp-issued-currency.ts
+++ b/src/XRP/model/xrp-issued-currency.ts
@@ -24,7 +24,7 @@ export default class XrpIssuedCurrency {
     if (!currency) {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
-        'Issued currency protobuf does not contain valid `currency` field.',
+        'IssuedCurrency protobuf does not contain `currency` field.',
       )
     }
     const xrpCurrency = XrpCurrency.from(currency)
@@ -35,7 +35,7 @@ export default class XrpIssuedCurrency {
     } catch {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
-        'Issued currency protobuf does not contain valid `value` field.',
+        'Cannot construct BigInt from IssuedCurrency protobuf `value` field.',
       )
     }
 
@@ -43,7 +43,7 @@ export default class XrpIssuedCurrency {
     if (!issuer) {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
-        'Issued currency protobuf does not contain valid `issuer` field.',
+        'IssuedCurrency protobuf does not contain `issuer` field.',
       )
     }
     return new XrpIssuedCurrency(xrpCurrency, value, issuer)
--- a/src/XRP/model/xrp-issued-currency.ts
+++ b/src/XRP/model/xrp-issued-currency.ts
@@ -1,3 +1,4 @@
+import { XrpError, XrpErrorType } from '..'
 import bigInt, { BigInteger } from 'big-integer'
 import { IssuedCurrencyAmount } from '../Generated/web/org/xrpl/rpc/v1/amount_pb'
 import XrpCurrency from './xrp-currency'
@@ -5,6 +6,7 @@ import XrpCurrency from './xrp-currency'
 /*
  * An issued currency on the XRP Ledger
  * @see: https://xrpl.org/basic-data-types.html#specifying-currency-amounts
+ * @see: https://xrpl.org/currency-formats.html#issued-currency-amounts
  */
 export default class XrpIssuedCurrency {
   /**
@@ -19,20 +21,32 @@ export default class XrpIssuedCurrency {
     issuedCurrency: IssuedCurrencyAmount,
   ): XrpIssuedCurrency | undefined {
     const currency = issuedCurrency.getCurrency()
-    const xrpCurrency = currency && XrpCurrency.from(currency)
+    if (!currency) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `currency` field.',
+      )
+    }
+    const xrpCurrency = XrpCurrency.from(currency)
 
     let value
     try {
       value = bigInt(issuedCurrency.getValue())
     } catch {
-      value = undefined
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `value` field.',
+      )
     }
 
-    if (value) {
-      const issuer = issuedCurrency.getIssuer()?.getAddress()
-      return new XrpIssuedCurrency(xrpCurrency, value, issuer)
+    const issuer = issuedCurrency.getIssuer()?.getAddress()
+    if (!issuer) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `issuer` field.',
+      )
     }
-    return undefined
+    return new XrpIssuedCurrency(xrpCurrency, value, issuer)
   }
 
   /**
--- a/src/XRP/model/xrp-issued-currency.ts
+++ b/src/XRP/model/xrp-issued-currency.ts
@@ -17,9 +17,7 @@ export default class XrpIssuedCurrency {
    * @returns an XrpIssuedCurrency with its fields set via the analogous protobuf fields.
    * @see https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L28
    */
-  public static from(
-    issuedCurrency: IssuedCurrencyAmount,
-  ): XrpIssuedCurrency | undefined {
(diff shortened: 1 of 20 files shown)


src/XRP/model/xrp-issued-currency.ts
```diff
@@ -19,20 +21,32 @@ export default class XrpIssuedCurrency {
     issuedCurrency: IssuedCurrencyAmount,
   ): XrpIssuedCurrency | undefined {
     const currency = issuedCurrency.getCurrency()
-    const xrpCurrency = currency && XrpCurrency.from(currency)
+    if (!currency) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `currency` field.',
```
amiecorso, line 27:
```suggestion
        'IssuedCurrency protobuf does not contain `currency` field.',
```

amiecorso: suggesting we always use the official name of the type when referring to the protobuf

src/XRP/model/xrp-issued-currency.ts
```diff
@@ -19,20 +21,32 @@ export default class XrpIssuedCurrency {
...
+      )
+    }
+    const xrpCurrency = XrpCurrency.from(currency)
```
amiecorso, line 30: The nice thing about what we're doing is that once all of these classes are throwing descriptive errors, we don't need to worry about surrounding this for checks as to whether the `XrpCurrency` object was successfully constructed from the `Currency` proto, because we know that `XrpCurrency.from` will already throw if that's the case.

src/XRP/model/xrp-issued-currency.ts
```diff
@@ -19,20 +21,32 @@ export default class XrpIssuedCurrency {
...
 
     let value
     try {
       value = bigInt(issuedCurrency.getValue())
     } catch {
-      value = undefined
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `value` field.',
```
amiecorso, line 38:
```suggestion
        'Cannot construct BigInt from IssuedCurrency protobuf `value` field.',
```

src/XRP/model/xrp-issued-currency.ts
```diff
@@ -19,20 +21,32 @@ export default class XrpIssuedCurrency {
...
     }
 
-    if (value) {
-      const issuer = issuedCurrency.getIssuer()?.getAddress()
-      return new XrpIssuedCurrency(xrpCurrency, value, issuer)
+    const issuer = issuedCurrency.getIssuer()?.getAddress()
+    if (!issuer) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Issued currency protobuf does not contain valid `issuer` field.',
```
amiecorso, line 46:
```suggestion
        'IssuedCurrency protobuf does not contain `issuer` field.',
```

amiecorso: in this case, that if statement is only going to be triggered if `issuer` is undefined, so I feel like it makes more sense to be specific in our error message that the field isn't merely invalid, but it doesn't even exist.

test/XRP/fakes/fake-xrp-protobufs.ts
```diff
@@ -346,16 +346,24 @@ testSignerEntryProto.setSignerWeight(testSignerWeightProto)
...
-testInvalidIssuedCurrencyProto.setIssuer(accountAddressProto)
-testInvalidIssuedCurrencyProto.setValue(testInvalidIssuedCurrencyValue)
+// Invalid IssuedCurrencyAmount protos
+const testInvalidValueIssuedCurrencyProto = new IssuedCurrencyAmount()
+testInvalidValueIssuedCurrencyProto.setCurrency(testCurrencyProto)
+testInvalidValueIssuedCurrencyProto.setIssuer(accountAddressProto)
+testInvalidValueIssuedCurrencyProto.setValue(testInvalidIssuedCurrencyValue)
+
+const testInvalidCurrencyIssuedCurrencyProto = new IssuedCurrencyAmount()
+testInvalidValueIssuedCurrencyProto.setIssuer(accountAddressProto)
```
amiecorso, line 356:
```suggestion
testInvalidCurrencyIssuedCurrencyProto.setIssuer(accountAddressProto)
```

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -160,13 +162,38 @@ describe('Protocol Buffer Conversion', function (): void {
...
+      },
+      XrpError,
+      'Issued currency protobuf does not contain valid `value` field.',
     )
+  })
 
-    // THEN the result is undefined
-    assert.isUndefined(issuedCurrency)
+  it('Convert IssuedCurrency with bad issuer', function (): void {
+    // GIVEN an issued currency protocol buffer with a non numeric value
```
amiecorso, line 176:
```suggestion
    // GIVEN an issued currency protocol buffer with a missing issuer field
```

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -160,13 +162,38 @@ describe('Protocol Buffer Conversion', function (): void {
...
+      () => {
+        XrpIssuedCurrency.from(testInvalidIssuerIssuedCurrencyProto)
+      },
+      XrpError,
+      'Issued currency protobuf does not contain valid `issuer` field.',
+    )
+  })
+
+  it('Convert IssuedCurrency with bad currency', function (): void {
+    // GIVEN an issued currency protocol buffer with a non numeric value
```
amiecorso, line 188:
```suggestion
    // GIVEN an issued currency protocol buffer with a missing currency field
```

amiecorso (changes requested):
This is also looking good - again, just a few aesthetic suggestions and one or two catches in constructing your fake protos (most of my bugs in constructing fake protos involve setting fields on the wrong object because I typed or auto-completed the wrong name :) 

I'm also seeing that we actually probably should create a chain of dependent branches... because the behavior you see at a higher-level object (such as in this `XrpIssuedCurrency` which contains an `XrpCurrency`) is going to be affected by the behavior of lower-level objects being used to construct them (as you're seeing in having to update the behavior of tests for higher level objects.)  Luckily we've got things in topological order, so hopefully chaining your branches will prevent much duplicated work.

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -159,14 +161,39 @@ describe('Protocol Buffer Conversion', function (): void {
...
-    const issuedCurrency = XrpIssuedCurrency.from(
-      testInvalidIssuedCurrencyProto,
+    // GIVEN an issued currency protocol buffer with an invalid value field
+    // WHEN the protocol buffer is converted to a native TypeScript type THEN an error is thrown.
+    assert.throws(
+      () => {
+        XrpIssuedCurrency.from(testInvalidIssuedCurrencyProtoBadValue)
+      },
+      XrpError,
+      'Cannot construct BigInt from IssuedCurrency protobuf `value` field.',
```
tedkalaw, line 171: this feels pretty brittle to me, as it means that changes to the log text will result in a test failure. i think there's a couple solves here

1) honestly, i think it's probably ok to not check the error message
2) from what i can tell, the main reason we need to do this is because XrpError doesn't encode enough information. you could do something where you implement a new error type that has enough information...honestly this feels like overkill right now

i would go with 1

tedkalaw (changes requested):
seems reasonable to me; i think we should remove checks in tests for the messages

test/XRP/xrp-protocol-buffer-conversion.test.ts
mvadari: Alright, I'll take them out.

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -478,14 +506,17 @@ describe('Protocol Buffer Conversion', function (): void {
...
+    // WHEN the protocol buffer is converted to a native TypeScript type THEN an error is thrown.
+    assert.throws(
+      () => {
+        XrpTransaction.from(
+          testInvalidGetTransactionResponseProto,
+          XrplNetwork.Test,
+        )
+      },
+      XrpError,
+      'Cannot construct BigInt from IssuedCurrency protobuf `value` field.',
```
tedkalaw, line 518: can we remove the rest of the error message checks as well?

mvadari: Oops missed this one

amiecorso (approved):
agree with Ted's comments, looks like you've addressed them all, I think this looks great.

tedkalaw (approved):
nice!

PR #720, opened 2020-09-22
Adds DefaultXrpClient::getTransactionResult

tedkalaw:
Adds `DefaultXrpClient::getTransactionResult`, which populates a `TransactionResult` with the data it needs.

### Context of Change
I'm going to use this for `DepositPreauth`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
DefaultXrpClient uses `getTransactionResult` to calculate a `TransactionResult` in `enableDepositPreauth` instead of constructing a `TransactionResult` by hand.

## Test Plan
CI, tests

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -132,18 +132,7 @@ export default class DefaultXrpClient implements XrpClientDecorator {
   public async getPaymentStatus(
     transactionHash: string,
   ): Promise<TransactionStatus> {
-    const transactionStatus = await this.getRawTransactionStatus(
-      transactionHash,
-    )
-
-    // Return pending if the transaction is not validated.
-    if (!transactionStatus.isValidated) {
-      return TransactionStatus.Pending
-    }
-
-    return transactionStatus.transactionStatusCode?.startsWith('tes')
-      ? TransactionStatus.Succeeded
-      : TransactionStatus.Failed
+    return await this.getTransactionStatus(transactionHash)
   }
 
   /**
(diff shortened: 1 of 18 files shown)


src/XRP/default-xrp-client.ts
```diff
@@ -533,4 +526,24 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
     return Utils.toHex(response.getHash_asU8())
   }
+
+  /**
+   * Returns a detailed TransactionResult for a given XRPL transaction hash.
+   *
+   * @param transactionHash The transaction hash to populate a TransactionResult for.
+   * @returns A Promise which resolves to a TransactionResult associated with the given transaction hash.
+   */
+  private async getTransactionResult(
```
tedkalaw, line 536: didn't add tests for this since it's an impl detail of DefaultXrpClient and will be put through its paces by the diff transaction types. lmk though if you think i should test

mvadari (approved):

src/XRP/default-xrp-client.ts
```diff
@@ -533,4 +526,24 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+    transactionHash: string,
+  ): Promise<TransactionResult> {
+    const rawStatus = await this.getRawTransactionStatus(transactionHash)
+    const isValidated = rawStatus.isValidated
+    const transactionStatus = await this.getPaymentStatus(transactionHash)
```
amiecorso, line 541: So really my only thought is:  this `getPaymentStatus` name is misleading, because the logic here is intended to recover the transaction status for any type of transaction, not just Payments (a relic of when the SDK only ever did and was only ever supposed to do XRP Payments).  The body of `getPaymentStatus` is: 

```
  /**
   * Retrieve the transaction status for a Payment given transaction hash.
   *
   * Note: This method will only work for Payment type transactions which do not have the tf_partial_payment attribute set.
   * @see https://xrpl.org/payment.html#payment-flags
   *
   * @param transactionHash The hash of the transaction.
   * @returns The status of the given transaction.
   */
  public async getPaymentStatus(
    transactionHash: string,
  ): Promise<TransactionStatus> {
    const transactionStatus = await this.getRawTransactionStatus(
      transactionHash,
    )

    // Return pending if the transaction is not validated.
    if (!transactionStatus.isValidated) {
      return TransactionStatus.Pending
    }

    return transactionStatus.transactionStatusCode?.startsWith('tes')
      ? TransactionStatus.Succeeded
      : TransactionStatus.Failed
  }
```
and I'm thinking we should just pull the relevant bits of logic (i.e. most of it) out of this (especially because we're ALREADY making a call to `getRawTransactionStatus` and put it right here in `getTransactionResult`.  Maybe there's a way to refactor wherever is using `getPaymentStatus` and have it use `getTransactionResult`, and then just get rid of `getPaymentStatus`... I haven't looked into it closely enough.

amiecorso (approved):
See my notes on `getPaymentStatus` - otherwise looks good to me and I agree that we can skip the test for now.

src/XRP/default-xrp-client.ts
tedkalaw: per our discussion, i moved the logic into `getTransactionStatus` and then am using it in both places - lmk what you think

i also added https://app.asana.com/0/0/1195443262464650/f to track this

amiecorso (approved):

PR #721, opened 2020-09-22
Adds DefaultXrpClient::authorizeSendingAccount

tedkalaw:
This PR adds a basic, non-reliable implementation of authorizing DepositPreauth.

Note that the name is weird - would love ideas for a better name.

### Context of Change
DepositPreauth is a useful primitive, particularly for issued currencies.

See docs here:
https://xrpl.org/depositpreauth.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
CI, tests

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -477,15 +477,7 @@ export default class DefaultXrpClient implements XrpClientDecorator {
       wallet,
     )
 
-    const rawStatus = await this.getRawTransactionStatus(transactionHash)
-    const isValidated = rawStatus.isValidated
-    const transactionStatus = await this.getPaymentStatus(transactionHash)
-
-    return new TransactionResult(
-      transactionHash,
-      transactionStatus,
-      isValidated,
-    )
+    return await this.getTransactionResult(transactionHash)
   }
 
   /**
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -448,7 +448,7 @@ export default class DefaultXrpClient implements XrpClientDecorator {
   }
 
   /**
-   * Enables DepositPreauth and authorizes an XRPL account to deposit to this XRPL account.
+   * Enables DepositPreauth and authorizes an XRPL account to send to this XRPL account.
    *
    * @set https://xrpl.org/depositpreauth.html
    *
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -132,18 +132,7 @@ export default class DefaultXrpClient implements XrpClientDecorator {
   public async getPaymentStatus(
     transactionHash: string,
   ): Promise<TransactionStatus> {
-    const transactionStatus = await this.getRawTransactionStatus(
-      transactionHash,
-    )
-
-    // Return pending if the transaction is not validated.
-    if (!transactionStatus.isValidated) {
-      return TransactionStatus.Pending
-    }
-
-    return transactionStatus.transactionStatusCode?.startsWith('tes')
-      ? TransactionStatus.Succeeded
-      : TransactionStatus.Failed
+    return await this.getTransactionStatus(transactionHash)
   }
 
   /**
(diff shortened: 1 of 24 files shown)


src/XRP/default-xrp-client.ts
```diff
@@ -456,6 +458,43 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
 
+  /**
+   * Enables DepositPreauth and authorizes an XRPL account to deposit to this XRPL account.
+   *
+   * @set https://xrpl.org/depositpreauth.html
+   *
+   * @param xAddressToAuthorize The X-Address of the account to enable DepositPreauth for.
+   * @param wallet The wallet associated with the XRPL account enabling DepositPreauth and that will sign the request.
+   */
+  public async authorizeDepositPreauth(
```
tedkalaw, line 469: this name is weird, but since the docs specify that you can only either authorize or unauthorize, it felt useful to have separate apis for the two instead of a more abstract function that does both so you can't hurt yourself in that way. let me know what you think. feels like it could also be `depositPreauthAuthorize`

keefertaylor: > it felt useful to have separate apis for the two instead of a more abstract function that does both so you can't hurt yourself in that way.

Agree.

keefertaylor (approved):

src/XRP/default-xrp-client.ts
```diff
@@ -456,6 +458,43 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
+  /**
+   * Enables DepositPreauth and authorizes an XRPL account to deposit to this XRPL account.
+   *
+   * @set https://xrpl.org/depositpreauth.html
+   *
+   * @param senderXAddress The X-Address of the sender to enable DepositPreauth for.
+   * @param wallet The wallet associated with the XRPL account enabling DepositPreauth and that will sign the request.
+   */
+  public async authorizeDepositPreauth(
+    senderXAddress: string,
```
tedkalaw, line 470: i renamed this to `senderXAddress` from xAddressToAuthorize` to match with the docs but it's way more confusing and i regret this choice. i will change it back

src/XRP/default-xrp-client.ts
amiecorso: I also agree.  I would perhaps name them `authorizeSendingAccount` or `authorizePayer` / `unauthorizeSendingAccount` etc.   There's no reason we have to stick to "DepositPreauth" schema.

amiecorso (approved):
So the implementation of `authorizeDepositPreauth` looks good to me.  What's missing though is:
- adding this API to the `XrpClientDecorator`, `XrpClientInterface`, `XrpClient` and then potentially `ReliableSubmissionXrpClient` and `FakeXrpClient` because these implement the interfaces.  Maybe you're waiting to do this until you implement it for the reliable submission client?

src/XRP/default-xrp-client.ts
tedkalaw: good point @amiecorso - i'll try both of those out and see how it feels

tedkalaw:
yep! starting from here, but the next piece will be to get this going in reliablesubmission + add integration tests. i expect those will be bigger so wanted to make sure i was on the right track here and reduce noise further down the line

src/XRP/default-xrp-client.ts
```diff
@@ -445,6 +447,43 @@ export default class DefaultXrpClient implements XrpClientDecorator {
     return await this.getTransactionResult(transactionHash)
   }
 
+  /**
+   * Enables DepositPreauth and authorizes an XRPL account to send to this XRPL account.
+   *
+   * @set https://xrpl.org/depositpreauth.html
```
mvadari, line 453: typo - I assume this is supposed to be `@see`

mvadari (approved):
One small typo

src/XRP/default-xrp-client.ts
tedkalaw: got it

PR #725, opened 2020-09-23
Audits `XrpAccountSet` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpAccountSet`.

### Context of Change

### Type of Change

## Before / After

`XrpAccountSet` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,44 @@ export default class XrpAccountSet {
...
+    // transferRate cannot be more than 2000000000 or less than 1000000000,
+    // except for the special case 0 meaning no fee.
+    if (transferRate) {
+      if (transferRate > 2000000000) {
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'AccountSet protobuf field `transferRate` is above 2000000000.',
+        )
+      }
+      if (transferRate < 1000000000 && transferRate != 0) {
```
tedkalaw, line 43: can we put these in consts? can reuse in the msg as well

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,44 @@ export default class XrpAccountSet {
...
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'AccountSet protobuf field `transferRate` is below 1000000000.',
+        )
+      }
+    }
     const tickSize = accountSet.getTickSize()?.getValue()
-
+    // Valid values for tickSize are 3 to 15 inclusive, or 0 to disable.
+    if (tickSize && (tickSize < 3 || tickSize > 15) && tickSize != 0) {
```
tedkalaw, line 52: can we either put these numbers in const or perhaps add a function called `isValidTickSize`?

amiecorso (approved):
This looks great, fake protobufs are nicely named and organized.  Technically there's one line of the `isValidTickSize` method that's not being covered by tests (a tick size of 0), but I'm ok with that because the logic is definitely straightforward enough to reason about.

amiecorso (commented):
sorry, re-requested a review from myself because I had this question and I posed it to Elliot: "weird question for you Elliot:  any idea what happens if both the setFlag and a clearFlag fields are populated on an AccountSet transaction?  The interesting case would be where both fields are indicating the same flag… the order in which they are processed would matter.  However, there’s nothing in the documentation describing this order of execution or stating that both fields can’t be set at the same time…"

Will get back to you when I hear back.

amiecorso:
> sorry, re-requested a review from myself because I had this question and I posed it to Elliot: "weird question for you Elliot: any idea what happens if both the setFlag and a clearFlag fields are populated on an AccountSet transaction? The interesting case would be where both fields are indicating the same flag… the order in which they are processed would matter. However, there’s nothing in the documentation describing this order of execution or stating that both fields can’t be set at the same time…"
> 
> Will get back to you when I hear back.

Indeed, there is a check that Elliot found for me that doesn't allow set and clear for same flag on same transaction: https://github.com/ripple/rippled/blob/a8d481c2a54267d80bd539003f56481424b9ae63/src/ripple/app/tx/impl/SetAccount.cpp#L81.  Let's go ahead and add the same check here.   From what I understand, it's ok if both fields are set, but not if they indicate the same flag.

amiecorso (changes requested):
See comment about set and clear flag

src/XRP/model/xrp-account-set.ts
```diff
@@ -36,6 +78,19 @@ export default class XrpAccountSet {
     )
   }
 
+  private static isValidTickSize(tickSize: number) {
+    const minTickSize = 3
+    const maxTickSize = 15
+    const disableTickSize = 0
+    if (
+      (minTickSize <= tickSize && tickSize <= maxTickSize) ||
+      tickSize == disableTickSize
```
tedkalaw, line 87: can we just return this logical expression?

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,56 @@ export default class XrpAccountSet {
...
+    if (domain?.toLowerCase() != domain) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'AccountSet protobuf field `domain` is not lowercase.',
+      )
+    }
     const emailHash = accountSet.getEmailHash()?.getValue_asU8()
     const messageKey = accountSet.getMessageKey()?.getValue_asU8()
     const setFlag = accountSet.getSetFlag()?.getValue()
+    if (clearFlag && setFlag && clearFlag == setFlag) {
```
tedkalaw, line 33: can we use `===`? is type coercion here on purpose?

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,56 @@ export default class XrpAccountSet {
...
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'AccountSet protobuf fields `clearFlag` and `setFlag` are equal.',
+      )
+    }
     const transferRate = accountSet.getTransferRate()?.getValue()
+    // transferRate cannot be more than 2000000000 or less than 1000000000,
```
tedkalaw, line 40: i think you can remove this comment, the consts are well-named

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,56 @@ export default class XrpAccountSet {
    * @return an XrpAccountSet with its fields set via the analogous protobuf fields.
    * @see https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L100
    */
-  public static from(accountSet: AccountSet): XrpAccountSet | undefined {
+  public static from(accountSet: AccountSet): XrpAccountSet {
     const clearFlag = accountSet.getClearFlag()?.getValue()
     const domain = accountSet.getDomain()?.getValue()
+    // domain must be lowercase
```
tedkalaw, line 23: error message + logic explain this well imo

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,56 @@ export default class XrpAccountSet {
...
+        XrpErrorType.MalformedProtobuf,
+        'AccountSet protobuf fields `clearFlag` and `setFlag` are equal.',
+      )
+    }
     const transferRate = accountSet.getTransferRate()?.getValue()
+    // transferRate cannot be more than 2000000000 or less than 1000000000,
+    // except for the special case 0 meaning no fee.
+    const maxTransferRate = 2000000000
+    const minTransferRate = 1000000000
+    const specialCaseTransferRate = 0
```
tedkalaw, line 44: you can move these inside of the if below, since they're only relevant in that block

src/XRP/model/xrp-account-set.ts
mvadari: No, I just forgot that Javascript uses `===`

src/XRP/model/xrp-account-set.ts
```diff
@@ -16,15 +17,52 @@ export default class XrpAccountSet {
    * @return an XrpAccountSet with its fields set via the analogous protobuf fields.
    * @see https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L100
    */
-  public static from(accountSet: AccountSet): XrpAccountSet | undefined {
+  public static from(accountSet: AccountSet): XrpAccountSet {
     const clearFlag = accountSet.getClearFlag()?.getValue()
     const domain = accountSet.getDomain()?.getValue()
+    if (domain?.toLowerCase() != domain) {
```
tedkalaw, line 23: `!==` here as well

we should setup a lint rule to catch this...thought we had that on

tedkalaw (approved):

amiecorso (approved):
looks great to me, thanks for adding the additional set==clear test

PR #726, opened 2020-09-23
Audits `XrpDepositPreauth` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpDepositPreauth`.

### Context of Change

### Type of Change

## Before / After

`XrpDepositPreauth` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks

### DIFF
--- a/src/XRP/model/xrp-deposit-preauth.ts
+++ b/src/XRP/model/xrp-deposit-preauth.ts
@@ -35,6 +35,12 @@ export default class XrpDepositPreauth {
         undefined,
         xrplNetwork == XrplNetwork.Test || xrplNetwork == XrplNetwork.Dev,
       )
+      if (!authorizeXAddress) {
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'Cannot construct XAddress from DepositPreauth protobuf `authorize` field.',
+        )
+      }
       return new XrpDepositPreauth(authorizeXAddress, undefined)
     }
 
@@ -44,6 +50,12 @@ export default class XrpDepositPreauth {
         undefined,
         xrplNetwork == XrplNetwork.Test || xrplNetwork == XrplNetwork.Dev,
       )
+      if (!unauthorizeXAddress) {
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'Cannot construct XAddress from DepositPreauth protobuf `unauthorize` field.',
+        )
+      }
       return new XrpDepositPreauth(undefined, unauthorizeXAddress)
     }
 
--- a/src/XRP/model/xrp-deposit-preauth.ts
+++ b/src/XRP/model/xrp-deposit-preauth.ts
@@ -1,3 +1,4 @@
+import { XrpError, XrpErrorType } from '..'
 import { XrplNetwork } from 'xpring-common-js'
 import XrpUtils from '../xrp-utils'
 import { DepositPreauth } from '../Generated/web/org/xrpl/rpc/v1/transaction_pb'
@@ -27,6 +28,13 @@ export default class XrpDepositPreauth {
       .getUnauthorize()
       ?.getValue()
       ?.getAddress()
+    // You must provide either Authorize or Unauthorize, but not both.
+    if (authorize && unauthorize) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'DepositPreauth protobuf provides both `Authorize` and `Unauthorize` fields.',
+      )
+    }
 
     let authorizeXAddress
     let unauthorizeXAddress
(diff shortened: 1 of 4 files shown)


src/XRP/model/xrp-deposit-preauth.ts
```diff
@@ -27,6 +28,13 @@ export default class XrpDepositPreauth {
       .getUnauthorize()
       ?.getValue()
       ?.getAddress()
+    // You must provide either Authorize or Unauthorize, but not both.
+    if (authorize && unauthorize) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'DepositPreauth protobuf provides both `Authorize` and `Unauthorize` fields.',
+      )
+    }
```
mvadari, line 37: Not covered in tests because it appears that it's not actually reachable, since `DepositPreauth` doesn't let both `authorize` and `unauthorize` be set together.

tedkalaw: oh interesting - if you setup a protobuf with both fields set, it fails?

src/XRP/model/xrp-deposit-preauth.ts
```diff
@@ -42,6 +50,11 @@ export default class XrpDepositPreauth {
         undefined,
         xrplNetwork == XrplNetwork.Test || xrplNetwork == XrplNetwork.Dev,
       )
+    } else {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'DepositPreauth protobuf provides neither `Authorize` nor `Unauthorize` fields.',
+      )
```
tedkalaw, line 57: this seems reasonable.

if you want...i would also be down if you refactored the logic here. something like

```
if (authorize) {
  const authorizeXAddress = the thing it does
  return XrpDepositPreauth(authorizeXAddress, undefined)
}

if (unauthorize) {
  const unauthorizeXAddress = the thing it does
  return new XrpDepositPreauth(undefined, unauthorizeXAddress)
}

throw new XrpError(XrpErrorType.MalformedProtobuf
```

this tightens up variable space and simplifies the branching. totally up to you though, as i know you didn't write the original logic!

src/XRP/model/xrp-deposit-preauth.ts
```diff
@@ -21,12 +22,19 @@ export default class XrpDepositPreauth {
...
     depositPreauth: DepositPreauth,
     xrplNetwork: XrplNetwork,
-  ): XrpDepositPreauth | undefined {
+  ): XrpDepositPreauth {
     const authorize = depositPreauth.getAuthorize()?.getValue()?.getAddress()
     const unauthorize = depositPreauth
       .getUnauthorize()
       ?.getValue()
       ?.getAddress()
+    // You must provide either Authorize or Unauthorize, but not both.
```
tedkalaw, line 31: i think you can remove this - the logic + console message express this invariant clearly imo

tedkalaw (approved):

src/XRP/model/xrp-deposit-preauth.ts
mvadari: No, the second one you set just nulls the first one.

test/XRP/fakes/fake-xrp-transaction-type-protobufs.ts
```diff
@@ -433,6 +436,29 @@ testInvalidCheckCashProto.setAmount(testAmountProto)
...
+// Invalid DepositPreauth proto (neither authorize nor unauthorize)
+const testInvalidDepositPreauthProtoNoAuthUnauth = new DepositPreauth()
+
+// Invalid DepositPreauth proto (bad authorize)
+const testInvalidAccountAddressProto = new AccountAddress()
+testInvalidAccountAddressProto.setAddress(testInvalidDestination)
+
+const testInvalidAuthorizeProto = new Authorize()
+testInvalidAuthorizeProto.setValue(testInvalidAccountAddressProto)
+
```
amiecorso, line 448:
```suggestion

// Invalid DepositPreauth proto (bad unauthorize)
```

test/XRP/xrp-transaction-type-proto-conversion.test.ts
```diff
@@ -358,7 +361,40 @@ describe('Protobuf Conversions - Transaction Types', function (): void {
       undefined,
       true,
     )
-    assert.equal(depositPreauth?.unauthorizeXAddress, expectedXAddress)
+    assert.equal(depositPreauth.unauthorizeXAddress, expectedXAddress)
+  })
+
+  it('Convert DepositPreauth protobuf to XrpDepositPreauth object - neither authorize nor unauthorize', function (): void {
+    // GIVEN a DespoitPreauth protocol buffer neither authorize nor unauthorize field set.
```
amiecorso, line 368:
```suggestion
    // GIVEN a DepositPreauth protocol buffer neither authorize nor unauthorize field set.
```

amiecorso (approved):
looks great, agree with Ted's notes, just one small typo and missing comment

PR #729, opened 2020-09-24
Audits `XrpSigner` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpSigner`.

### Context of Change

### Type of Change

## Before / After

`XrpSigner` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


src/XRP/model/xrp-signer.ts
```diff
@@ -15,10 +16,28 @@ export default class XrpSigner {
    */
   public static from(signer: Signer): XrpSigner | undefined {
     const account = signer.getAccount()?.getValue()?.getAddress()
```
mvadari, line 18: Does this need to be converted to an `XAddress`?

amiecorso: good catch, yeah let's do that.  SDK philosophy is that all addresses that appear in SDK types should be X-Address encoded.  Be sure to update the description for the `account` param in the main constructor.  I think just adding ", encoded as an X-Address." to the end of the existing description should do.

I guess you'll probably have to update the existing test too.

amiecorso (changes requested):

src/XRP/model/xrp-signer.ts
```diff
@@ -15,10 +16,28 @@ export default class XrpSigner {
    */
   public static from(signer: Signer): XrpSigner | undefined {
```
amiecorso, line 17: I think we can maybe get rid of `| undefined` here too?

amiecorso (changes requested):

src/XRP/model/xrp-signer.ts
mvadari: Oops, missed that

tedkalaw (approved):

src/XRP/model/xrp-signer.ts
```diff
@@ -13,22 +16,50 @@ export default class XrpSigner {
...
+      )
+    }
     const transactionSignature = signer
       .getTransactionSignature()
       ?.getValue_asU8()
-    return new XrpSigner(account, signingPublicKey, transactionSignature)
+    if (!transactionSignature) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Signer protobuf is missing `TnxSignature` field.',
```
amiecorso, line 45:
```suggestion
        'Signer protobuf is missing `transactionSignature` field.',
```

amiecorso: i guess we decided long ago in SDK land that we don't abbreviate "transaction".  Unless this is actually how the proto is written?? in which case go for it, though I think it would maybe be spelled "txn" instead of "tnx"

amiecorso (approved):
few notes to take a look at but overall looks great, hitting approve

src/XRP/model/xrp-signer.ts
mvadari: It's how it was written in the docs. The `Tnx` was probably a typo though.

PR #733, opened 2020-09-25
Audits `XrpEscrowCancel` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpEscrowCancel`.

### Context of Change

### Type of Change

## Before / After

`XrpEscrowCancel` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


tedkalaw (approved):

src/XRP/model/xrp-escrow-cancel.ts
```diff
@@ -20,21 +21,33 @@ export default class XrpEscrowCancel {
...
     escrowCancel: EscrowCancel,
     xrplNetwork: XrplNetwork,
-  ): XrpEscrowCancel | undefined {
+  ): XrpEscrowCancel {
     const owner = escrowCancel.getOwner()?.getValue()?.getAddress()
     if (!owner) {
-      return undefined
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'EscrowCancel protobuf is missing valid `Owner` field.',
```
amiecorso, line 29:
```suggestion
        'EscrowCancel protobuf is missing valid `owner` field.',
```

amiecorso: I think all fields should be lower-case... I know I've missed a few of these in other PRs.  mostly an aesthetic thing, not a big deal and I actually don't care if you fix it, just noticing here due to the inconsistency with this `Owner` and the reference to the same field just below

amiecorso (approved):

src/XRP/model/xrp-escrow-cancel.ts
mvadari: Ah ok, I was capitalizing because that's how it was in the documentation, so I thought that would be better.

PR #739, opened 2020-09-28
Audits `XrpPaymentChannelClaim` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpPaymentChannelClaim`.

### Context of Change

### Type of Change

## Before / After

`XrpPaymentChannelClaim` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


src/XRP/model/xrp-payment-channel-claim.ts
```diff
@@ -42,6 +45,12 @@ export default class XrpPaymentChannelClaim {
       ?.getValue_asB64()
 
     const publicKey = paymentChannelClaim.getPublicKey()?.getValue_asB64()
+    if (signature && !publicKey) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'PaymentChannelClaim protobuf does not contain `publicKey` field when `signature` field is provided.',
+      )
+    }
```
mvadari, line 53: The `signature` is `Required unless the sender of the transaction is the source address of the channel.`
The `publicKey` is `Required unless the sender of the transaction is the source address of the channel and the Signature field is omitted.`
If I understand that correctly, if there is a `signature` there must be a `publicKey`

tedkalaw: yeah, it reads that way to me as well

tedkalaw (approved):

amiecorso (approved):

PR #740, opened 2020-09-28
Audits `XrpPayment` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpPayment`.

### Context of Change

### Type of Change

## Before / After

`XrpPayment` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


src/XRP/model/xrp-payment.ts
```diff
@@ -18,20 +19,23 @@ export default class XrpPayment {
...
-    if (!amount) {
-      return undefined // amount is required
+    if (!paymentAmountValue) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Payment protobuf is missing required `amount` field.',
+      )
     }
+    const amount = XrpCurrencyAmount.from(paymentAmountValue)
+    // For non-XRP amounts, the nested field names in `amount` are lower-case.
```
amiecorso, line 31: I think we can omit checks for all of these lower-case situations.  i'm not sure if there are any similar checks implemented in `XrpCurrencyAmount`, but if we one day decide to add them they'd probably go in that class anyway.

src/XRP/model/xrp-payment.ts
```diff
@@ -60,14 +58,29 @@ export default class XrpPayment {
...
-    if (paymentSendMaxValue && !sendMax) {
-      return undefined
+    if (sendMax && amount.drops) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Payment protobuf `sendMax` field should be omitted for XRP-to-XRP interactions.',
+      )
     }
+    // For non-XRP amounts, the nested field names in `sendMax` MUST be lower-case.
+    // `sendMax` must be supplied for cross-currency/cross-issue payments.
```
amiecorso, line 78: I believe you can implement this check in much the same way you did the `path` and XRP-to-XRP payments... I think we'd just ask about `amount.issuedCurrency` or whatever the equivalent is, && sendMax.  I think that any payment that's not XRP-to-XRP requires a sendMax... pretty sure about that.  We can always investigate and change later if need be.

amiecorso (approved):
really, really nice.  see inline notes about omitting case tests and potentially adding the sendMax check for issuedCurrency payments if it works out.  approving now

src/XRP/model/xrp-payment.ts
mvadari: `XrpCurrencyAmount` and `XrpIssuedCurrency` don't specify any checks like this that need to be implemented. 

I'll just leave in the comments for now, just so it's noted, without a TODO or anything.

src/XRP/model/xrp-payment.ts
```diff
@@ -18,20 +19,23 @@ export default class XrpPayment {
...
-    payment: Payment,
-    xrplNetwork: XrplNetwork,
-  ): XrpPayment | undefined {
+  public static from(payment: Payment, xrplNetwork: XrplNetwork): XrpPayment {
     const paymentAmountValue = payment.getAmount()?.getValue()
-    const amount =
-      paymentAmountValue && XrpCurrencyAmount.from(paymentAmountValue)
-    if (!amount) {
-      return undefined // amount is required
+    if (paymentAmountValue == undefined) {
```
tedkalaw, line 24: `===` pl0x

tedkalaw (changes requested):
looks good; let's just change to `===`. will add a task to track us adding this lint rule

src/XRP/model/xrp-payment.ts
```diff
@@ -60,14 +58,34 @@ export default class XrpPayment {
...
-    if (paymentSendMaxValue && !sendMax) {
-      return undefined
+    if (sendMax != undefined && amount.drops != undefined) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Payment protobuf `sendMax` field should be omitted for XRP-to-XRP interactions.',
+      )
     }
+    // For non-XRP amounts, the nested field names in `sendMax` MUST be lower-case.
+    if (sendMax == undefined && amount.issuedCurrency != undefined) {
```
tedkalaw, line 78: `===`/`!==`

tedkalaw (commented):
one more place

tedkalaw (approved):

PR #741, opened 2020-09-28
Creates `CoreXrplClient` and refactors `DefaultXrpClient`

amiecorso:
In preparation for the addition of a new client, the `IssuedCurrencyClient`, this refactor anticipates the fact that there is common, core functionality involved in interacting with the XRP Ledger that should be pushed to a more fundamental layer of abstraction, and shared among both (and any future) clients.

I tried to refactor the common reliable submission logic found in `ReliableSubmissionXrpClient` by moving it into `CoreXrplClient` and then adding an instance of a `CoreXrplClient`to `ReliableSubmissionXrpClient`... however, even after working through the complications this poses in the constructor of `XrpClient` (which constructs a `ReliableSubmissionXrpClient`) ... there was a nightmare in the test files, since `ReliableSubmissionXrpClient` is designed to be tested with an injected `DefaultXrpClient` ... as soon as it contains two clients, disaster ensues.  Will have to sleep on this.

### Context of Change
Preparation for introduction of IOU functionality and large design choices around that.

### Type of Change

## Before / After
Core functionality for interacting with XRP Ledger moved into `CoreXrplClient`

## Test Plan
No tests removed, CI still passes.


tedkalaw:
wait so this is still in progress?

src/XRP/common-xrpl-client.ts
```diff
@@ -0,0 +1,277 @@
...
+import TransactionResult from './model/transaction-result'
+import isNode from '../Common/utils'
+
+/** A margin to pad the current ledger sequence with when submitting transactions. */
+const maxLedgerVersionOffset = 10
+
+/**
+ * CommonXrplClient is a client which interacts with the XRP Ledger.
+ */
+export default class CommonXrplClient {
```
tedkalaw, line 30: what other shared functionality do you expect to be here? as of now, it seems like this is kind of an `XrplTransactionClient`, but i don't know enough about xrpl to know what else there will be

is there a better name for this class? we're doing this because we expect to implement a class for sending XRP transactions and one for IssuedCurrencies, correct?

i guess what i'm wondering is...could this just be called `XrplClient`? `XrpLedgerClient`? `LedgerClient`? the class used before to send transactions is called `XrpClient`, and its new counterpart is likely `IssuedCurrenciesClient`

src/XRP/common-xrpl-client.ts
```diff
@@ -0,0 +1,277 @@
...
+   *
+   * @param networkClient A network client which will manage remote RPCs to Rippled.
+   * @param network The network this XrpClient is connecting to.
+   */
+  public constructor(
+    public readonly networkClient: XrpNetworkClient,
+    readonly network: XrplNetwork,
+  ) {}
+
+  public async getOpenLedgerSequence(): Promise<number> {
```
tedkalaw, line 61: can we doc this and the other public functions?

src/XRP/default-xrp-client.ts
```diff
@@ -43,14 +38,14 @@ import { LedgerSpecifier } from './Generated/web/org/xrpl/rpc/v1/ledger_pb'
...
-
-/** A margin to pad the current ledger sequence with when submitting transactions. */
-const maxLedgerVersionOffset = 10
+import CommonXrplClient from './common-xrpl-client'
 
 /**
  * DefaultXrpClient is a client which interacts with the XRP Ledger.
  */
 export default class DefaultXrpClient implements XrpClientDecorator {
+  public commonXrplClient: CommonXrplClient
```
tedkalaw, line 47: can we make this private? seems bad if callers are using the commonXrplClient

src/XRP/default-xrp-client.ts
```diff
@@ -43,14 +38,14 @@ import { LedgerSpecifier } from './Generated/web/org/xrpl/rpc/v1/ledger_pb'
 import SendXrpDetails from './model/send-xrp-details'
 import { AccountSetFlag } from './model/account-set-flag'
 import TransactionResult from './model/transaction-result'
-
-/** A margin to pad the current ledger sequence with when submitting transactions. */
-const maxLedgerVersionOffset = 10
+import CommonXrplClient from './common-xrpl-client'
 
 /**
  * DefaultXrpClient is a client which interacts with the XRP Ledger.
```
tedkalaw, line 44: is there a more specific comment we can update this to? this is the same comment for CommonXrplClient

src/XRP/common-xrpl-client.ts
```diff
@@ -0,0 +1,277 @@
+/* eslint-disable no-restricted-syntax */
```
tedkalaw, line 1: what's this for

src/XRP/common-xrpl-client.ts
amiecorso: So my only hesitation in renaming to `XrpTransactionClient` is that there is some common functionality such as `account_info`, which has to do with more than just transactions... and I imagine that could grow in the future.  Also, I want to use the name `XrpLedgerClient` for the top-level client that is going to compose the `IssuedCurrencyClient` and the existing `XrpClient`  (assuming we go that route).  I think keeping "common" in the name does a better job indicating that this is a bottom-level client that deals with any and all shared xrpl infrastructure

src/XRP/common-xrpl-client.ts
amiecorso: mm, yep, meant to do that. doc'ed

src/XRP/default-xrp-client.ts
amiecorso: apparently we can!  the `DefaultXrpClient` is private by the time the public API is exposed but I agree that this is better practice.

src/XRP/default-xrp-client.ts
amiecorso: yep, updated both

src/XRP/common-xrpl-client.ts
amiecorso: not sure but I was able to remove it with no repercussions so it's gone in both clients now.. probably a relic of long ago

amiecorso:
changes made, not still in progress

src/XRP/common-xrpl-client.ts
tedkalaw: in the end i don't rly care since all this ends up being private

"Common" strikes me as a bit of an odd prefix for some reason, but gogogo. i maybe like Core better. i don't like Base because it implies some sort of class hierarchy. if i think of something better i'll suggest it one day

tedkalaw (commented):
is there anything we can do about the codecov drop? is there an easy way for us to defend it? if so, would be nice to reclaim some of the codecov to spend on other things

tedkalaw (approved):
seems good, leggo

i will leave it up to you to decide if the 0.11% codecov loss is worth it

src/XRP/common-xrpl-client.ts
amiecorso: renamed to `CoreXrplClient` - I think that's great

mvadari (approved):

PR #743, opened 2020-09-28
Audits `XrpPaymentChannelCreate` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpPaymentChannelCreate`.

### Context of Change

### Type of Change

## Before / After

`XrpPaymentChannelCreate` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks

### DIFF
--- a/src/XRP/model/xrp-payment-channel-create.ts
+++ b/src/XRP/model/xrp-payment-channel-create.ts
@@ -60,14 +60,14 @@ export default class XrpPaymentChannelCreate {
     if (settleDelay === undefined) {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
-        'PaymentChannelCreate protobuf does not contain `SettleDelay` field.',
+        'PaymentChannelCreate protobuf does not contain `settleDelay` field.',
       )
     }
     const publicKey = paymentChannelCreate.getPublicKey()?.getValue_asB64()
     if (!publicKey) {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
-        'PaymentChannelCreate protobuf does not contain `PublicKey` field.',
+        'PaymentChannelCreate protobuf does not contain `publicKey` field.',
       )
     }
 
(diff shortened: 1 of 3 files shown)


tedkalaw (approved):
do we need to handle the ambiguous falsy cases here as well?

src/XRP/model/xrp-payment-channel-create.ts
```diff
@@ -42,13 +49,26 @@ export default class XrpPaymentChannelCreate {
...
+        XrpErrorType.MalformedProtobuf,
+        'Cannot construct XAddress from PaymentChannelCreate protobuf `destination` and `destinationTag` fields.',
+      )
+    }
 
     const settleDelay = paymentChannelCreate.getSettleDelay()?.getValue()
+    if (settleDelay === undefined) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'PaymentChannelCreate protobuf does not contain `SettleDelay` field.',
```
amiecorso, line 63: lowercase SettleDelay maybe?  you can see that these fields are all always lower case https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L278

test/XRP/xrp-transaction-type-proto-conversion.test.ts
```diff
@@ -1006,16 +1010,59 @@ describe('Protobuf Conversions - Transaction Types', function (): void {
...
+    assert.throws(() => {
+      XrpPaymentChannelCreate.from(
+        testInvalidPaymentChannelCreateProtoBadDestination,
+        XrplNetwork.Test,
+      )
+    }, XrpError)
+  })
+
+  it('Convert PaymentChannelCreate protobuf to XrpPaymentChannelCreate object - missing PublicKey', function (): void {
+    // GIVEN a PaymentChannelCreate protocol buffer missing the PublicKey field.
```
amiecorso, line 1047: obvs an insignificant test comment, but just noticing the field casing here too and same for PublicKey

amiecorso (approved):
looks great just some of my same super nits about casing

PR #753, opened 2020-09-29
Corrects falsiness assumptions

mvadari:

When converting from GRPC classes to Typescript classes, the code often uses `if (!field)` to determine if an object is undefined. However, this catches other falsey cases as well, such as if the field equals `0` or `""`. This PR corrects those assumptions where needed. 

### Context of Change

This was noticed while auditing all the Typescript wrapper classes.

### Type of Change

## Before / After

Falsiness assumptions have been corrected. 

## Test Plan

CI passes. 

## Future Tasks
More corrections will be made in future auditing PRs.

### DIFF
--- a/src/XRP/model/xrp-account-set.ts
+++ b/src/XRP/model/xrp-account-set.ts
@@ -29,7 +29,11 @@ export default class XrpAccountSet {
     const emailHash = accountSet.getEmailHash()?.getValue_asU8()
     const messageKey = accountSet.getMessageKey()?.getValue_asU8()
     const setFlag = accountSet.getSetFlag()?.getValue()
-    if (clearFlag != undefined && setFlag && clearFlag === setFlag) {
+    if (
+      clearFlag != undefined &&
+      setFlag != undefined &&
+      clearFlag === setFlag
+    ) {
       throw new XrpError(
         XrpErrorType.MalformedProtobuf,
         'AccountSet protobuf fields `clearFlag` and `setFlag` are equal.',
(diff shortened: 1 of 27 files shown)


amiecorso (approved):
great catch on this

src/XRP/model/xrp-account-set.ts
```diff
@@ -29,7 +29,11 @@ export default class XrpAccountSet {
     const emailHash = accountSet.getEmailHash()?.getValue_asU8()
     const messageKey = accountSet.getMessageKey()?.getValue_asU8()
     const setFlag = accountSet.getSetFlag()?.getValue()
-    if (clearFlag != undefined && setFlag && clearFlag === setFlag) {
+    if (
+      clearFlag != undefined &&
+      setFlag != undefined &&
```
tedkalaw, line 34: `!==` please

src/XRP/model/xrp-currency-amount.ts
```diff
@@ -19,11 +19,11 @@ export default class XrpCurrencyAmount {
     // Mutually exclusive: either drops or issuedCurrency is set in an XRPCurrencyAmount
     const issuedCurrencyAmount = currencyAmount.getIssuedCurrencyAmount()
     const xrpAmount = currencyAmount.getXrpAmount()
-    if (issuedCurrencyAmount && !xrpAmount) {
+    if (issuedCurrencyAmount != undefined && xrpAmount == undefined) {
       const issuedCurrency = XrpIssuedCurrency.from(issuedCurrencyAmount)
       return new XrpCurrencyAmount(undefined, issuedCurrency)
     }
-    if (xrpAmount && !issuedCurrencyAmount) {
+    if (xrpAmount != undefined && issuedCurrencyAmount == undefined) {
```
tedkalaw, line 26: same with these `===` and `!==`

tedkalaw (changes requested):
can we use the `===` operators here instead of `==`? this should technically work because undefined is only equal to itself but afaik we don't use type coercion in the sdk for these checks

tedkalaw (approved):

PR #754, opened 2020-09-30
Expands `TransactionStatus` Options and Increases Reliable Submission Information

amiecorso:
This PR adds two members to the `TransactionStatus` enum: `LastLedgerPassed` and `MalformedTransaction` to increase the granularity of information delivered to the user of the SDK about the final outcome of their transaction.  It also adds a field `final` to the `TransactionResult` class to indicate whether the information there such as the `status` and `validated` fields are forever final, or potentially subject to change for that transaction.  Any `TransactionResult` returned from reliable submission methods will be able to set that field to `true`.

There is a small refactor in the reliable submission infrastructure that allows for passing the information of whether a transaction's `lastLedgerSequence` has been surpassed (making it impossible to ever be included in a validated ledger) in order for the correct `TransactionStatus` to be assigned in a `TransactionResult`.

### Context of Change
There are two distinct categories of outcome that we care about WRT a transaction on the XRPL:
1) whether it could be successfully, logically applied to the ledger - something that's evaluated in the open ledger 
    - was the transaction correctly formed? or is it malformed and thus permanently rejectable
    - did it fail for logical (business) reasons?  
    - is this failure deemed permanent, or potentially reversible if it's tried again in the future?
    - did it succeed?
2) whether it was actually included in a validated ledger
    - well-formed but failed transactions still get included in validated ledgers (burns txn cost)
    - successful/well-formed transactions may never be included in a validated ledger for other reasons (fee dynamics, passing of lastLedgerSequence etc.)

So determining the actual outcome of a transaction in the sense of **did my transaction officially exert its intended effect on the ledger?** and **why or why not?**  requires understanding both of the above categories.  The `FinalTransactionStatus` is intended to unequivocally summarize the first question of whether the effect was achieved.  This PR helps us improve our answering of the second question, why or why not.

### Type of Change

## Before / After
Now, the reliable submission client should NEVER return `FinalTransactionStatus.Pending` -- something that used to happen in the case where the transaction was *not* included in a validated ledger _because its lastLedgerSequence had passed_.  In this case, we return `FinalTransactionStatus.LastLedgerPassed` which is a failure outcome but with more detail.

## Test Plan
CI.  Update reliable submission tests to expect the right value.

## Future Tasks

Ted and I want to take a thoughtful look at how all of this infrastructure (including the pass through `rawTransactionStatus` is still necessary, or whether it can be simplified at all.


test/XRP/common-xrpl-client.test.ts
```diff
@@ -24,7 +24,7 @@ import {
   LastLedgerSequence,
 } from '../../src/XRP/Generated/node/org/xrpl/rpc/v1/common_pb'
 import { AccountRoot } from '../../src/XRP/Generated/node/org/xrpl/rpc/v1/ledger_objects_pb'
-import TransactionResult from '../../src/XRP/model/transaction-result'
+import TransactionResult from '../../src/XRP/model/final-transaction-result'
```
tedkalaw, line 27: shall we use `FinalTransactionResult` here?

test/Xpring/xpring-client.test.ts
```diff
@@ -8,7 +8,7 @@ import TransactionStatus from '../../src/XRP/transaction-status'
 import XpringClient from '../../src/Xpring/xpring-client'
 
 import XpringError from '../../src/Xpring/xpring-error'
-import TransactionResult from '../../src/XRP/model/transaction-result'
+import TransactionResult from '../../src/XRP/model/final-transaction-result'
```
tedkalaw, line 11: `FinalTransactionResult`?

test/XRP/fakes/fake-common-xrpl-client.ts
```diff
@@ -3,19 +3,25 @@ import { Wallet } from '../../../src/index'
...
 import RawTransactionStatus from '../../../src/XRP/raw-transaction-status'
-import TransactionResult from '../../../src/XRP/model/transaction-result'
+import TransactionResult from '../../../src/XRP/model/final-transaction-result'
 
 class FakeCommonXrplClient implements CommonXrplClientInterface {
   public constructor(
-    public awaitFinalTransactionStatusValue: Result<RawTransactionStatus>,
+    public awaitFinalTransactionStatusValue: Result<{
+      rawTransactionStatus: RawTransactionStatus
+      lastLedgerPassed: boolean
```
tedkalaw, line 12: might be useful to make a named type for this imo. alternatively, i guess you could do `ReturnType< typeof CommonXrpClientInterface.awaitFinalTransactionStatusValue>`

test/XRP/common-xrpl-client.test.ts
```diff
@@ -214,15 +214,10 @@ describe('Common XRPL Client', function (): void {
       wallet,
     )
 
-    // TODO: (amiecorso) in separate PR, update the generation of a `TransactionResult` such that:
-    // - consider renaming to FinalTransactionResult
```
tedkalaw, line 218: i *think* i'm ok with this - however, does this mean that we think there is no value in having `TransactionResult` represent non-finalized (intermediate) outcomes?

another way we could do this while maintaining that flexibility is adding a flag to `TransactionResult` called `isComplete` or something

tedkalaw (commented):
this is cool. i may build on top of this so i can just be done with this damn reliable depositPreauth stuff

test/XRP/common-xrpl-client.test.ts
amiecorso: yep, I think I updated this... not sure why it's not erroring

amiecorso:
> this is cool. i may build on top of this so i can just be done with this damn reliable depositPreauth stuff

That's awesome! was hoping you would.  I'll prob build on top too

test/XRP/fakes/fake-common-xrpl-client.ts
amiecorso: mm yeah I thought about that too... maybe I can make a little class right there in the `Core` client

test/XRP/common-xrpl-client.test.ts
amiecorso: I actually just realized this this morning that I think `TransactionResult` (now `FinalTransactionResult`) is actually used elsewhere in non-reliable-sub places... thus making it distinctly *not* final so lemme think about the best way to address that.

amiecorso: Ugh yeah so the *one* issue with calling this `FinalTransactionResult` is that this type is returned from methods in the `DefaultXrpClient`, which doesn't implement any reliable sending... It may be an internal class, but that becomes unacceptably ambiguous.  I think I need to just change the name back.... adding a flag to transactionResult is probably the move.

test/XRP/common-xrpl-client.test.ts
amiecorso: oh wait maybe I do... cause you get to name the default export whatever you want in typescript?

test/XRP/common-xrpl-client.test.ts
amiecorso: did all that

test/Xpring/xpring-client.test.ts
amiecorso: moved everything back to `TransactionResult` anyway

src/XRP/core-xrpl-client.ts
```diff
@@ -334,10 +341,21 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
    */
   private getFinalTransactionStatus(
     rawTransactionStatus: RawTransactionStatus,
+    lastLedgerPassed: boolean,
   ): TransactionStatus {
+    // I don't know if this is actually possible... just making the compiler happy in safest way
+    if (!rawTransactionStatus.transactionStatusCode) {
+      throw XrpError.malformedResponse
+    }
+    if (rawTransactionStatus.transactionStatusCode.startsWith('tem')) {
```
amiecorso, line 350: note that Ted's reliable sub deposit preauth stuff will hit this line

mvadari (approved):

src/XRP/core-xrpl-client.ts
```diff
@@ -313,15 +316,19 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
...
-    const rawTransactionStatus = await this.waitForFinalTransactionOutcome(
-      transactionHash,
-      wallet,
+    const {
+      rawTransactionStatus,
+      lastLedgerPassed,
+    } = await this.waitForFinalTransactionOutcome(transactionHash, wallet)
+    const finalStatus = this.getFinalTransactionStatus(
+      rawTransactionStatus,
+      lastLedgerPassed,
```
tedkalaw, line 325: do we really need two pieces of information to determine the transaction status? i'm looking at https://xrpl.org/transaction-results.html and it seems to be the case that...

`tec`, `tef`, `tel`, `tem`, and `ter` status codes can NEVER be included in the current ledger; that is, this bit will always be false.

`tes` status codes mean, by definition, that the transaction WAS included in the ledger

does using `lastLedgerPassed` provide more information than expanding TransactionResult to include the rest of the codes? alternatively, it seems like you could get the same logic by having a function called `isSuccessResult`. would this not cover all of the cases we want?

src/XRP/core-xrpl-client.ts
```diff
@@ -422,7 +443,10 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
       rawTransactionStatus = await this.getRawTransactionStatus(transactionHash)
     }
     /* eslint-enable no-await-in-loop */
-
-    return rawTransactionStatus
+    const lastLedgerPassed = latestLedgerSequence >= lastLedgerSequence
+    return {
+      rawTransactionStatus: rawTransactionStatus,
```
tedkalaw, line 448: is it possible for `rawTransactionStatus` to NOT start with `tes` and have `lastLedgerPassed === true`? 

my gut tells me that the value of `lastLedgerPassed` is dividing two disjoint sets in `rawTransactionStatus`

src/XRP/core-xrpl-client.ts
tedkalaw: another thought: what if we did

```
type FinalTransactionStatus = TransactionStatus.Succeeded | TransactionStatus.Failed | TransactionStatus.Malformed | TransactionStatus.Unknown
```

and then return that in `waitForFinalTransactionOutcome`? then we can abstract away the intermediate pending state

src/XRP/core-xrpl-client.ts
tedkalaw: put another way - i suspect that the trickiness here revolves around the fact that this function can return `TransactionStatus.Pending` at all, which is not a `FinalTransactionStatus`

src/XRP/core-xrpl-client.ts
```diff
@@ -266,11 +266,14 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
     const rawStatus = await this.getRawTransactionStatus(transactionHash)
     const isValidated = rawStatus.isValidated
     const transactionStatus = await this.getTransactionStatus(transactionHash)
-
+    const final = isValidated ? true : false
```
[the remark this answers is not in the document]
amiecorso: hah yep, a classic newb

src/XRP/core-xrpl-client.ts
tedkalaw: got it, per our offline discussion, distributed systems are hard

src/XRP/core-xrpl-client.ts
tedkalaw: got it the above comment doesn't make sense because `rawTransactionStatus` is a diff object. dang

src/XRP/core-xrpl-client.ts
tedkalaw: thx for the schoolin' on the ledger, learned a TON today.

my only remaining question is: do we expect to use this function in several places? if `lastLedgerPassed === true`, then `getFinalTransactionStatus` will always return `TransactionStatus.LastLedgerPassed`.

perhaps it would be less ambiguous if we did something like

```
const finalStatus = lastLedgerPassed ? TransactionStatus.LastLedgerPassed : this.getFinalTransactionStatus(rawTransactionStatus)
```

passing `lastLedgerPassed` to `getFinalTransactionStatus` implies that there's some combination of the two where the result can differ. nbd tho

tedkalaw (approved):
kewl

src/XRP/core-xrpl-client.ts
```diff
@@ -360,7 +378,10 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
   public async waitForFinalTransactionOutcome(
     transactionHash: string,
     sender: Wallet,
-  ): Promise<RawTransactionStatus> {
+  ): Promise<{
+    rawTransactionStatus: RawTransactionStatus
+    lastLedgerPassed: boolean
```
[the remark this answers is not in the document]
amiecorso: it does not, as discussed offline

src/XRP/transaction-status.ts
```diff
@@ -9,6 +12,9 @@ enum TransactionStatus {
   /** The transaction was included in a finalized ledger and succeeded. */
   Succeeded,
 
+  /** The transaction's last ledger sequence has been passed - the transaction will never be included in a validated ledger. */
+  LastLedgerPassed,
```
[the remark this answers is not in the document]
amiecorso: I'm going to call this `LastLedgerSequenceExpired` to make it more obvious that this is a permanent failure case.

src/XRP/model/transaction-result.ts
```diff
@@ -4,23 +4,17 @@ import TransactionStatus from '../transaction-status'
...
-  ) {
-    this.hash = hash
-    this.status = status
-    this.validated = validated
-  }
+    public readonly hash: string,
+    public readonly status: TransactionStatus,
+    public readonly validated: boolean,
+    public readonly final: boolean,
+  ) {}
```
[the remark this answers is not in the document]
amiecorso: As discussed offline, went for now with the addition of static constructors that remove the assignment of `final` and create more descriptive code.  We can refactor to an interface when we reassess the infrastructure at large.

src/XRP/core-xrpl-client.ts
```diff
@@ -334,10 +341,21 @@ export default class CoreXrplClient implements CoreXrplClientInterface {
    */
   private getFinalTransactionStatus(
     rawTransactionStatus: RawTransactionStatus,
+    lastLedgerPassed: boolean,
   ): TransactionStatus {
+    // I don't know if this is actually possible... just making the compiler happy in safest way
```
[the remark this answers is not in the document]
amiecorso: I don't know if we can (or at least, if we can in a worthwhile way) because we still need the granularity of investigating whether it starts with a `tes` or not later on in this function....  I decided to take care of this by altering `RawTransactionStatus` to throw an error if the transaction response doesn't contain a result code (therefore that field can't later be undefined), because that indeed should be an error case and we should catch it at the source.

src/XRP/core-xrpl-client.ts
amiecorso: Part one of these comments aren't true, but we discussed this offline.

amiecorso:
Agree with that last comment Ted... the ability to determine that `lastLedgerSequence` has passed resides in the reliable submission logic, and therefore the authority to determine that the final `TransactionStatus` is `.LastLedgerSequenceExpired` should really live there too.

PR #755, opened 2020-09-30
Audits `XrpTransaction` Protobuf Wrapper

mvadari:

This PR adds additional logic validations and descriptive error throwing to the class `XrpTransaction`.

### Context of Change

### Type of Change

## Before / After

`XrpTransaction` performs logic/error checking, based on the business logic in the docs.

## Test Plan

## Future Tasks


src/XRP/model/xrp-transaction.ts
```diff
@@ -67,39 +101,66 @@ export default class XrpTransaction {
...
 
-    const sourceTag = transaction.getSourceTag()?.getValue()
+    const signingPublicKey = transaction.getSigningPublicKey()?.getValue_asB64()
+    if (signingPublicKey == undefined) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Transaction protobuf is missing valid `signingPublicKey` field.',
+      )
+    }
+    if (!signers && signingPublicKey == '') {
```
tedkalaw, line 111: `===`

src/XRP/model/xrp-transaction.ts
```diff
@@ -67,39 +101,66 @@ export default class XrpTransaction {
         .map((signer) => XrpSigner.from(signer, xrplNetwork))
     }
 
-    const sourceTag = transaction.getSourceTag()?.getValue()
+    const signingPublicKey = transaction.getSigningPublicKey()?.getValue_asB64()
+    if (signingPublicKey == undefined) {
```
tedkalaw, line 105: `===`

src/XRP/model/xrp-transaction.ts
```diff
@@ -26,26 +27,59 @@ export default class XrpTransaction {
...
-    const signingPublicKey = transaction.getSigningPublicKey()?.getValue_asU8()
+    const fee = transaction.getFee()?.getDrops()
+    if (!fee) {
+      throw new XrpError(
+        XrpErrorType.MalformedProtobuf,
+        'Transaction protobuf is missing valid `fee` field.',
+      )
+    }
+    const intFee = Number(fee)
+    if (isNaN(intFee)) {
```
tedkalaw, line 69: is `getDrops` a strong? i think we want to use `BigInteger` here instead cc @amiecorso

src/XRP/default-xrp-client.ts
```diff
@@ -322,11 +322,7 @@ export default class DefaultXrpClient implements XrpClientDecorator {
             getTransactionResponse,
             this.network,
           )
-          if (!xrpTransaction) {
-            throw XrpError.paymentConversionFailure
-          } else {
-            payments.push(xrpTransaction)
-          }
+          payments.push(xrpTransaction)
```
tedkalaw, line 325: is this change related? it seems reasonable but can't tell

tedkalaw (changes requested):

src/XRP/model/xrp-transaction.ts
mvadari: I was leaving it stored as a `string` for now (`intFee` isn't used at all after this check), just using it as a temp variable to check if `fee` represents an integer value or not.

src/XRP/default-xrp-client.ts
mvadari: It's a result of `XrpTransaction.from()` not being able to return undefined anymore.

src/XRP/model/xrp-transaction.ts
amiecorso: https://howtodoinjava.com/typescript/equals-vs-strict-equals/  interesting, don't think I knew this.  What's the risk of using `==` specifically in the context of a comparison with `undefined`?  will it give erroneous results?  I'm wondering if there are other places where we need to update this.

Also, is there a typescript operator that compares not just for same type and same value, but same object?  kinda the way `===` essentially compares memory addresses in... ? java I think?

src/XRP/model/xrp-transaction.ts
```diff
@@ -67,39 +101,66 @@ export default class XrpTransaction {
...
-        }
-        type = payment && XrpTransactionType.Payment
+        paymentFields = XrpPayment.from(payment, xrplNetwork)
+        type = XrpTransactionType.Payment
         break
       }
+      case Transaction.TransactionDataCase.TRANSACTION_DATA_NOT_SET: {
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'Transaction protobuf is missing transaction data.',
```
amiecorso, line 146:
```suggestion
          'Transaction protobuf is missing transaction data case.',
```

src/XRP/model/xrp-transaction.ts
```diff
@@ -67,39 +101,66 @@ export default class XrpTransaction {
...
+        )
+      }
       default:
-        // Unsupported transaction type.
-        return undefined
+        throw new XrpError(
+          XrpErrorType.MalformedProtobuf,
+          'Transaction protobuf has an unsupported transaction type.',
```
amiecorso, line 152:
```suggestion
          'Transaction protobuf is of a transaction type unsupported by xpring-js.',
```

src/XRP/xrp-client-decorator.ts
```diff
@@ -115,7 +115,7 @@ export default interface XrpClientDecorator {
    * @throws An error if the transaction hash was invalid.
    * @returns An {@link XrpTransaction} object representing an XRP Ledger transaction.
    */
-  getPayment(transactionHash: string): Promise<XrpTransaction | undefined>
```
amiecorso, line 118: so satisfying

test/XRP/fakes/fake-xrp-protobufs.ts
```diff
@@ -467,6 +500,127 @@ testInvalidPaymentTransaction.setSigningPublicKey(signingPublicKeyProto)
 testInvalidPaymentTransaction.setTransactionSignature(transactionSignatureProto)
 testInvalidPaymentTransaction.setPayment(testInvalidPaymentProtoNoAmount)
 
+// ... no account
```
amiecorso, line 503: extremely great job on all these invalid test cases.  I'm going to be annoying and ask for one more:  can you make a fake transaction proto that's fine with the exception of a bad `Payment` as its `transaction` field?  I know we have a bad payment protobuf just laying around so hopefully it's simple enough.  It'll get us codecov on the one missing line.

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -651,13 +710,136 @@ describe('Protocol Buffer Conversion', function (): void {
...
+      XrpTransaction.from(
+        testInvalidGetTransactionResponseProtoNoSigningPublicKey,
+        XrplNetwork.Test,
+      )
+    }, XrpError)
+  })
+
+  it('Convert PAYMENT Transaction with no signers', function (): void {
+    // GIVEN a GetTransactionResponse protocol buffer with a Transaction field missing a signers field
+    // with an empty signingPublicKey field
```
amiecorso, line 801:
```suggestion
```

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -651,13 +710,136 @@ describe('Protocol Buffer Conversion', function (): void {
...
+      XrpTransaction.from(
+        testInvalidGetTransactionResponseProtoNoSignature,
+        XrplNetwork.Test,
+      )
+    }, XrpError)
+  })
+
+  it('Convert PAYMENT Transaction with no data', function (): void {
+    // GIVEN a GetTransactionResponse protocol buffer with a Transaction field missing a data field
+    // with an empty signingPublicKey field
```
amiecorso, line 824:
```suggestion
```

test/XRP/xrp-protocol-buffer-conversion.test.ts
```diff
@@ -651,13 +710,136 @@ describe('Protocol Buffer Conversion', function (): void {
...
+      XrpTransaction.from(
+        testInvalidGetTransactionResponseProtoNoData,
+        XrplNetwork.Test,
+      )
+    }, XrpError)
+  })
+
+  it('Convert PAYMENT Transaction with no hash', function (): void {
+    // GIVEN a GetTransactionResponse protocol buffer with a Transaction field missing a hash field
+    // with an empty signingPublicKey field
```
amiecorso, line 836:
```suggestion
```

amiecorso (changes requested):
lookin good! just a few requests

amiecorso:
considering pinging us on slack when you've updated the PR because I've found that "re-request review" often fails to put the PR back in our queue...  really annoying

src/XRP/model/xrp-transaction.ts
tedkalaw: seems reasonable to me - just wanted to make sure there wasn't a "standard" way we validate amounts elsewhere. gogogo

src/XRP/model/xrp-transaction.ts
tedkalaw: > https://howtodoinjava.com/typescript/equals-vs-strict-equals/ interesting, don't think I knew this. What's the risk of using == specifically in the context of a comparison with undefined? will it give erroneous results? I'm wondering if there are other places where we need to update this.

in this case, probably nothing - the only thing that `== undefined` is `null`. however, `==` does have some interesting behaviors: https://www.ecma-international.org/ecma-262/5.1/#sec-11.9.3

as such it's pretty common to turn on linting (https://eslint.org/docs/rules/eqeqeq) and have only one equality operator

> Also, is there a typescript operator that compares not just for same type and same value, but same object? kinda the way === essentially compares memory addresses in... ? java I think?

`===` compares memory addresses!! this is really important to how libraries like `React` and other functional and functional-y libraries work, particularly where places where changes need to be listened for

```
const obj1 = {}
const obj2 = {}
obj1 === obj2
> false
const obj1Ref = obj1
obj1 === obj1Ref
> true

src/XRP/model/xrp-transaction.ts
mvadari: My understanding of how the protobuf object works is that the data case gets set automatically, based on which data it has. So if it has no data case, then it's because it has no data.

test/XRP/fakes/fake-xrp-protobufs.ts
mvadari: It won't, because a bad `Payment` will trigger an error in `XrpPayment`. The one missing line is unreachable, but I can't think of a way to get rid of it, because as far as Typescript is aware, `payment` can still be `undefined`, even though the protobuf object won't let that be true when the data case is `Payment`.

src/XRP/model/xrp-transaction.ts
mvadari: Haven't seen this sort of check being applied anywhere, so idk if there's a "standard" way

amiecorso: yeah nope there's not

src/XRP/model/xrp-transaction.ts
amiecorso: I think you're right about that, but I added this word just to make it clear that it's not the "transaction" field that's missing, but rather the type... since this is mostly human-facing.  Either way works for me

test/XRP/fakes/fake-xrp-protobufs.ts
amiecorso: Ohhh I see.  I guess one move would be to remove that check entirely because it probably is an impossible case logically, but that seems less defensive and unnecessary.  We can take the codecov hit

amiecorso (approved):
lookin good to me

test/XRP/fakes/fake-xrp-protobufs.ts
mvadari: Unfortunately Typescript gets mad if it's removed, since it could be `undefined`, so I don't think it can be removed.

tedkalaw (approved):

PR #756, opened 2020-10-01
Makes fakedRawTransactionStatusValue reusable in test

tedkalaw:
Creates a new `fakedRawTransactionStatusValue` at the beginning of every test run.

### Context of Change
I was updating ReliableSubmissionXrpClient.test for the reliable authorizeSendingAccount. I noticed that our tests use `fakedRawTransactionStatusValue` as a means of simulating a reliable submission happening by mutating a field, but we never actually reset this value. This field is injected pretty high up the stack, so if we want to reuse it, we need to reset it in the `beforeEach`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
The enableDepositAuth test should test as intended now.

## Test Plan
CI


test/XRP/reliable-submission-xrp-client.test.ts
```diff
@@ -53,6 +43,17 @@ describe('Reliable Submission XRP Client', function (): void {
       fakedGetPaymentValue,
       fakedTransactionResultValue,
     )
+
+    this.fakedRawTransactionStatusValue = new RawTransactionStatus(
```
tedkalaw, line 47: could have also set `fakedRawTransactionStatusValue.validated = false` in here but the state changes are already interesting to reason about...felt better to just make a new one every time.

amiecorso (approved):
yoooo, thanks for taking care of this.  I noticed this too and the promptly forgot about it.  Seems reasonable.

mvadari (approved):

PR #764, opened 2020-10-03
Organizes XRP Directory in `xpring-js`

amiecorso:
`src/XRP` is getting a bit unwieldy, and we're about to add even more.  This PR creates better organization in preparation for `IssuedCurrencyClient`, including:
- a rename of `XRP/model` dir to `XRP/protobuf-wrappers`
- an updating of the `index.ts` file in `XRP/protobuf-wrappers` to include ALL the wrappers (don't worry, this doesn't go top-level until we export from a higher index)
- moving some things that were in `model` but were NOT protobuf wrappers into `XRP/shared`
- moving some things that were at top level in `XRP` but were NOT clients into `XRP/shared`
- `XRP/shared` now contains shared low-level infra such as flag enums, `TransactionStatus`, `RawTransactionStatus`, `TransactionResult`, `SendXrpDetails`, `XrpError`, etc.
- creating a new dir `XRP/network-clients` for the network clients (especially as we may add another one to handle the JSON RPC in our necessary hybrid approach going forward.)
- the top-level `XRP` dir now only contains clients/decorators/interfaces
- a few more index updates

Totally open to better names or schemes.
Not expecting a detailed review of 69 files - the fact that the thing compiles and existing tests pass means all the import wires were correctly updated.

### Context of Change
It was time.

### Type of Change

## Before / After
Less eye fatigue.

## Test Plan
CI still passes


src/XRP/protobuf-wrappers/index.ts
```diff
@@ -0,0 +1,27 @@
+export { default as XrpAccountDelete } from './xrp-account-delete'
```
mvadari, line 1: Does this not import correctly as `export XrpAccountSet from './xrp-account-set'`?

src/XRP/shared/index.ts
```diff
@@ -0,0 +1,10 @@
+export { default as AccountRootFlag } from './account-root-flags'
```
mvadari, line 1: same question here

src/XRP/protobuf-wrappers/index.ts
amiecorso: hmm, y'know it actually doesn't seem to.  It gives me a "cannot find name XrpAccountSet" ...  Typescript Ted - can you weight in here?  am I doing these index files right?

tedkalaw: i think that's how it goes. afaict it replaces the need to do an import + an export.

https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export

tedkalaw (approved):

mvadari (approved):
nice

PR #757, opened 2020-10-01
Adds authorizeSendingAccount to ReliableSubmissionXrpClient + Decorator

tedkalaw:
## High Level Overview of Change
This PR adds a basic reliable implementation of `authorizeSendingAccount` to `ReliableSubmissionXrpClient` and `XrpDecorator`.

Note that thorough integration testing will happen in the next PR, when this behavior is exposed in `XrpClient`.

### Context of Change
We ultimately want to expose `authorizeSendingAccount` in `XrpClient` to consumers, where we will actually do an integration test to see that things are working. However, we need to propagate things up by updating `XrpClientDecorator`, related tests and, of course, adding the actual implementation.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

## Before / After
ReliableSubmissionXrpClient has authorizeSendingAccount

## Test Plan
CI, the test I added

PR #774, opened 2020-10-06
Adds unauthorizeSendingAccount to ReliableSubmissionXrpClient + Decorator

tedkalaw:
This PR adds a basic reliable implementation of unauthorizeSendingAccount to ReliableSubmissionXrpClient and XrpDecorator.

Note that thorough integration testing will happen in the next PR, when this behavior is exposed in XrpClient.

### Context of Change
Unauthorize analog to https://github.com/xpring-eng/Xpring-JS/pull/757

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
ReliableSubmissionXrpClient has unauthorizeSendingAccount

## Test Plan
CI, the test I added


test/XRP/fakes/fake-xrp-client.ts
```diff
@@ -18,6 +18,7 @@ class FakeXrpClient implements XrpClientDecorator {
     public getPaymentValue: Result<XrpTransaction>,
     public enableDepositAuthValue: Result<TransactionResult>,
     public authorizeSendingAccountValue: Result<TransactionResult>,
+    public unauthorizeSendingAccountValue: Result<TransactionResult>,
```
tedkalaw, line 21: in https://github.com/xpring-eng/Xpring-JS/pull/757 we discussed just having a `this.successfulTransaction`, but that wouldn't quite work here - it would mean that we would need to explicitly call `this.successfulTransaction` in our functions, which would remove the benefits of DI. i know the constructor sux but seems like this is ok for now. 

in the future i propose we update this constructor to take an object where the defaults for each of the functions is a lambda that throws so that callers can only DI the things they care about

amiecorso: oh I see what you mean... it would remove our ability to alter the individual return values in the case that we don't want them all to be successful, thus defeating the purpose of direct injection (DI?) .

amiecorso (approved):
beauty

mvadari (approved):

PR #778, opened 2020-10-07
Creates `IssuedCurrencyClient` and corresponding test file.

amiecorso:
This PR creates the skeleton of an `IssuedCurrencyClient` upon which we can begin implementing the required methods for handling Issued Currencies on the XRPL.  This is modeled closely after the `DefaultXrpClient`.  Note that it doesn't implement an interface or a decorator... We're going to begin by implementing methods here, and make larger design choices in the future when it comes time to expose these methods publicly and in conjunction with the existing `XrpClient` methods.

### Context of Change
IOUs gogogo!

### Type of Change

## Test Plan
CI

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -1,0 +1,53 @@
+import { XrplNetwork } from 'xpring-common-js'
+import GrpcNetworkClient from './network-clients/grpc-xrp-network-client'
+import GrpcNetworkClientWeb from './network-clients/grpc-xrp-network-client.web'
+import { XrpNetworkClient } from './network-clients/xrp-network-client'
+import isNode from '../Common/utils'
+import CoreXrplClient from './core-xrpl-client'
+
+/**
+ * IssuedCurrencyClient is a client for working with Issued Currencies on the XRPL.
+ * @see https://xrpl.org/issued-currencies-overview.html
+ */
+export default class IssuedCurrencyClient {
+  private coreXrplClient: CoreXrplClient
+
+  /**
+   * Create a new IssuedCurrencyClient.
+   *
+   * The IssuedCurrencyClient will use gRPC to communicate with the given endpoint.
+   *
+   * @param grpcUrl The URL of the gRPC instance to connect to.
+   * @param network The network this IssuedCurrencyClient is connecting to.
+   * @param forceWeb If `true`, then we will use the gRPC-Web client even when on Node. Defaults to false. This is mainly for testing and in the future will be removed when we have browser testing.
+   */
+  public static issuedCurrencyClientWithEndpoint(
+    grpcUrl: string,
+    network: XrplNetwork,
+    forceWeb = false,
+  ): IssuedCurrencyClient {
+    return isNode() && !forceWeb
+      ? new IssuedCurrencyClient(new GrpcNetworkClient(grpcUrl), network)
+      : new IssuedCurrencyClient(new GrpcNetworkClientWeb(grpcUrl), network)
+  }
+
+  /**
+   * Create a new IssuedCurrencyClient with a custom network client implementation.
+   *
+   * In general, clients should prefer to call `issuedCurrencyClientWithEndpoint`. This constructor is provided to improve testability of this class.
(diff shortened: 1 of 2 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -0,0 +1,53 @@
...
+   * @param network The network this IssuedCurrencyClient is connecting to.
+   */
+  public constructor(
+    private readonly networkClient: XrpNetworkClient,
+    readonly network: XrplNetwork,
+  ) {
+    this.coreXrplClient = new CoreXrplClient(networkClient, network)
+  }
+
+  public shutUpCompiler(): void {
```
tedkalaw, line 49: it's getting mad bc this stuff is unused right? you could use `/* eslint-disable no-unused-vars */ ` instead, although i expect that these are going to get used pretty much immediately

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,22 @@
+import { XrplNetwork } from 'xpring-common-js'
+import IssuedCurrencyClient from '../../src/XRP/issued-currency-client'
+import { FakeXRPNetworkClient } from './fakes/fake-xrp-network-client'
+import 'mocha'
+
+const fakeSucceedingNetworkClient = new FakeXRPNetworkClient()
+// const fakeErroringNetworkClient = new FakeXRPNetworkClient(
+//   FakeXRPNetworkClientResponses.defaultErrorResponses,
+// )
```
tedkalaw, line 9: what's this for

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('Default XRP Client', function (): void {
+  it('Dummy description', function (): void {
+    // GIVEN an IssuedCurrencyClient.
+    const issuedCurrencyClient = new IssuedCurrencyClient(
+      fakeSucceedingNetworkClient,
+      XrplNetwork.Test,
+    )
+
+    // WHEN the compiler is told to shut up THEN everything is fine.
```
tedkalaw, line 19: lol

tedkalaw (approved):
seems reasonable...would be down to silence eslint here intstead. let's be diligent about figuring out if we need the extra level of indirection. my sense is we'll be ok here since we get dependency injection through the network client maybe

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import IssuedCurrencyClient from '../../src/XRP/issued-currency-client'
+import { FakeXRPNetworkClient } from './fakes/fake-xrp-network-client'
+import 'mocha'
+
+const fakeSucceedingNetworkClient = new FakeXRPNetworkClient()
+// const fakeErroringNetworkClient = new FakeXRPNetworkClient(
+//   FakeXRPNetworkClientResponses.defaultErrorResponses,
+// )
+
+describe('Default XRP Client', function (): void {
```
mvadari, line 11: Should be renamed

mvadari (approved):

test/XRP/issued-currency-client.test.ts
amiecorso: ah thank you

src/XRP/issued-currency-client.ts
amiecorso: ah yeah good idea I'll just get around with ESLint so that I'm not merging garbage to master

amiecorso: dude I can't figure out eslint so i'm just going with it and Mayukha can remove this in her PR

test/XRP/issued-currency-client.test.ts
amiecorso: that's going to be needed shortly for the unit tests so I left the infra in

src/XRP/issued-currency-client.ts
mvadari: I can't remove it entirely :(, just one of the lines (I don't use the network client at all)

PR #779, opened 2020-10-07
Implements `requireAuthorizedTrustlines` on `IssuedCurrencyClient`

mvadari:

This PR adds the method `requireAuthorizedTrustlines` to the `IssuedCurrencyClient`, which enables Require Authorization on the specified account. This is achieved by submitting an AccountSet transaction with a set flag set to 2 (see https://xrpl.org/accountset.html#accountset-flags).

### Context of Change

Part of the IOU effort. 

### Type of Change

## Before / After

`requireAuthorizedTrustlines` exposed on `IssuedCurrencyClient`

## Test Plan

CI passes. Wrote unit tests for the function that has full codecov and passes.

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -76,7 +76,10 @@ export default class IssuedCurrencyClient {
       wallet,
     )
 
-    return await this.coreXrplClient.getTransactionResult(transactionHash)
+    return await this.coreXrplClient.getFinalTransactionResultAsync(
+      transactionHash,
+      wallet,
+    )
   }
 
   public shutUpCompiler(): void {
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -57,7 +57,7 @@ export default class IssuedCurrencyClient {
    *
    * @param wallet The wallet associated with the XRPL account enabling Require Authorization and that will sign the request.
    * @returns A promise which resolves to a TransactionResult object that contains the hash of the submitted AccountSet transaction,
-   *          the preliminary status, and whether the transaction has been included in a validated ledger yet.
+   *          the final status, and whether the transaction was included in a validated ledger.
    */
   public async requireAuthorizedTrustlines(
     wallet: Wallet,
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -56,8 +56,7 @@ export default class IssuedCurrencyClient {
    * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#require-auth
    *
    * @param wallet The wallet associated with the XRPL account enabling Require Authorization and that will sign the request.
-   * @returns A promise which resolves to a TransactionResult object that contains the hash of the submitted AccountSet transaction,
-   *          the final status, and whether the transaction was included in a validated ledger.
+   * @returns A promise which resolves to a TransactionResult object that represents the result of this transaction.
    */
   public async requireAuthorizedTrustlines(
     wallet: Wallet,
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -44,7 +44,7 @@ export default class IssuedCurrencyClient {
    * @param network The network this IssuedCurrencyClient is connecting to.
    */
   public constructor(
-    private readonly networkClient: XrpNetworkClient,
+    networkClient: XrpNetworkClient,
     readonly network: XrplNetwork,
   ) {
     this.coreXrplClient = new CoreXrplClient(networkClient, network)
@@ -80,9 +80,5 @@ export default class IssuedCurrencyClient {
       transactionHash,
       wallet,
     )
-  }
-
-  public shutUpCompiler(): void {
-    console.log(this.networkClient)
   }
 }
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -84,6 +84,5 @@ export default class IssuedCurrencyClient {
 
   public shutUpCompiler(): void {
     console.log(this.networkClient)
-    console.log(this.coreXrplClient)
   }
 }
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -1,0 +1,53 @@
+import { XrplNetwork } from 'xpring-common-js'
+import GrpcNetworkClient from './network-clients/grpc-xrp-network-client'
+import GrpcNetworkClientWeb from './network-clients/grpc-xrp-network-client.web'
+import { XrpNetworkClient } from './network-clients/xrp-network-client'
+import isNode from '../Common/utils'
+import CoreXrplClient from './core-xrpl-client'
+
+/**
+ * IssuedCurrencyClient is a client for working with Issued Currencies on the XRPL.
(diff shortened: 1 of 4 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Create a new IssuedCurrencyClient with a custom network client implementation.
+   *
+   * In general, clients should prefer to call `issuedCurrencyClientWithEndpoint`. This constructor is provided to improve testability of this class.
+   *
+   * @param networkClient A network client which will manage remote RPCs to Rippled.
+   * @param network The network this IssuedCurrencyClient is connecting to.
+   */
+  public constructor(
+    private readonly networkClient: XrpNetworkClient,
```
amiecorso, line 47:
```suggestion
    networkClient: XrpNetworkClient,
```

amiecorso: I realized that, at least for the moment, we don't actually need direct access to the gRPC network client in this class - we just need it to feed to the `CoreXrplClient` constructor.. so by removing it as an instance variable we can get rid of everything related to shutUpCompiler.

src/XRP/issued-currency-client.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Enable Require Authorization for this XRPL account.
+   *
+   * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#require-auth
+   *
+   * @param wallet The wallet associated with the XRPL account enabling Require Authorization and that will sign the request.
+   * @returns A promise which resolves to a TransactionResult object that contains the hash of the submitted AccountSet transaction,
+   *          the preliminary status, and whether the transaction has been included in a validated ledger yet.
```
amiecorso, line 60:
```suggestion
   *          the final status, and whether the transaction was included in a validated ledger.
```

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    // WHEN the compiler is told to shut up THEN everything is fine.
+    issuedCurrencyClient.shutUpCompiler()
+  })
+  it('Enable Deposit Auth - successful response', async function (): Promise<
+    void
+  > {
+    // GIVEN a IssuedCurrencyClient with mocked networking that will return a successful hash for submitTransaction
+    const xrpClient = new IssuedCurrencyClient(
```
amiecorso, line 32:
```suggestion
    const issuedCurrencyClient = new IssuedCurrencyClient(
```

amiecorso: just to keep everything straight

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      fakeSucceedingNetworkClient,
+      XrplNetwork.Test,
+    )
+
+    const wallet = (await walletFactory.generateRandomWallet())!.wallet
```
amiecorso, line 37: In the unit tests, this is something that we can perform once in a `beforeAll` method since the generated account won't actually be manipulated in any way.

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    // GIVEN an IssuedCurrencyClient.
+    const issuedCurrencyClient = new IssuedCurrencyClient(
+      fakeSucceedingNetworkClient,
+      XrplNetwork.Test,
+    )
+
+    // WHEN the compiler is told to shut up THEN everything is fine.
+    issuedCurrencyClient.shutUpCompiler()
+  })
+  it('Enable Deposit Auth - successful response', async function (): Promise<
```
amiecorso, line 28:
```suggestion
  it('requireAuthorizedTrustlines - successful response', async function (): Promise<
```

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    // THEN a transaction hash exists and is the expected hash
+    const expectedTransactionHash = Utils.toHex(
+      FakeXRPNetworkClientResponses.defaultSubmitTransactionResponse().getHash_asU8(),
+    )
+
+    assert.exists(transactionHash)
+    assert.strictEqual(transactionHash, expectedTransactionHash)
+  })
+
+  it('Enable Deposit Auth - submission failure', async function (): Promise<
```
amiecorso, line 52:
```suggestion
  it('requireAuthorizedTrustlines - submission failure', async function (): Promise<
```

test/XRP/issued-currency-client.test.ts
amiecorso: Not sure if we want to call this "`Require Authorized Trust Lines`" the way Keefer styled the `DefaultXrpClient` tests... or reflect the method name directly like I've done here... ?  the method name directly always made more sense to me.

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    void
+  > {
+    // GIVEN a IssuedCurrencyClient which will fail to submit a transaction.
```
amiecorso, line 55:
```suggestion
    // GIVEN an IssuedCurrencyClient which will fail to submit a transaction.
```

amiecorso: I'm weird about this because otherwise we sound so dumb in the mind's ear

test/XRP/issued-currency-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      XrplNetwork.Test,
+    )
+
+    // WHEN the compiler is told to shut up THEN everything is fine.
+    issuedCurrencyClient.shutUpCompiler()
+  })
+  it('Enable Deposit Auth - successful response', async function (): Promise<
+    void
+  > {
+    // GIVEN a IssuedCurrencyClient with mocked networking that will return a successful hash for submitTransaction
```
amiecorso, line 31:
```suggestion
    // GIVEN an IssuedCurrencyClient with mocked networking that will return a successful hash for submitTransaction
```

amiecorso (changes requested):
This is off to a great start - some comments, mostly minor.  The other piece of this will be to write some integration tests as well...  Since we haven't decided exactly how we're going to structure our top-level clients, we can start by just creating an integration test file called `issued-currency-client-integration.test.ts` and creating some integration tests along the lines of what's in `xrp-client-integration.test.ts`.  Take a look at what's done for `enableDepositAuth` -- the integration test will basically involve making the real request to testnet, and then querying the account for it's account info and checking that the correct flag has actually been set.

src/XRP/issued-currency-client.ts
mvadari: oh lit

tedkalaw (approved):
nice!

src/XRP/shared/account-root-flags.ts
```diff
@@ -10,7 +10,15 @@
  * @see https://xrpl.org/accountroot.html#accountroot-flags
  */
 class AccountRootFlags {
-  static LSF_DEPOSIT_AUTH = 1 << 24
+  static LSF_DEPOSIT_AUTH = 1 << 24 // 0x01000000, 16777216
```
amiecorso, line 13: nice thanks for adding all

amiecorso (approved):
Looks great! nice job on the integration test

PR #780, opened 2020-10-07
Creates a new `JsonRpcNetworkClient`

amiecorso:
This PR creates a new network client for use with the rippled JSON RPC. 

Check out my comment below please.

### Context of Change
While the original vision for the SDK was to use gRPC for all networking needs, there is missing functionality from the current gRPC implementation, and it's likely that this functionality will never be added / gRPC may be deprecated at some point.  We can use this JSON client to hit the JSON RPC directly for needs we may have that aren't available in gRPC, such as `account_lines` and `gateway_balances` in the context of Issued Currencies.

### Type of Change

## Test Plan
This client will be put through its paces in subsequent PRs that use it to make real network requests.

### DIFF
--- a/src/XRP/network-clients/json-rpc-network-client.ts
+++ b/src/XRP/network-clients/json-rpc-network-client.ts
@@ -19,12 +19,12 @@ export default class JsonRpcNetworkClient {
   }
 
   public async submitRequest(
-    requestObject: JsonRequest,
+    jsonRequest: JsonRequest,
   ): Promise<AxiosResponse<unknown>> {
     const requestOptions: AxiosRequestConfig = {
       url: '/',
       method: 'post',
-      data: requestObject,
+      data: jsonRequest,
       headers: { 'Content-Type': 'application/json' },
     }
     return await this.axiosInstance.request(requestOptions)
--- a/src/XRP/network-clients/json-rpc-network-client.ts
+++ b/src/XRP/network-clients/json-rpc-network-client.ts
@@ -12,12 +12,25 @@ interface JsonRequest {
 export default class JsonRpcNetworkClient {
   private readonly axiosInstance: AxiosInstance
 
+  /**
+   * Create a new JsonRpcNetworkClient.
+   *
+   * @param jsonUrl The URL of the rippled node to query.
+   * @see https://xrpl.org/get-started-with-the-rippled-api.html
+   */
   public constructor(jsonUrl: string) {
     this.axiosInstance = axios.create({
       baseURL: jsonUrl,
     })
   }
 
+  /**
+   * Submit a JSON request to the rippled server.
+   *
+   * @param jsonRequest A JSON object specifying the method and params of the request.
+   * @see https://xrpl.org/request-formatting.html#json-rpc-format
+   * @returns The response from the rippled server.
+   */
   public async submitRequest(
     jsonRequest: JsonRequest,
   ): Promise<AxiosResponse<unknown>> {
--- a/src/XRP/network-clients/json-rpc-network-client.ts
+++ b/src/XRP/network-clients/json-rpc-network-client.ts
@@ -1,0 +1,32 @@
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios'
+
+interface JsonRequest {
+  method: string
+  params: unknown
+}
+
+/**
+ * A network client for interacting with the rippled JSON RPC.
+ * @see https://xrpl.org/rippled-api.html
+ */
+export default class JsonRpcNetworkClient {
+  private readonly axiosInstance: AxiosInstance
+
+  public constructor(jsonUrl: string) {
+    this.axiosInstance = axios.create({
+      baseURL: jsonUrl,
+    })
+  }
+
+  public async submitRequest(
+    requestObject: JsonRequest,
+  ): Promise<AxiosResponse<unknown>> {
+    const requestOptions: AxiosRequestConfig = {
+      url: '/',
+      method: 'post',
+      data: requestObject,
+      headers: { 'Content-Type': 'application/json' },
+    }
+    return await this.axiosInstance.request(requestOptions)
+  }
+}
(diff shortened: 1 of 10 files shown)


amiecorso:
Is security a thing we need to do anything about?
Are my `unknown` types ok? 
Do we want to create a network client pyramid using interfaces or whatever where our ledger clients just have to have ONE network client, which in turn has several network clients, thus abstracting that fact away?  Or does it not matter because 1) the implementations of our ledger clients are private anyway so it doesn't really matter if they have more than one network client (which would only be true for `IssuedCurrencyClient` anyway) and 2) we will still have to provide the option for input of two different URLs upon instantiation of a top-level client in the SDK because gRPC and JSON RPC use different ports so...

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios'
+
+interface JsonRequest {
```
tedkalaw, line 3: can you doc this? a more descriptive name may also be helpful here (`JsonRequest` is a fairly common action on the interwebz)

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
...
+  method: string
```
tedkalaw, line 4: is this method different from REST methods? ie, is this an `rpcMethod` or some such?

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
...
+  params: unknown
```
tedkalaw, line 5: per your q, this is reasonable to me for now - eventually you could imagine a world where we have types that extend this one. would need to know more about the problem to opine on it though. 

my suspicion is this is either going to be some sort of interesting union type or JsonRequest is going to be a generic

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Submit a JSON request to the rippled server.
+   *
+   * @param jsonRequest A JSON object specifying the method and params of the request.
+   * @see https://xrpl.org/request-formatting.html#json-rpc-format
+   * @returns The response from the rippled server.
+   */
+  public async submitRequest(
+    jsonRequest: JsonRequest,
+  ): Promise<AxiosResponse<unknown>> {
```
tedkalaw, line 36: can we abstract away `AxiosResponse`? this is leaking details about the underlying implementation, as the caller would have to know about `axios` to consume what this function returns

if you wanted to maintain an abstraction but still have flexibility for later, it may be helpful to do

```
type JsonRpcResponse<T> = AxiosResponse<T>
```

and then use `JsonRpcResponse` (or, ideally, a better name probably lol) instead of `AxiosResponse`

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
...
+export default class JsonRpcNetworkClient {
+  private readonly axiosInstance: AxiosInstance
+
+  /**
+   * Create a new JsonRpcNetworkClient.
+   *
+   * @param jsonUrl The URL of the rippled node to query.
+   * @see https://xrpl.org/get-started-with-the-rippled-api.html
+   */
+  public constructor(jsonUrl: string) {
```
tedkalaw, line 21: this constructor seems ok. the only thing that gives me pause here is we lose DI and will need to use `nock` for testing...we may want to revisit this later

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public async submitRequest(
+    jsonRequest: JsonRequest,
+  ): Promise<AxiosResponse<unknown>> {
+    const requestOptions: AxiosRequestConfig = {
+      url: '/',
+      method: 'post',
+      data: jsonRequest,
+      headers: { 'Content-Type': 'application/json' },
+    }
```
tedkalaw, line 42: you can just inline this object imho

mvadari (approved):

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: yep, in subsequent PR I move this to another file with additional JSON interface definitions. doc'ed and renamed here for now

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: yeah, this field is where you specify which rpc command you're invoking.  it needs to be called `method` because that's what rippled is expecting

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: yeah so I kinda realized this blurred boundary and in a subsequent PR, make it such that the `JsonRpcNetworkClient` will actually have a method for each specific RPC, much like the current gRPC-based network client does now (which is also super helpful for DI testing).  Therefore the `AxiosResponse` stays within the network client class and will just returns a JSON object that conforms to some expected type.

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: yeah so related to my response to your comment one above - instead of DI, the unit testing uses fake network clients that can be directly injected into the ledger clients.  So adding methods to this network client for each RPC will make that possible.

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: style points wherever I can get 'em

tedkalaw (approved):
ok

PR #783, opened 2020-10-08
Implements `allowUnauthorizedTrustlines` on `IssuedCurrencyClient`

mvadari:
## High Level Overview of Change

This PR adds the method `allowUnauthorizedTrustlines` to the `IssuedCurrencyClient`, which disables Require Authorization on the specified account. This is achieved by submitting an AccountSet transaction with a clear flag set to 2 (see https://xrpl.org/accountset.html#accountset-flags).

### Context of Change

Part of the IOU effort. 

### Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)

## Before / After

`allowUnauthorizedTrustlines` exposed on `IssuedCurrencyClient`

## Test Plan

CI passes. Wrote unit tests for the function that has full codecov and passes.

PR #787, opened 2020-10-09
Implements `disallowIncomingXrp` on `IssuedCurrencyClient`

mvadari:

This PR adds the method `disallowIncomingXrp` to the `IssuedCurrencyClient`, which enables DisallowXrp on the specified account. This is achieved by submitting an AccountSet transaction with a set flag set to 3 (see https://xrpl.org/accountset.html#accountset-flags).

This PR also refactors the flag functions to avoid repeating so much code, both in the client and in integration testing.

### Context of Change

Part of the IOU effort. 

### Type of Change

## Before / After

`disallowIncomingXrp` exposed on `IssuedCurrencyClient`

## Test Plan

CI passes. Wrote unit tests for the function that has full codecov and passes.

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -148,7 +148,7 @@ export default class IssuedCurrencyClient {
   /**
    * Enable Disallow XRP for this XRPL account.
    *
-   * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#require-auth
+   * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#disallow-xrp
    *
    * @param wallet The wallet associated with the XRPL account enabling Disallow XRP and that will sign the request.
    * @returns A promise which resolves to a TransactionResult object that represents the result of this transaction.
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -131,6 +131,38 @@ export default class IssuedCurrencyClient {
     })
     return trustLines
   }
+
+  /**
+   * Disable Require Authorization for this XRPL account.
+   *
+   * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#require-auth
+   *
+   * @param wallet The wallet associated with the XRPL account disabling Require Authorization and that will sign the request.
+   * @returns A promise which resolves to a TransactionResult object that represents the result of this transaction.
+   */
+  public async allowUnauthorizedTrustlines(
+    wallet: Wallet,
+  ): Promise<TransactionResult> {
+    const clearFlag = new ClearFlag()
+    clearFlag.setValue(AccountSetFlag.asfRequireAuth)
+
+    const accountSet = new AccountSet()
+    accountSet.setClearFlag(clearFlag)
+
+    const transaction = await this.coreXrplClient.prepareBaseTransaction(wallet)
+    transaction.setAccountSet(accountSet)
+
+    const transactionHash = await this.coreXrplClient.signAndSubmitTransaction(
+      transaction,
+      wallet,
+    )
+
+    return await this.coreXrplClient.getFinalTransactionResultAsync(
+      transactionHash,
+      wallet,
+    )
+  }
+
   /**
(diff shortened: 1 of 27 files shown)


mvadari:
Codecov will go back to 100% on the diff once https://github.com/xpring-eng/Xpring-JS/pull/783 is merged to master

src/XRP/issued-currency-client.ts
```diff
@@ -95,7 +102,7 @@ export default class IssuedCurrencyClient {
   }
 
   /**
-   * Retrieves information about an account's trust lines, which maintain balances of all non-XRP currencies and assets.
+   * * Retrieves information about an account's trust lines, which maintain balances of all non-XRP currencies and assets.
```
tedkalaw, line 105: super nit: remove extra `*`

tedkalaw (approved):
nice, seems pretty reasonable to me

src/XRP/issued-currency-client.ts
```diff
@@ -64,21 +64,28 @@ export default class IssuedCurrencyClient {
...
-   * @see https://xrpl.org/become-an-xrp-ledger-gateway.html#require-auth
-   *
+   * Helper function. Sets/clears a flag value.
+   * @param flag The desired flag that is being changed.
+   * @param enable Whether the flag is being enabled (true if enabling, false if disabling).
    * @param wallet The wallet associated with the XRPL account enabling Require Authorization and that will sign the request.
    * @returns A promise which resolves to a TransactionResult object that represents the result of this transaction.
    */
-  public async requireAuthorizedTrustlines(
+  private async changeFlag(
```
amiecorso, line 73: nice

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -26,6 +26,45 @@ const issuedCurrencyClient = IssuedCurrencyClient.issuedCurrencyClientWithEndpoi
   XrplNetwork.Test,
 )
 
+async function runFlagIntegrationTestSuccess(
```
amiecorso, line 29: also nice.  can we move this guy to `test/XRP/helpers/xrp-test-utils.ts` ?

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -94,38 +133,24 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+    // GIVEN an existing testnet account
+    // WHEN disallowIncomingXrp is called
+    const result = await issuedCurrencyClient.disallowIncomingXrp(wallet)
+
+    // THEN the transaction was successfully submitted and the correct flag was set on the account.
+    await runFlagIntegrationTestSuccess(
+      wallet,
+      result,
+      AccountRootFlag.LSF_DISALLOW_XRP,
     )
```
amiecorso, line 154: How do you feel about creating another integration test that attempts to send an XRP transaction to this account that has set `disllowIncomingXrp`?  We'd basically want to just create a second faucet wallet and then try to use the `send` method on the `XrpClient` to send an XRP transaction to this account.  Ideally the transaction result will fail (I'm also just curious to understand what happens in this case and check that the SDK handles it correctly and returns an informative transaction status).  Note that the `send` method only returns a transaction hash, and then you have to follow up with `getPaymentStatus` to see the full `TransactionResult` object (I think).  We want to refactor that (so that `send` returns a full `TransactionResult`) cause it's old trash but requires a deprecation so... we're getting around to it.

mvadari: I tried writing that, but the transaction succeeds for some reason. I don't have a good enough understanding of the system to understand why.

amiecorso: Note for posterity: As we just discovered together, the docs here https://xrpl.org/accountset.html#accountset-flags indicate that this flag isn't actually enforced by rippled, but rather is an indicator to client applications.  Therefore we can't implement this type of integration test.

amiecorso (approved):
v nice

tedkalaw (approved):
homer-bush

PR #806, opened 2020-10-14
Creates `IssuedCurrencyClient::getGatewayBalances`

amiecorso:
This PR implements the `gateway_balances` JSON RPC as the `getGatewayBalances` API in the SDK `IssuedCurrencyClient`.
The returned object is a `GatewayBalances`... a natively implemented typescript class that mimics the JSON response from the underlying API.  Open to better suggestions around how to present this information.

### Context of Change
IOU effort

### Type of Change

## Before / After
IssuedCurrencyClient now has `getGatewayBalances` API

## Test Plan
New unit and integration tests written, CI

## Future Tasks

Expose publicly


src/XRP/shared/gateway-balances.ts
```diff
@@ -0,0 +1,80 @@
...
+import { XrpError, XrpErrorType } from '.'
+import {
+  GatewayBalancesResponse,
+  CurrencyValuePair,
+} from './rippled-json-rpc-schema'
+
+/**
+ * Represents an amount of an Issued Currency.
+ */
+export class IssuedCurrencyValue {
```
mvadari, line 11: Don't you want the issuer to be stored here as well?

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -81,4 +83,34 @@ export default class JsonRpcNetworkClient {
...
+  public async getGatewayBalances(
+    account: string,
+    hotwallet?: string | Array<string>,
+  ): Promise<GatewayBalancesResponse> {
+    const gatewayBalancesRequest = {
+      method: 'gateway_balances',
+      params: [
+        {
+          account: account,
+          hotwallet: hotwallet,
```
mvadari, line 104: I believe this can also be written as just `hotwallet,`

mvadari (commented):
A few minor comments, I also think you need a little better coverage on the diff

src/XRP/issued-currency-client.ts
```diff
@@ -94,6 +106,67 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @param account The account for which to retrieve balance information, encoded as an X-Address.
+   * @param hotwallet (Optional) An operational address to exclude from the balances issued, or an array of such addresses,
+   *                   encoded as X-Addresses.
+   * @see https://xrpaddress.info/
+   * @returns A GatewayBalances object containing information about an account's balances.
+   */
+  public async getGatewayBalances(
+    account: string,
+    hotwallet?: string | Array<string>,
```
tedkalaw, line 122: quick q - why an array? is this a common pattern that we expect? 

`hotwallet` is a SUPER opinionated term, which isn't bad, but...will need to think more about this. could this be `exclude`?

src/XRP/issued-currency-client.ts
```diff
@@ -94,6 +106,67 @@ export default class IssuedCurrencyClient {
...
+    // TODO: verify address format of hotwallet (and add test)
+    const gatewayBalancesResponse: GatewayBalancesResponse = await this.jsonNetworkClient.getGatewayBalances(
+      classicAddress.address,
+      hotwalletArray,
+    )
+
+    if (gatewayBalancesResponse.result.error) {
+      if (gatewayBalancesResponse.result.error == 'actNotFound') {
+        throw XrpError.accountNotFound
+      } else if (gatewayBalancesResponse.result.error == 'invalidHotwallet') {
```
tedkalaw, line 158: might be worth making an enum for these. is there a well-defined set of errors for gateway balances responses, or is it just a coincidence that `actNotFound` appears both here and in `getTrustLines`?

src/XRP/issued-currency-client.ts
```diff
@@ -94,6 +106,67 @@ export default class IssuedCurrencyClient {
...
+        }
+      }
+    }
+    // TODO: verify address format of hotwallet (and add test)
+    const gatewayBalancesResponse: GatewayBalancesResponse = await this.jsonNetworkClient.getGatewayBalances(
+      classicAddress.address,
+      hotwalletArray,
+    )
+
+    if (gatewayBalancesResponse.result.error) {
```
tedkalaw, line 155: i think this would be easier to reason about if you inverted it and returned early, ie

```
if (!gatewayBalancesResponse.result.error) {
  return new GatewayBalances(gatewayBalancesResponse)
}
```

src/XRP/issued-currency-client.ts
tedkalaw: if you returned early above, you could convert this into a switch and do

```
switch (gatewayBalancesResponse.result.error) {
  case 'actNotFound':
    throw the error
  case 'invalidHotWallet'
    throw the error
  default:
    throw unknown error
}
```

src/XRP/network-clients/json-rpc-network-client.ts
tedkalaw: yee, and `account` as `account`

https://alligator.io/js/object-property-shorthand-es6/

src/XRP/shared/gateway-balances.ts
```diff
@@ -0,0 +1,80 @@
...
+  readonly currency: string
+
+  readonly value: string
+
+  public constructor(currencyValuePair: CurrencyValuePair) {
+    this.currency = currencyValuePair.currency
+    this.value = currencyValuePair.value
+  }
+}
```
tedkalaw, line 20: do you expect this class to ever be more than a bag of values?

if not, imo you should just do

```
interface IssuedCurrencyValue {
  readonly currency: string
  readonly value: string
}

export function issuedCurrencyValueFromPair({currency, value}) {
  return {
    currency,
    value
  }
}
```

src/XRP/shared/gateway-balances.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Represents information about a gateway's issued currency balances on the XRPL, as of the most recent validated ledger version.
+ * @see https://xrpl.org/gateway_balances.html
+ */
+export default class GatewayBalances {
```
tedkalaw, line 26: same q for this guy - if you don't expect it to ever "do" anything, i think it's better to use an `interface`

since TS is a structural type system, creating a `class` implies that you have aspirations for the object to "do" stuff (a la OOP). if you instead make it an `interface` and construct it using a function (like `createGatewayBalance` or `initGatewayBalance` or whatever), it's super clear that this thing, particularly with readonly properties, represents a collection of data but isn't an actor itself

tedkalaw (commented):
good stuff with the fakes. some thoughts inline

src/XRP/shared/gateway-balances.ts
amiecorso: Nope... this type appears in the `gateway_balances` response where the issuer is always understood to be the account provided in the `gateway_balances` query, so it's different from the protobuf that represents issued currencies.

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: someday this'll sink in

src/XRP/issued-currency-client.ts
```diff
@@ -81,8 +85,16 @@ export default class IssuedCurrencyClient {
     )
 
     if (accountLinesResponse.result.error) {
-      throw XrpError.accountNotFound
+      if (accountLinesResponse.result.error === 'actNotFound') {
```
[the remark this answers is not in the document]
amiecorso: good idea - it's actually used in two places, and there's also the `invalidHotwallet` that you note below.  I made a class for these with static attributes to represent the error messages.  To answer your question below - these aren't exactly standard... there are a few error messages that apparently are standard, and then a list of what the various messages might be documented for each API, and it seems there's a lot of consistency.  I think it's reasonable to just add these as constants on a case-by-case basis... this might be the last time we implement something like this for JSON RPC at least, but we might have to revisit this if we do a big refactor off of gRPC

src/XRP/issued-currency-client.ts
amiecorso: The array is simply how the JSON RPC handles it, because there may be several addresses that you want to exclude from your notion of balances.  I think you're right though that hotwallet is very opinionated and not even that clear.  I went ahead and renamed this to `addressesToExclude` - a bit more cumbersome but certainly more clear.

src/XRP/issued-currency-client.ts
amiecorso: went for the switch case

src/XRP/issued-currency-client.ts
amiecorso: agreed, swapped

src/XRP/shared/gateway-balances.ts
amiecorso: sweet yeah way better, and I didn't even need the constructor because this is just adding typing to JSON objects that already exist - no transformations even happening

src/XRP/shared/gateway-balances.ts
amiecorso: dope went interfacey here too

src/XRP/issued-currency-client.ts
```diff
@@ -126,6 +136,71 @@ export default class IssuedCurrencyClient {
...
+      } else {
+        for (const address of accountsToExclude) {
+          const excludeClassicAddress = XrpUtils.decodeXAddress(address)
+          if (!excludeClassicAddress) {
+            throw XrpError.xAddressRequired
+          }
+          addressesToExcludeArray.push(excludeClassicAddress.address)
+        }
+      }
+    }
```
tedkalaw, line 178: if `accountsToExclude` is always an array, then this can all turn into

```
accountsToExclude.forEach((account) => {
  const excludeClassicAddress = XrpUtils.decodeXAddress(accountToExclude)
  if (!excludeClassicAddress) {
    throw XrpError.xAddressRequired
  }
}
```

src/XRP/network-clients/json-rpc-network-client.ts
```diff
@@ -85,4 +87,34 @@ export default class JsonRpcNetworkClient
...
+   * Submits a gateway_balances request to the rippled JSON RPC.
+   * @see https://xrpl.org/gateway_balances.html
+   *
+   * @param account The XRPL account for which to retrieve balances.
+   * @param addressesToExclude (Optional) An operational address to exclude from the balances issued, or an array of such addresses.
+   * @see https://xrpl.org/issuing-and-operational-addresses.html
+   */
+  public async getGatewayBalances(
+    account: string,
+    addressesToExclude?: string | Array<string>,
```
tedkalaw, line 101: as above, can this just always be an array? it seems like it would simplify a lot of stuff

src/XRP/shared/gateway-balances.ts
```diff
@@ -0,0 +1,76 @@
...
+  readonly obligations?: { [currencyCode: string]: string }
+
+  /** (May be omitted) The ledger index of the ledger version that was used to generate this response. */
+  readonly ledgerHash?: string | undefined
+}
+
+export function gatewayBalancesFromResponse(
+  gatewayBalancesResponse: GatewayBalancesResponse,
+): GatewayBalances {
+  if (gatewayBalancesResponse.result.validated === false) {
```
tedkalaw, line 48: it's pretty idiomatic to just do 
```
if (!gatewayBalancesResponse.result.validated)
```

for bools

tedkalaw, line 48: we don't actually seem to care about the `gatewayBalancesResponse` object once we get into this function - what if we just did

`const {result} = gatewayBalancesResponse`

? then you don't have to keep typing `gatewayBalancesResponse` every time

src/XRP/shared/rippled-error-messages.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * A class for creating constants that represent possible error messages received from the rippled JSON RPC.
+ */
+export default class RippledErrorMessages {
+  static accountNotFound = 'actNotFound'
+  static invalidExcludedAddress = 'invalidHotwallet'
```
tedkalaw, line 6: this seems ok but i think it would be fine also to have an enum for these

tedkalaw, line 6: is this right? i would think it should be `invalidHotWallet` but i never know the casing with rippled

src/XRP/shared/rippled-json-rpc-schema.ts
```diff
@@ -18,19 +18,24 @@ interface AccountLinesResponse {
...
+  result: {
+    error?: string
+    account?: string
+    assets?: { [account: string]: CurrencyValuePair[] }
+    balances?: { [account: string]: CurrencyValuePair[] }
+    ledger_hash?: string
+    ledger_index?: number
+    obligations?: { [currencyCode: string]: string }
+    status: string
+    validated?: boolean
```
tedkalaw, line 36: lol this is an intense type. is there some mutual exclusion between these types? ie, is it really the case that there can be between 1 and 9 different keys? or will some of these appear when a request is successful and not show up when a request is not successful?

test/XRP/protos-models-flags-utils/gateway-balances.test.ts
```diff
@@ -0,0 +1,257 @@
...
+        assets: {
+          r9F6wk8HkXrgYWoJ7fsv4VrUBVoqDVtzkH: [
+            {
+              currency: 'BTC',
+              value: '5444166510000000e-26',
+            },
+            {
+              currency: 'EUR',
+              value: '7000000000000000e-26',
+            },
```
tedkalaw, line 25: this fake looks pretty similar to the one in the next test - can we make a base object that has the required fields and compose more complicated fakes? it's hard to tell at a glance what's missing and what's not

src/XRP/shared/gateway-balances.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  readonly balances?: { [account: string]: IssuedCurrencyValue[] }
+
+  /**
+   * (Optional, omitted if empty) Total amounts issued to addresses not excluded, as a map of currencies to the total value issued.
+   */
+  readonly obligations?: { [currencyCode: string]: string }
+
+  /** (May be omitted) The ledger index of the ledger version that was used to generate this response. */
+  readonly ledgerHash?: string | undefined
```
mvadari, line 42: This doesn't need `undefined` since it has the `?`

src/XRP/issued-currency-client.ts
```diff
@@ -126,6 +136,71 @@ export default class IssuedCurrencyClient {
...
+    const gatewayBalancesResponse: GatewayBalancesResponse = await this.jsonNetworkClient.getGatewayBalances(
+      classicAddress.address,
+      addressesToExcludeArray,
+    )
+
+    if (!gatewayBalancesResponse.result.error) {
+      return gatewayBalancesFromResponse(gatewayBalancesResponse)
+    } else {
```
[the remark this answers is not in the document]
amiecorso: oh right

src/XRP/issued-currency-client.ts
```diff
@@ -126,6 +136,71 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @param account The account for which to retrieve balance information, encoded as an X-Address.
+   * @param accountsToExclude (Optional) An operational address to exclude from the balances issued, or an array of such addresses,
+   *                   encoded as X-Addresses.
+   * @see https://xrpaddress.info/
+   * @returns A GatewayBalances object containing information about an account's balances.
+   */
+  public async getGatewayBalances(
+    account: string,
+    accountsToExclude?: string | Array<string>,
```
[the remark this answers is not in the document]
amiecorso: i agree

src/XRP/issued-currency-client.ts
amiecorso: So I can get rid of the check for that ugly check for a string type, but in this case I actually still have to create the array of classic addresses because this is going directly into a JSON request (as opposed to being serialized and signed) so it can't actually take X-addresses.  I'll name it a bit better

src/XRP/network-clients/json-rpc-network-client.ts
amiecorso: yar

src/XRP/shared/gateway-balances.ts
amiecorso: duh i'm such a newb

amiecorso: although I guess my subconscious rationale wanted to make it more clear that this IS a bool since we're always checking all sorts of truthy and falsey things in JS, but I will definitely change this

amiecorso: yes

src/XRP/shared/rippled-error-messages.ts
amiecorso: ah gotcha didn't think typescript supported string literal enums but it does

amiecorso: nice catch, it is indeed that

src/XRP/shared/gateway-balances.ts
amiecorso: oh right, ty

test/XRP/protos-models-flags-utils/gateway-balances.test.ts
amiecorso: that's a great idea

amiecorso:
lol somehow my poor diff coverage magically resolved itself when I responded to all of these review comments...

src/XRP/issued-currency-client.ts
```diff
@@ -126,6 +132,58 @@ export default class IssuedCurrencyClient {
...
+
+    // check excludable addresses for X-Address format
+    const classicAddressesToExclude: Array<string> = []
+    accountsToExclude.map((xAddress) => {
+      const excludeClassicAddress = XrpUtils.decodeXAddress(xAddress)
+      if (!excludeClassicAddress) {
+        throw XrpError.xAddressRequired
+      }
+      classicAddressesToExclude.push(excludeClassicAddress.address)
+    })
```
tedkalaw, line 163: you can just do

```
const classicAddressesToExclude = accountsToExclude.map(() => {
  // code code code
})
```

`map` returns an `Array` already

src/XRP/shared/rippled-json-rpc-schema.ts
tedkalaw: got it, how about

```
BaseGatewayBalancesResult {
  status: string
}

interface GatewayBalancesErrorResult extends BaseGatewayBalancesResult {
  error: string
}

interface GatewayBalancesSuccessResult extends BaseGatewayBalancesResult {
  account: string
  // a whole lotta optional fields
}

interface GatewayBalancesResponse {
  result: GatewayBalancesSuccessResult | GatewayBalancesErrorResult
}
```

interestingly, this will make it so that typescript is able to differentiate between a `GatewayBalancesErrorResult` and a `GatewayBalancesSuccessResult` by checking if `error` is there. nice.

test/XRP/issued-currency-client.test.ts
```diff
@@ -369,6 +372,112 @@ describe('Issued Currency Client', function (): void {
...
+      fakeSucceedingGrpcClient,
+      fakeErroringJsonClient,
+      XrplNetwork.Test,
+    )
+    // WHEN getGatewayBalances is called THEN an error is propagated.
+    issuedCurrencyClient.getGatewayBalances(testAddress).catch((error) => {
+      assert.typeOf(error, 'Error')
+      assert.equal(error, XrpError.accountNotFound)
+      done()
+    })
```
tedkalaw, line 478: can we try/catch this?

test/XRP/issued-currency-client.test.ts
```diff
@@ -369,6 +372,112 @@ describe('Issued Currency Client', function (): void {
...
+    // WHEN getGatewayBalances is called with classic addresses to exclude (no X-address) THEN an error is propagated.
+    const classicAddress1 = 'rhhh49pFH96roGyuC4E5P4CHaNjS1k8gzM'
+    const classicAddress2 = 'r4DymtkgUAh2wqRxVfdd3Xtswzim6eC6c5'
+    issuedCurrencyClient
+      .getGatewayBalances(testAddress, [classicAddress1, classicAddress2])
+      .catch((error) => {
+        assert.typeOf(error, 'Error')
+        assert.equal(error, XrpError.xAddressRequired)
+        done()
+      })
```
tedkalaw, line 450: same hurr. there are others in this test but didn't wanna spam you

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -240,6 +238,44 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+    this.timeout(timeoutMs)
+    // GIVEN a valid address that doesn't actually exist on the ledger
+    const walletFactory = new WalletFactory(XrplNetwork.Test)
+    const address = (await walletFactory.generateRandomWallet())!.wallet.getAddress()
+
+    // WHEN getGatewayBalances is called for that address THEN an error is propagated.
+    issuedCurrencyClient.getGatewayBalances(address).catch((error) => {
+      assert.typeOf(error, 'Error')
+      assert.equal(error, XrpError.accountNotFound)
+    })
```
tedkalaw, line 257: rtry/catch? does `assertThrows` work here too? i forgot that was a thing and saw it in another test in this file

test/XRP/fakes/fake-json-network-client.ts
```diff
@@ -34,6 +37,9 @@ export class FakeJsonNetworkClientResponses {
     public readonly getAccountLinesResponse: Result<
       AccountLinesResponse
     > = FakeJsonNetworkClientResponses.defaultGetAccountLinesResponse(),
+    public readonly getGatewayBalancesReponse: Result<
```
tedkalaw, line 40: `getGatewayBalancesResponse`

tedkalaw (approved):
seems cool, left a couple thoughts inline

src/XRP/issued-currency-client.ts
amiecorso: thanks for teaching me javascript ted

test/XRP/issued-currency-client.test.ts
amiecorso: yerp, applied to all

test/XRP/issued-currency-client-integration.test.ts
amiecorso: I think assertThrows would also work but then doesn't give us the granularity of checking that the error types are the same?

test/XRP/fakes/fake-json-network-client.ts
amiecorso: excellent catch

amiecorso:
thanks for the approval - will wait to merge until we can circle back on the typing headache

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -240,6 +238,44 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+  // which will also avoid the need for maintenance after a testnet reset.
+
+  // TODO: Implement an integration test for an account with balances/assets/obligations once functionality exists for first creating things.
+  it('getGatewayBalances - account not found', async function (): Promise<
+    void
+  > {
+    this.timeout(timeoutMs)
+    // GIVEN a valid address that doesn't actually exist on the ledger
+    const walletFactory = new WalletFactory(XrplNetwork.Test)
+    const address = (await walletFactory.generateRandomWallet())!.wallet.getAddress()
```
mvadari, line 251: Is this necessary when there's a wallet generated in `beforeEach`?

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -240,6 +238,44 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+      assert.equal(error, XrpError.accountNotFound)
+    })
+  })
+
+  it('getGatewayBalances - account with no issued currencies', async function (): Promise<
+    void
+  > {
+    // GIVEN a valid, funded address that has not issued any currencies
+    const wallet = await XRPTestUtils.randomWalletFromFaucet()
+    const address = wallet.getAddress()
```
mvadari, line 265: same as above

mvadari (approved):

test/XRP/issued-currency-client-integration.test.ts
amiecorso: so this one actually is because I want a valid address that is NOT a faucet wallet, (because a faucet wallet actually does exist on the ledger -- it has been funded).  Luckily, this makes no network calls and is fast.

test/XRP/issued-currency-client-integration.test.ts
amiecorso: The reason I did this here is because I imagine other integration tests *might* create trustlines using the wallet generated in the `before` (which is what we're working with rn, as opposed to a `beforeEach`) and I didn't want the *order* of tests to matter.  this will be re-evaluated in the context of all the integration tests when we take care of the outstanding Asana task to do a deep dive.

PR #814, opened 2020-10-16
Creates `XrpClient::sendXrp` (which returns a `TransactionResult`) and Deprecates `XrpClient::send`

amiecorso:
As best described [here](https://app.asana.com/0/1188384810103401/1198302721476927/f), the original `send` method on the `XrpClient` only returns a transaction hash.  The hash is returned when the outcome of the transaction is final (i.e. either the transaction was included in a validated ledger, was malformed to begin with, or has not yet been included AND its `lastLedgerSequence` field expired (surpassed by most recent validated ledger index)) - however, simply returning a hash makes it unclear which of these outcomes occurred.  Originally, it was intended that a user use `getPaymentStatus` on the resultant transaction hash to query the outcome of the transaction, however even this only returned either `TransactionStatus.Pending` (<-- an inaccurate and misleading status) or `TransactionStatus.Succeeded` / `TransactionStatus.Failed`.  It is unclear whether the transaction is on a validated ledger, and unclear why or why not.

Other transaction-sending methods use a `TransactionResult` object, which returns information about the transaction's status, inclusion or exclusion from a validated ledger, and whether or not this result is final.

This PR creates a new method: `XrpClient::sendXrp` (identical to `XrpClient::send` apart from the return value, which is a `TransactionResult` and it deprecates `XrpClient::send`, which will be removed from the SDK after two release cycles.

The renamed method is also important now that IOU work is poised to establish a `sendIssuance` or something along those lines (send no longer refers only to XRP).

### Type of Change

## Before / After
`sendXrp` exists and returns an informative `TransactionResult`.  `send` is deprecated.

## Test Plan
Existing tests have been refactored to test the new method.  At the top level, `send` now simply wraps a call to `sendXrp`.

### DIFF
--- a/src/XRP/default-xrp-client.ts
+++ b/src/XRP/default-xrp-client.ts
@@ -134,17 +134,17 @@ export default class DefaultXrpClient implements XrpClientDecorator {
   /**
    * Send the given amount of XRP from the source wallet to the destination address.
    *
-   * @param amount A `BigInteger`, number or numeric string representing the number of drops to send.
+   * @param amount A `BigInteger`, number or numeric string representing the number of drops of XRP to send.
    * @param destinationAddress A destination address to send the drops to.
    * @param sender The wallet that XRP will be sent from and which will sign the request.
-   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
+   * @returns A promise which resolves to a TransactionResult representing the pending outcome of the submitted transaction.
    */
-  public async send(
+  public async sendXrp(
     amount: BigInteger | number | string,
     destinationAddress: string,
     sender: Wallet,
-  ): Promise<string> {
-    return this.sendWithDetails({
+  ): Promise<TransactionResult> {
+    return await this.sendXrpWithDetails({
       amount,
       destination: destinationAddress,
       sender,
@@ -156,11 +156,11 @@ export default class DefaultXrpClient implements XrpClientDecorator {
    * for additional details to be specified for use with supplementary features of the XRP ledger.
    *
    * @param sendXrpDetails - a wrapper object containing details for constructing a transaction.
-   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
+   * @returns A promise which resolves to a TransactionResult representing the pending outcome of the submitted transaction.
    */
-  public async sendWithDetails(
+  public async sendXrpWithDetails(
     sendXrpDetails: SendXrpDetails,
-  ): Promise<string> {
+  ): Promise<TransactionResult> {
     const {
       amount: drops,
       sender,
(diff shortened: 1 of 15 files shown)


src/XRP/default-xrp-client.ts
```diff
@@ -156,11 +156,11 @@ export default class DefaultXrpClient implements XrpClientDecorator {
...
    *
    * @param sendXrpDetails - a wrapper object containing details for constructing a transaction.
-   * @returns A promise which resolves to a string representing the hash of the submitted transaction.
+   * @returns A promise which resolves to a TransactionResult representing the pending outcome of the submitted transaction.
    */
-  public async sendWithDetails(
+  public async sendXrpWithDetails(
     sendXrpDetails: SendXrpDetails,
-  ): Promise<string> {
+  ): Promise<TransactionResult> {
```
amiecorso, line 163: So here in `DefaultXrpClient` I had a decision to make... Currently, `DefaultXrpClient` implements `XrpClientDecorator`, which means that the return value of this method has to be the same as the return value in `ReliableSubmissionXrpClient` AKA a `TransactionResult`.  However, at this point in the process, all we actually need is the transaction hash... the other result values like the status and validated/unvalidated don't matter yet because it's for sure pending...  So what I COULD do is restore this to the original string return and then have `DefaultXrpClient` not implement `XrpClientDecorator` (because it's not actually a decorator... it doesn't decorate anything).  Or I can just leave it like this, I'm not sure it really matters.

mvadari (approved):

src/XRP/default-xrp-client.ts
tedkalaw: i think it's fine to use `TransacitonResult`. i like the more consistent interface...when i tried to use send it felt strange to me that it returned a string instead of `TransactionResult`

src/XRP/default-xrp-client.ts
```diff
@@ -219,7 +219,15 @@ export default class DefaultXrpClient implements XrpClientDecorator {
         .forEach((memo) => transaction.addMemos(memo))
     }
 
-    return this.coreXrplClient.signAndSubmitTransaction(transaction, sender)
+    const transactionHash = await this.coreXrplClient.signAndSubmitTransaction(
+      transaction,
+      sender,
+    )
+    return TransactionResult.getPendingTransactionResult(
```
tedkalaw, line 226: this is a random thought, but now that i'm seeing this, i think this function should be called `createPendingTransactionResult` - i thought get was doing some computation or something

tedkalaw (approved):
cool, this is great! thanks for doing this, i found it super confusing that `send` and `sendWithDetails` returned a string instead of a `transactionResult` lol

src/XRP/default-xrp-client.ts
amiecorso: I totally agree with that - I renamed this and also its counterpart `getFinalTransactionResult` to start with `create`

PR #831, opened 2020-10-20
Adds `peerAccount` Optional Param to `IssuedCurrencyClient::getTrustLines`

amiecorso:
The [`account_lines`](https://xrpl.org/account_lines.html) RPC allows for an optional `peer` field in the request, which, if provided, limits the results returned only to trust lines established between the querying account and the provided peer account.  This is a useful feature that is best to implement at the network layer in order to avoid an excessive amount of data between unnecessarily returned over the network.  

This PR implements that option in the `IssuedCurrencyClient`'s `getTrustLines` method.

### Context of Change
IOU effort

### Type of Change

## Before / After
Optional `peer` option available as `peerAccount` param.

## Test Plan
Added some unit tests to check that our method correctly validates the format of the peer address. 
Integration tests will be easier to write when we have more complete IOU functionality.

## Future Tasks

Add integration tests that specifically test for the filtering effect of the peer option.  Let's do this once we're able to establish and authorize trustlines using the `IssuedCurrencyClient`, so that we can control what we're evaluating.

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -8,7 +8,7 @@ import TrustLine from './shared/trustline'
 import JsonRpcNetworkClient from './network-clients/json-rpc-network-client'
 import { AccountLinesResponse } from './shared/rippled-json-rpc-schema'
 import { JsonNetworkClientInterface } from './network-clients/json-network-client-interface'
-import { XrpError } from './shared'
+import { XrpError, XrpErrorType } from './shared'
 import { AccountSetFlag } from './shared/account-set-flag'
 import TransactionResult from './shared/transaction-result'
 import { TransferRate } from './Generated/node/org/xrpl/rpc/v1/common_pb'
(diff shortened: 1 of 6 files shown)


test/XRP/issued-currency-client.test.ts
```diff
@@ -71,6 +71,25 @@ describe('Issued Currency Client', function (): void {
...
+
+    // WHEN getTrustLines is called with an invalid peerAccount address THEN an error is propagated.
+    const peerAddress = 'malformedAddress'
+    issuedCurrencyClient
+      .getTrustLines(testAddress, peerAddress)
+      .catch((error) => {
+        assert.typeOf(error, 'Error')
+        assert.equal(error, XrpError.xAddressRequired)
+        done()
+      })
```
tedkalaw, line 90: can we just await `issuedCurrencyClient.getTrustLines` and use a `try/catch` instead of a promise? you don't need the `done` callback anymore bc it acts synchronously

i noticed today that we do this in a lot of places in the sdk but i think it's confusing...it's the same thing as doing

```
try {
  await myFunction()
} catch (error) {
  errorHandler()
}
done()
```

note that you need to return the `done` callback from the test call or else it will yell at you

test/XRP/issued-currency-client.test.ts
```diff
@@ -98,6 +117,39 @@ describe('Issued Currency Client', function (): void {
...
+      assert.typeOf(error, 'Error')
+      // TODO: why doesn't this work?
+      // I get: AssertionError: expected [Error: invalidParams] to deeply equal [Error: invalidParams]
+      // ( and same when I use assert.equal instead of assert.deepEqual)
+      // assert.deepEqual(
+      //   error,
+      //   new XrpError(XrpErrorType.Unknown, 'invalidParams'),
+      // )
+      done()
+    })
```
tedkalaw, line 150: same here

src/XRP/issued-currency-client.ts
```diff
@@ -68,21 +68,41 @@ export default class IssuedCurrencyClient {
...
    * @param account The account for which to retrieve associated trust lines, encoded as an X-Address.
+   * @param peerAccount (Optional) The address of a second account. If provided, show only trust lines connecting the two accounts.
    * @see https://xrpaddress.info/
    * @returns An array of TrustLine objects, representing all trust lines associated with this account.
    */
-  public async getTrustLines(account: string): Promise<Array<TrustLine>> {
+  public async getTrustLines(
+    account: string,
+    peerAccount?: string,
+  ): Promise<Array<TrustLine>> {
```
tedkalaw, line 78: this seems reasonable. have you considered making a separate function for just a peer (ie, `getTrustLineWithPeer`)? 

i think i'm ok with this api but just wanted to throw that out there

tedkalaw (approved):

mvadari (approved):
Is `JsonRpcNetworkClient` supposed to implement `JsonNetworkClientInterface`?

src/XRP/issued-currency-client.ts
amiecorso: yeah I think i'm ok with this for now too.

test/XRP/issued-currency-client.test.ts
amiecorso: nice, yeah this must be a relic of me copying something Keefer once did and then allowing that format to propagate.  Fixed the format for all `getTrustLines` tests.

amiecorso:
> Is `JsonRpcNetworkClient` supposed to implement `JsonNetworkClientInterface`?

Mmm, yes it probably should huh?  especially because the `JsonNetworkClient` is typed as the interface... nice catch.

PR #832, opened 2020-10-21
Creates `issuedCurrencyPayment` method on `IssuedCurrencyClient`

amiecorso:
This PR creates the `issuedCurrencyPayment` method on the `IssuedCurrencyClient`.
`issuedCurrencyPayment` covers creating, sending, and redeeming issued currencies (all of which are the same transaction type) but does not (yet) cover cross-currency payments (sending issued currency -- or XRP -- such that the recipient actually ends up receiving a different currency) or currency exchanges (a cross-currency payment made to oneself).

This method will become the private workhorse behind more opinionated, public APIs, that all share the same underlying transaction type:
- `createIssuedCurrency`
- `sendIssuedCurrencyPayment`
- `redeemIssuedCurrency`
- `sendCrossCurrencyPayment`
- `exchangeCurrency`

Which of the above functionality is achieved by the issued currency payment depends on the value of parameters:
(note cross-currency payments and currency exchange not yet described here)

Case - ALL:
-  The destination address is NOT the same as the sending address of the transaction.
-  (This should only be true for currency exchange payments, which is a cross-currency payment to yourself.)
-   The SendMax field (if present) should always indicate the same currency and issuer as the Amount field... otherwise, it would be considered a cross-currency payment (which are triggered by different currencies in Amount v.s. SendMax fields).
- SendMax values are not parameterized here, and are instead derived from the provided values for issuer, currency, etc., therefore this method is incapable of initiating a cross-currency payment.

Case - Creating an issued currency:
-  The sending address of the transaction is the same address as the issuer of the `Amount` field.
-   No SendMax field is required because the Transfer Fee doesn't apply to transactions in which the originator or the destination address is the same as the issuer. (you don't get charged to handle your own money, or let your customers redeem their money with you.)

Case - dispensing an issued currency from an operational address (https://xrpl.org/issuing-and-operational-addresses.html):
-   The sending address (an operational address) is not the same address as the issuer of the `Amount` field
-   Issuer needs to allow rippling (https://xrpl.org/rippling.html).
-  A SendMax field is required here, because the transfer fee will apply as the sender of this txn is not the issuer.

Case - sending an issued currency payment from one XRPL address to another:
-  Identical to dispensing an issued currency from an operational address (which is just making an issued currency payment)

Case - redeeming an issued currency with the issuer:
-  The destination address is the same address as the issuer of the `Amount` field.
-  No SendMax field is required because the Transfer Fee doesn't apply (see first Case above)

### Context of Change
IOU effort

### Type of Change

## Before / After
`IssuedCurrencyClient` can issue, move and redeem issued currencies.

## Test Plan
Extensive integration tests added, and some unit tests for checking address formats.

## Future Tasks

Consider restructuring this logic by splitting into separate methods for each of issue/move/redeem.  Consider parameterizing this method further such that it becomes capable of cross-currency payments, and then create facades for the distinct cases by wrapping this core method.


test/XRP/issued-currency-client.test.ts
```diff
@@ -583,4 +584,77 @@ describe('Issued Currency Client', function (): void {
...
+        testAddress,
+        'FOO',
+        testClassicAddress,
+        '100',
+        0.5,
+      )
+      .catch((error) => {
+        assert.equal(error.errorType, XrpErrorType.XAddressRequired)
+      })
+  })
```
mvadari, line 659: Maybe add a failing case here

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+   *
+   *      A SendMax field is required here, because the transfer fee will apply as the sender of this txn is not the issuer.
+   *
+   *
+   * Case - sending an issued currency payment from one XRPL address to another:
+   *      Identical to dispensing an issued currency from an operational address (which is just making an issued currency payment)
+   *
+   *
+   * Case - redeeming an issued currency with the issuer:
+   *      The destination address is the same address as the issuer of the `Amount` field.
```
tedkalaw, line 431: > The specific case depends on the relationship among the parameters.

this is pretty intense

the downside with this is you have to rely on documentation and your knowledge of how the parameters interact in order to get things done, as opposed to a more opinionated api.

i can't tell yet if we actually want more abstract apis for each of these functions (i feel like if i were building a DEX or gateway that used this stuff, i would want them), but either way we would probably want a lower-level escape hatch so it seems reasonable for this function to be public.

if you *really* wanted the mega function, one way you could enforce some exclusivity with the type system would be to do something like

```
interface CaseOne {
  destination: string,
  transferFee: number,
}

interface CaseTwo {
  destination: string,
  mutuallyExclusiveField: number
}

type SendIssuedCurrencyOptions = CaseOne | CaseTwo

function sendIssuedCurrency(options: SendIssuedCurrencyOptions) ...
```

not sure if there's a sufficient number of combinations for this to be useful, though.

in the future, it might be good to make PRs that focus on the individual cases instead of doing it all at once. there might be some overlap in what earlier PRs implement, but it's easier to follow

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+   *                    The transfer fee associated with the issuing account, expressed as a percentage. (i.e. a value of .5 indicates a 0.5% transfer fee).
+   */
+  // TODO: (acorso) make this private if/when incorporated into higher level convenience methods
+  // TODO: (acorso) consider using an object for potentially long list of params
+  public async sendIssuedCurrency(
+    sender: Wallet,
+    destinationAddress: string,
+    currency: string,
+    issuer: string,
+    value: string,
```
tedkalaw, line 451: `amount`? 

also i noticed `send` is a `BigInteger | number | string` - do you expect us to offer an interface like that here as well?

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+    payment.setAmount(amount)
+
+    if (!(transferFee === undefined)) {
+      // Construct SendMax - value should be intended amount + relevant transfer fee
+      // Note that a transfer fee doesn't apply if the source address (of this transaction) and the issuing address of the currency being
+      // sent are the same (i.e. issuer sending currency directly) OR if the issuer of the currency and the destination are the same
+      // (i.e. redeeming an issued currency).
+      // However, it also doesn't hurt to include a SendMax field, it just won't nd up being used (added this to questions just to make sure).
+
+      const numericValue = Number(value)
```
tedkalaw, line 513: what if the number is a big integer? is it not possible to send big integer amounts of issued currencies?

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+      const sendMaxValue = Math.ceil((1 + transferFee / 100) * numericValue)
```
tedkalaw, line 514: can you leave a comment explaining why you're doing this calculation

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -417,4 +420,395 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+      operationalWallet,
+      customerWallet.getAddress(),
+      'FOO',
+      issuerWallet.getAddress(),
+      '100',
+      0.5,
+    )
+
+    console.log('Result of sending transaction WITH sendmax')
+    console.log(transactionResult)
```
tedkalaw, line 751: remove?

test/XRP/issued-currency-client.test.ts
```diff
@@ -583,4 +584,77 @@ describe('Issued Currency Client', function (): void {
...
+        this.wallet,
+        testClassicAddress,
+        'FOO',
+        testAddress,
+        '100',
+        0.5,
+      )
+      .catch((error) => {
+        assert.equal(error.errorType, XrpErrorType.XAddressRequired)
+      })
```
tedkalaw, line 635: can we use `try/catch` here instead of a promise?

test/XRP/issued-currency-client.test.ts
```diff
@@ -583,4 +584,77 @@ describe('Issued Currency Client', function (): void {
...
+        this.wallet,
+        testAddress,
+        'FOO',
+        testClassicAddress,
+        '100',
+        0.5,
+      )
+      .catch((error) => {
+        assert.equal(error.errorType, XrpErrorType.XAddressRequired)
+      })
```
tedkalaw, line 658: same here

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -417,4 +420,395 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
     assert.equal(createdTrustLine.limit, trustLineLimit)
     assert.equal(createdTrustLine.currency, trustLineCurrency)
   })
+
+  // TODO: (acorso) can any addresses be created in `before` and reused?  Can any test cases be combined?
```
tedkalaw, line 424: it's good to have tests be simple and focused if you find yourself trying to figure out how to given/when/then it, your test is probably doing too much

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -417,4 +420,395 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+  it('sendIssuedCurrency - issuing issued currency, combined cases', async function (): Promise<
```
tedkalaw, line 425: this function is extremely core to `issuedCurrencyClient` is half the code in this test - could we put this in its own file? `issued-currency-client-sendissuedcurrency` or something?

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+    // Note that the destinationTag doesn't need to be explicitly set here, because the ripple-binary-codec will decode this X-Address and
+    // assign the decoded destinationTag before signing.
+    payment.setAmount(amount)
+
+    if (!(transferFee === undefined)) {
+      // Construct SendMax - value should be intended amount + relevant transfer fee
+      // Note that a transfer fee doesn't apply if the source address (of this transaction) and the issuing address of the currency being
+      // sent are the same (i.e. issuer sending currency directly) OR if the issuer of the currency and the destination are the same
+      // (i.e. redeeming an issued currency).
+      // However, it also doesn't hurt to include a SendMax field, it just won't nd up being used (added this to questions just to make sure).
```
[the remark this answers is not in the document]
amiecorso: I think I might end up stripping this comment out eventually but good just in case

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -417,4 +420,395 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+    void
+  > {
+    this.timeout(timeoutMs)
+    // GIVEN two existing, funded testnet accounts serving as an issuing address and operational address, respectively
+    const issuerWallet = await XRPTestUtils.randomWalletFromFaucet()
+    const operationalWallet = await XRPTestUtils.randomWalletFromFaucet()
```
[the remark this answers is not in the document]
amiecorso: sweet - I made an Asana task here to do an audit of the `IssuedCurrencyClient` integration tests all at once, and find the best structure for minimizing wallet creation, such as separate test files and re-use as much as possible: https://app.asana.com/0/1190819649053561/1198956705900885/f

I think it'll make more sense to do this all at once.

test/XRP/issued-currency-client.test.ts
amiecorso: what do you mean?

amiecorso: added

src/XRP/issued-currency-client.ts
```diff
@@ -379,4 +385,163 @@ export default class IssuedCurrencyClient {
...
+   * @param issuer The issuing address of the issued currency being sent.
+   * @param value The amount of issued currency to send.
+   * @param transferFee Optional - can be omitted if sender address or destinationAddress are the same address as the issuer.
+   *                    The transfer fee associated with the issuing account, expressed as a percentage. (i.e. a value of .5 indicates a 0.5% transfer fee).
+   */
+  // TODO: (acorso) make this private if/when incorporated into higher level convenience methods
+  // TODO: (acorso) consider using an object for potentially long list of params
+  public async sendIssuedCurrency(
+    sender: Wallet,
+    destinationAddress: string,
```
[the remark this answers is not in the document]
amiecorso: sure can

src/XRP/issued-currency-client.ts
amiecorso: So the notes on this method definitely make it seem particularly intense... (I plan to remove these soon enough).  I think we may want to expose a more opinionated API, because that will help make it clear what information needs to be provided for each case.  For example, if you want to issue some new issued currency, the API simply won't ask you for the transfer fee, because it does not apply in this case.  Similarly, if you want to redeem some issued currency, it simply won't ask you for the destination, because it'll use the same address as for `issuer` as the destination and error if there's a problem.

I think it may make sense to leave this private... but if we are able to expose all possible parameters that one *might* want to use for each distinct use case, I think we could make this the private workhorse underpinning all of those cases.

src/XRP/issued-currency-client.ts
amiecorso: that's a good question... I suppose we could allow for working with any of those types, even though they'll be converted to a string for submission to the network.  I'll make an asana ticket for this for the moment: https://app.asana.com/0/1190819649053561/1198956705900887/f

src/XRP/issued-currency-client.ts
amiecorso: will fix this when we add support for big integers etc.

src/XRP/issued-currency-client.ts
amiecorso: yar

amiecorso: this was good actually, I made this way more precise by revisiting.  The original implementation would have potentially created `sendMax` values that were much larger than necessary.  Even though it actually shouldn't typically matter if the `sendMax` value is much larger than what will actually be sent (because rippled will only apply the relevant transfer fees and take no more than that) - it could be a problem if the `sendMax` value is unreasonably high AND someone is trying to make a cross-currency payment that may involve some offers in the DEX that have unreasonably high transfer fees... it would just gobble up those expensive offers because it CAN because it has a high `sendMax`, potentially unbeknownst to the user.

amiecorso: with this implementation, the `sendMax` value should be as tight as possible while still achieving its purpose

test/XRP/issued-currency-client-integration.test.ts
amiecorso: ah ty

test/XRP/issued-currency-client.test.ts
amiecorso: mm yep will do that in all of these tests

test/XRP/issued-currency-client-integration.test.ts
amiecorso: yeah... I feel like all of these tests are going to change very soon when I instead implement specific, opinionated APIs / I made that Asana task for a full audit of these issued currency client tests... lemme think about this for a second.

src/XRP/issued-currency-client.ts
tedkalaw: cool, that sounds reasonable to me. will comment on the naming possibilties in slack

src/XRP/protobuf-wrappers/xrp-issued-currency.ts
```diff
@@ -1,3 +1,9 @@
+// TODO: friends - should we consider renaming this class as `XrplIssuedCurrency`?  (I mean... all of these wrapper classes should really
```
tedkalaw, line 1: yeah, i think that's a good thing to do - definitely a confusing name. 

want to make a new task to track it? doesn't seem like the most pressing thing ever but a good chore for sure. i'm happy to do it

test/XRP/issued-currency-client-integration.test.ts
tedkalaw: nah this is cool for now. let's roll with this and figure out a better plan for how we organize tests - this is a grander problem in the sdk

tedkalaw (approved):

src/XRP/protobuf-wrappers/xrp-issued-currency.ts
mvadari: ++

mvadari (approved):

src/XRP/protobuf-wrappers/xrp-issued-currency.ts
amiecorso: i'll make a task!

PR #835, opened 2020-10-22
Creates Web Socket Network Client

mvadari:

Gateways need a subscription method on the SDK, to be alerted when transactions occur that trigger off-ledger actions. This is very possible with the new IOU support that is currently being implemented in the SDK (for example, when IOUs are redeemed and gateways need to repay people off-ledger). 

However, the `subscribe` method in the `rippled` API requires the Web Socket API. This PR sets up a web socket network client and some basic functions that the `IssuedCurrencyClient` will interact with.

### Context of Change

A consequence of the IOU effort

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

IssuedCurrencyClient has a new method for subscribing to incoming payments for an account. 

## Test Plan

CI passes.

## Future Tasks
* Add some sort of `subscribe` method to `IssuedCurrencyClient`


src/XRP/issued-currency-client.ts
```diff
@@ -45,6 +51,8 @@ export default class IssuedCurrencyClient {
   public static issuedCurrencyClientWithEndpoint(
     grpcUrl: string,
     jsonUrl: string,
+    webSocketUrl: string,
+    handleWebSocketErrorMessage: (data: string) => void,
```
mvadari, line 55: Not sure if this is the best way to handle web socket errors

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Create a new WebSocketNetworkClient.
+   *
+   * @param webSocketUrl The URL of the rippled node to query.
+   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   */
+  public constructor(
+    webSocketUrl: string,
+    handleErrorMessage: (message: string) => void,
```
mvadari, line 36: not sure if this is the best way to handle web socket error messages

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @param request The object to send over the web socket.
+   * @returns The API response from the web socket.
+   */
+  private async sendApiRequest(
+    request: WebSocketRequestOptions,
+  ): Promise<WebSocketStatusResponse> {
+    while (this.socket.readyState === 0) {
+      await this.sleep(5)
+    }
```
mvadari, line 105: should I use something like `setInterval` instead of this while loop?

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,111 @@
...
+  SigningPubKey: string
+  TransactionType: string
+  TxnSignature: string
+  date: number
+  hash: string
+}
+
+type EmptyObject = {
+  [K in any]: never
+}
```
mvadari, line 103: This was the best way I could find of creating an empty object (used in `WebSocketStatusResponse`), if there's a better way I'd be happy to change it to that

tedkalaw (changes requested):
can we split the implementation of `WebSocketNetworkClient` and `IssuedCurrencyClient`? it would be easier to review + we could be more confident in its correctness if this were perhaps 2 perhaps, like

1) `WebSocketNetworkClient` + tests
2) Use `WebSocketNetworkClient` in `IssuedCurrencyClient`

test/XRP/web-socket-network-client.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  })
+
+  it('subscribeToAccount - bad address', async function (): Promise<void> {
+    this.timeout(timeoutMs)
+
+    const address = 'badAddress'
+    const subscriptionId = 'subscribe_transaction_' + address
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    const callback = (_data: WebSocketTransactionResponse) => {}
```
tedkalaw, line 108: i think you could just inline these

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+    const options = {
+      id,
+      command: 'subscribe',
+      accounts: [account],
+    }
+    const response = await this.sendApiRequest(options)
+    if (response.status !== 'success') {
+      throw new XrpError(
+        XrpErrorType.Unknown,
+        'Subscription request for ' + account + ' failed',
```
amiecorso, line 143: I wonder why a subscription request might fail... do you think there is any other information potentially lurking in the response that might be useful to pass to the user to help them fix their issue?

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+export default class WebSocketNetworkClient {
+  private readonly socket: WebSocket
+  private accountCallbacks: Map<
+    string,
+    (data: WebSocketTransactionResponse) => void
+  >
+  private messageCallbacks: Map<string, (data: WebSocketResponse) => void>
+  private waiting: Map<string, WebSocketResponse | undefined>
+
```
tedkalaw, line 26: since we're initializing `messageCallbacks` and `waiting` to empty maps, can we do

```
private waiting: Map<string, WebSocketResponse | undefined> = new Map()
```

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+    this.socket = new WebSocket(webSocketUrl)
+    this.accountCallbacks = new Map()
+    this.waiting = new Map()
+
+    this.messageCallbacks = new Map()
+    this.messageCallbacks.set('response', (data: WebSocketResponse) => {
+      const dataStatusResponse = data as WebSocketStatusResponse
+      this.waiting.set(dataStatusResponse.id, data)
+    })
+    this.messageCallbacks.set('transaction', this.handleTransaction)
```
tedkalaw, line 46: you could initialize it all at once if you wanted with kv pairs, like

```
this.messageCallbacks = new Map(
  ['response', thecallback],
  ['transaction, this.handleTransaction].
)
```

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+      this.handleErrorMessage('Error: ' + (event as string))
+    })
+  }
+
+  /**
+   * Properly handles incoming transactions from the web socket.
+   *
+   * @param data The web socket response received from the web socket.
+   */
+  private handleTransaction = (data: WebSocketResponse) => {
```
tedkalaw, line 77: can we use a function declaration here instead? the sdk seems to prefer that for class functions

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+ * A network client for interacting with the rippled WebSocket API.
+ * @see https://xrpl.org/rippled-api.html
+ */
+export default class WebSocketNetworkClient {
+  private readonly socket: WebSocket
+  private accountCallbacks: Map<
+    string,
+    (data: WebSocketTransactionResponse) => void
+  >
+  private messageCallbacks: Map<string, (data: WebSocketResponse) => void>
```
tedkalaw, line 24: does it work to do ```Map<[dataType: string], (data: WebSocketResponse) => void>```?

not sure if it works in generics

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+   * @param request The object to send over the web socket.
+   * @returns The API response from the web socket.
+   */
+  private async sendApiRequest(
+    request: WebSocketRequestOptions,
+  ): Promise<WebSocketStatusResponse> {
+    while (this.socket.readyState === 0) {
+      await sleep(5)
+    }
+    if (this.socket.readyState !== 1) {
```
tedkalaw, line 104: can you make an enum for the states? does one exist already in the websocket module?

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,111 @@
+/* Schema for adding type information to Web Socket objects. */
+
+/**
+ * The standard format for a request to the JSON RPC exposed by a rippled node.
+ * @see https://xrpl.org/request-formatting.html
+ */
+interface WebSocketRequestOptions {
+  command: string
```
tedkalaw, line 8: it might be nice to make an enum for the different commands: https://xrpl.org/public-rippled-methods.html. you could also use it in `subscribeToAccount`

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,111 @@
...
+  id: string
+  streams?: string[]
+  accounts?: string[]
```
amiecorso, line 11: hmmm, where do `streams` and `accounts` come from?  looking at "https://xrpl.org/request-formatting.html" I don't see them as part of a standard websocket request.  Oh I see, they're specific to the `subscribe` request... Is there a better way to create this type by keeping `id` and `command` in their own generic request type, and then maybe creating specific request types by creating a union of the generic request with the specific fields for the request?  (just thinking about this since we'll probably be adding API options to the ws client)

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,111 @@
...
+    TransactionResult: string
+    delivered_amount?: string
+  }
+  status: string
+  transaction: WebSocketTransaction
+  type: string
+  validated: boolean
+}
+
+type ChangedNode = CreatedNode | ModifiedNode | DeletedNode
```
tedkalaw, line 41: can you doc these interfaces with what they're intended for?

src/XRP/shared/rippled-web-socket-schema.ts
tedkalaw: this is super interesting. i think your type is pretty good here! would `Record<never, never>` work? `Record<string, any>` is apparently a common type for "any object"

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,111 @@
...
+  Sequence: number
+  SigningPubKey: string
+  TransactionType: string
+  TxnSignature: string
+  date: number
+  hash: string
+}
+
+type EmptyObject = {
+  [K in any]: never
```
amiecorso, line 102: what is this `[K in any]` biz?

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw, line 77: can we make this type be a `WebSocketTransactionResponse` and let the caller figure out the type? that should keep the type checking in `handleTransaction` tighter

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+   * Subscribes to incoming transactions to a given account.
+   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   *
+   * @param id The ID used for the subscription.
+   * @param callback The function called whenever a new transaction is received.
+   * @param account The account from which to subscribe to incoming transactions, encoded as a classic address.
+   * @returns The response from the websocket confirming the subscription.
+   */
+  public async subscribeToAccount(
+    id: string,
```
tedkalaw, line 129: similarly as above, thinking perhaps `subscriptionId` here - mostly because when i saw `subscribeToAccount`, i thought `id` corresponded to the account

have you played with other interfaces for this function? for some reason it feels a bit cleaner to me like

```
public async subscribeToAccount(
  account: string,
  subscriptionId: string,
  callback: callbackType,
```

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw, line 143: nit: you can use string interpolation here, like
```
`Subscription request for ${account} failed`
```

test/XRP/web-socket-network-client.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    const callback = (data: WebSocketTransactionResponse) => {
+      messageReceived = true
+      assert.equal(data.engine_result, 'tesSUCCESS')
+      assert.equal(data.engine_result_code, 0)
+      assert.equal(
+        data.engine_result_message,
+        'The transaction was applied. Only final in a validated ledger.',
+      )
+      assert.equal(data.meta.TransactionResult, 'tesSUCCESS')
+      assert.equal(data.status, 'closed')
```
amiecorso, line 63: a little confused by this `data.status` ?  What does `'closed'` mean?  I see here that the `status` field in the response should read `success` ?

test/XRP/web-socket-network-client.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  it('subscribeToAccount - bad address', async function (): Promise<void> {
+    this.timeout(timeoutMs)
+
+    const address = 'badAddress'
+    const subscriptionId = 'subscribe_transaction_' + address
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    const callback = (_data: WebSocketTransactionResponse) => {}
+    // GIVEN a test address that has at least one trust line on testnet
```
amiecorso, line 109:
```suggestion
    // GIVEN a test address that is malformed.
```

amiecorso (changes requested):
Looking great!  Mostly just comments and questions -- going to request changes just so that I'm forced to look at this again before it's approved.  Ping me once Ted has reviewed and you've had a chance for some responses.

src/XRP/network-clients/web-socket-network-client.ts
mvadari: So I should initialize it in this declaration instead of in the constructor?

src/XRP/network-clients/web-socket-network-client.ts
mvadari: I think that would look messier with the `response` callback

src/XRP/network-clients/web-socket-network-client.ts
mvadari: Nope, it gets mad at me

src/XRP/network-clients/web-socket-network-client.ts
mvadari: I get an error if I do that, where I set it in `messageCallbacks`

src/XRP/network-clients/web-socket-network-client.ts
mvadari: Couldn't find one quickly so I just made one

src/XRP/shared/rippled-web-socket-schema.ts
mvadari: Should I dump the whole list there now? I created the enum for now, but it's only got the one method atm

src/XRP/network-clients/web-socket-network-client.ts
mvadari: The only caller of this function is the `message` event listener, which doesn't know anything specific about the handler and just knows it takes a `WebSocketResponse` object, so that won't work.

src/XRP/network-clients/web-socket-network-client.ts
mvadari: I'm not picky about parameter ordering, it's mainly ordered that way based on how I added params as I developed the function.

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+      await sleep(5)
+    }
+    const response = this.waiting.get(request.id)
+    this.waiting.delete(request.id)
+
+    return response as WebSocketStatusResponse
+  }
+
+  /**
+   * Subscribes to incoming transactions to a given account.
```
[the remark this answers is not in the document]
mvadari: Hm I ran some tests earlier and didn't receive information about outgoing transactions, which is why I left the doc like that, but I just tried again and it worked, so I'll fix that.

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,155 @@
...
+   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   *
+   * @param id The ID used for the subscription.
+   * @param callback The function called whenever a new transaction is received.
+   * @param account The account from which to subscribe to incoming transactions, encoded as a classic address.
```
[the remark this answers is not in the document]
mvadari: I was thinking that if you added multiple accounts, you'd probably want a different callback for each account, rather than having them all feed through the same function (and a list of functions is a little messy). 

I certainly can do that, but I could also just add another method (maybe in another PR?) for the option to do multiple accounts.

src/XRP/shared/rippled-web-socket-schema.ts
mvadari: This was the only way I could find to create an empty object (`{}`)

test/XRP/web-socket-network-client.test.ts
mvadari: My understanding is that this refers to the status of the transaction/ledger referred to in this response

see https://xrpl.org/subscribe.html#transaction-streams

src/XRP/network-clients/web-socket-network-client-interface.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The WebSocketNetworkClientInterface defines the calls available via the rippled WebSocket API.
+ */
+export interface WebSocketNetworkClientInterface {
+  /**
+   * Subscribes to incoming transactions to a given account.
+   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   *
+   * @param subscriptionId The ID used for the subscription.
+   * @param callback The function called whenever a new transaction is received.
+   * @param account The account from which to subscribe to incoming transactions, encoded as an X-Address.
```
tedkalaw, line 16: extreme nit - ordering of params is different from the function signature

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: i think that would be reasonable yeah. up to you though

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: yeah don't do this it looks horrible lol

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -0,0 +1,169 @@
+/* Schema for adding type information to Web Socket objects. */
+
+enum WebSocketReadyState {
+  CONNECTING,
+  OPEN,
+  CLOSING,
+  CLOSED,
```
tedkalaw, line 7: is the upper cased-ness of this how ws likes it? typically you case things like `Connecting`

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: yeah, that sounds right. `this` behaves in kind of an interesting way in JS so you'll have to do

```
(data: WebSocketResponse) => this.handleTransaction(data)
```

since we seem to have chosen to go with function declarations, seems better to be standardized here

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,169 @@
...
+      } else {
+        this.handleErrorMessage(
+          'Received unhandlable message: ' + (event.data as string),
+        )
+      }
+    })
+
+    this.socket.addEventListener('close', (event) => {
+      this.handleErrorMessage(
+        'Web socket disconnected, ' + (event.reason as string),
```
tedkalaw, line 66: you can use string interpolation here

`Web socket disconnected, ${event.reason as string}`

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,169 @@
...
+      )
+    })
+
+    this.socket.addEventListener('error', (event) => {
+      this.handleErrorMessage('Error: ' + (event as string))
```
tedkalaw, line 71: same here

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: well, it's only called to handle the `transaction` message type, right? you can enforce the type there by doing

```
this.messageCallbacks.set(
  'transaction', 
  (data: WebSocketResponse) => this.handleTransaction(data as WebSocketTransactionResponse),
)
```

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -0,0 +1,169 @@
...
+    this.messageCallbacks = new Map()
+    this.messageCallbacks.set('response', (data: WebSocketResponse) => {
+      const dataStatusResponse = data as WebSocketStatusResponse
+      this.waiting.set(dataStatusResponse.id, data)
+    })
+    this.messageCallbacks.set('transaction', this.handleTransaction)
+
+    this.socket.addEventListener('message', (event) => {
+      const parsedData = JSON.parse(event.data) as WebSocketResponse
+      const dataType = parsedData.type
```
tedkalaw, line 53: can we call this `messageType`? seems like this is the type of the message associated with the payload

src/XRP/network-clients/web-socket-network-client.ts
mvadari: Unfortunately, Typescript gets mad because it thinks the variable `event` isn't being used if I do that

src/XRP/network-clients/web-socket-network-client.ts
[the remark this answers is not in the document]
amiecorso: That may be true... I could also see the case where we'd want the same callback for any account, where maybe we update something for the account in some sort of visualizer or something, in an identical way for a variety of accounts.  However, I think it would make perfect sense to do this as a separate API in the future so we can roll with this for now.

test/XRP/web-socket-network-client.test.ts
amiecorso: hmm, yeah that's the only context in which I've really heard "closed" used as well.. and the reason that concerns me is that I think we should only be pulling data from *validated* ledgers - not open or closed...  also, the `data.validated` field is `true`... 

Looking at the documentation, I don't see any other options to pass into the request to specify validated-only ledgers, and the documentation indicates that subscribing to an account *should* only return data from validated ledgers.  This, combined with the fact that the `validated` field is indeed `true` makes me feel good enough about things... but it's a bit weird.  We can keep our radar out for this weirdness when we build our demo.

amiecorso (approved):

test/XRP/web-socket-network-client.test.ts
mvadari: Giving an `accounts` field means that we only get finalized transactions - if we used `accounts_proposed` instead, we'd also get not-finalized transactions

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: wow that's super annoying lol

tedkalaw (approved):
looks great!!

PR #843, opened 2020-10-26
Adds IssuedCurrencyClient::authorizeTrustLine

tedkalaw:
Adds `IssuedCurrencyClient::authorizeTrustLine`, which is used to create an authorized trust line between two accounts.

Most of the work in here was me adding a helper, since we'll be sending other kinds of trust set transactions.

### Context of Change
Authorized trust lines make it so that a party can't send you an issued currency unless you allow it.

See https://xrpl.org/authorized-trust-lines.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Authorized trust lines can be created between two accounts.

## Test Plan
CI, the unit test I added

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -10,6 +10,7 @@ import {
 import {
   TrustSet,
   AccountSet,
+  Transaction,
 } from './Generated/web/org/xrpl/rpc/v1/transaction_pb'
 
 import isNode from '../Common/utils'
@@ -25,6 +26,7 @@ import TransactionResult from './shared/transaction-result'
 import { AccountLinesResponse } from './shared/rippled-json-rpc-schema'
 import TrustLine from './shared/trustline'
 import { TransferRate } from './Generated/node/org/xrpl/rpc/v1/common_pb'
+import TrustSetFlag from './shared/trust-set-flag'
 
 /**
  * IssuedCurrencyClient is a client for working with Issued Currencies on the XRPL.
@@ -329,13 +331,67 @@ export default class IssuedCurrencyClient {
     amount: string,
     wallet: Wallet,
   ): Promise<TransactionResult> {
-    if (!XrpUtils.isValidXAddress(issuerXAddress)) {
+    const trustSetTransaction = await this.prepareTrustSetTransaction(
+      issuerXAddress,
+      currencyName,
+      amount,
+      wallet,
+    )
+
+    const transactionHash = await this.coreXrplClient.signAndSubmitTransaction(
+      trustSetTransaction,
+      wallet,
+    )
+
+    return await this.coreXrplClient.getFinalTransactionResultAsync(
+      transactionHash,
+      wallet,
+    )
+  }
+
+  public async authorizeTrustLine(
+    accountToTrust: string,
+    currencyName: string,
+    wallet: Wallet,
+  ): Promise<TransactionResult> {
+    const trustSetTransaction = await this.prepareTrustSetTransaction(
+      accountToTrust,
+      currencyName,
+      '0',
+      wallet,
+    )
+
+    const flags = new Flags()
+    flags.setValue(TrustSetFlag.tfSetfAuth)
+    trustSetTransaction.setFlags(flags)
+
+    const transactionHash = await this.coreXrplClient.signAndSubmitTransaction(
+      trustSetTransaction,
+      wallet,
+    )
+
+    return await this.coreXrplClient.getFinalTransactionResultAsync(
+      transactionHash,
+      wallet,
+    )
+  }
+
+  private async prepareTrustSetTransaction(
+    accountToTrust: string,
+    currencyName: string,
+    amount: string,
+    wallet: Wallet,
+  ): Promise<Transaction> {
+    if (!XrpUtils.isValidXAddress(accountToTrust)) {
       throw XrpError.xAddressRequired
     }
-    const classicAddress = XrpUtils.decodeXAddress(issuerXAddress)
+    const classicAddress = XrpUtils.decodeXAddress(accountToTrust)
     if (!classicAddress) {
       throw XrpError.xAddressRequired
     }
+    // TODO (tedkalaw): Use X-Address directly when ripple-binary-codec supports X-Addresses.
+    const issuerAccountAddress = new AccountAddress()
+    issuerAccountAddress.setAddress(classicAddress.address)
 
     if (currencyName === 'XRP') {
       throw new XrpError(
@@ -343,11 +399,6 @@ export default class IssuedCurrencyClient {
         'createTrustLine: Trust lines can only be created for Issued Currencies',
       )
     }
-
-    // TODO (tedkalaw): Use X-Address directly when ripple-binary-codec supports X-Addresses.
-    const issuerAccountAddress = new AccountAddress()
-    issuerAccountAddress.setAddress(classicAddress.address)
-
     const currency = new Currency()
     currency.setName(currencyName)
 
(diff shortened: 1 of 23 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -329,25 +331,108 @@ export default class IssuedCurrencyClient {
...
+      currencyName,
+      amount,
+      undefined,
+      wallet,
+    )
+  }
+
+  /**
+   * Creates an authorized trust line between this XRPL account and another account.
+   * An authorized trust line will only be created if `requiredAuthorizedTrustlines` is called first.
```
tedkalaw, line 345: what sucks is that if you DO NOT called `requireAuthorizedTrustlines` first, it will not error - no trust line will be created instead. as such, i *could* check to see if the correct account flag is set, but then we take on the additional cost of another fetch from the ledger.

i propose we wait until we are building out our demo to see how annoying this is in practice, but let me know what you think

src/XRP/issued-currency-client.ts
```diff
@@ -329,25 +331,108 @@ export default class IssuedCurrencyClient {
...
+
+    const transactionHash = await this.coreXrplClient.signAndSubmitTransaction(
+      trustSetTransaction,
+      wallet,
+    )
+
+    return await this.coreXrplClient.getFinalTransactionResultAsync(
+      transactionHash,
+      wallet,
+    )
```
tedkalaw, line 400: split this into 2 because `prepareTrustSetTransaction` has a lot of state changing going on

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -33,7 +33,7 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
 
   // A Wallet with some balance on Testnet.
   let wallet: Wallet
-  before(async function () {
+  beforeEach(async function () {
```
tedkalaw, line 36: `before` is called once before the entire test suite is run. this means that the `wallet` object in every test was actually the same, making the tests have side-effects on each other. 

this was causing my new test to fail, because a trust line were previously created from. this sort of sucks because it's slower but it makes the tests easier to reason about

src/XRP/issued-currency-client.ts
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+      currencyName,
+      amount,
+      undefined,
+      wallet,
+    )
+  }
+
+  /**
+   * Creates an authorized trust line between this XRPL account and another account.
+   * An authorized trust line will only be created if `requiredAuthorizedTrustlines` is called first.
```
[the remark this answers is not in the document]
amiecorso: I refreshed my reading on authorized trust lines and I propose the documentation for this method reads more like this:
``` Creates an authorized trust line between this XRPL account (issuing account) and another account.  Note that the other account must also create a trust line to this issuing account in order to establish a trust line with a non-zero limit.  If this method is called before the other account creates a trust line, a trust line with a limit of 0 is created.  However, this is only true if this issuing account has already required Authorized Trustlines (see https://xrpl.org/authorized-trust-lines.html), otherwise no trust line is created. ```

File: src/XRP/issued-currency-client.ts [hunk 5]
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @see https://xrpl.org/authorized-trust-lines.html
+   *
+   * @param accountToAuthorize The X-Address of the address to authorize.
```
amiecorso, line 369:
```suggestion
   * @param accountToAuthorize The X-Address of the account with which to authorize a trust line.
```

src/XRP/issued-currency-client.ts
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+   * @param currencyName The currency to authorize a trust line for.
+   * @param wallet The wallet creating the trust line.
```
amiecorso, line 371:
```suggestion
   * @param wallet The wallet creating the authorized trust line.
```

src/XRP/issued-currency-client.ts
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+      accountToAuthorize,
+      currencyName,
+      '0',
+      TrustSetFlag.tfSetfAuth,
+      wallet,
+    )
+  }
+
+  /*
+   * Creates and sends a TrustSet transaction to be sent and executed on the XRPL.
```
amiecorso, line 390:
```suggestion
   * Creates and sends a TrustSet transaction to the XRPL.
```

src/XRP/issued-currency-client.ts
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+   * @param accountToTrust The account to extend trust to with a trust line.
+   * @param currencyName The name of the currency to create a trust line for.
+   * @param amount The maximum amount of debt to allow on this trust line.
+   * @param wallet A wallet associated with the account extending trust.
+   */
+  private async sendTrustSetTransaction(
+    accountToTrust: string,
+    currencyName: string,
+    amount: string,
+    flags: number | undefined,
```
amiecorso, line 401: Would this not be better as an optional param?

src/XRP/issued-currency-client.ts
```diff
@@ -349,25 +351,108 @@ export default class IssuedCurrencyClient {
...
+   * @param accountToTrust The account to extend trust to with a trust line.
+   * @param currencyName The name of the currency to create a trust line for.
+   * @param amount The maximum amount of debt to allow on this trust line.
+   * @param wallet A wallet associated with the account extending trust.
+   */
+  private async prepareTrustSetTransaction(
+    accountToTrust: string,
+    currencyName: string,
+    amount: string,
+    flags: number | undefined,
```
amiecorso, line 435: same q about optional param

test/XRP/issued-currency-client-integration.test.ts
amiecorso: yeah... my `sendIssuedCurrency` integration tests are super wallet-heavy and I end up creating new wallets in each one to avoid side effects as well.  However, maybe we can figure out which tests don't cause side effects (or aren't affected by previous side effects) and use the same wallet across all of those as established in a `before`, and otherwise establish wallets in each test on an as-needed basis, instead of calling `beforeEach` for EVERY single test.. which may slow things down...

amiecorso: Any chance you can put this back to `before` and create your new wallets inside your tests specifically?

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -420,4 +420,34 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+
+    const [createdTrustLine] = trustLines
+    const classicAddress = XrpUtils.decodeXAddress(accountToTrust.getAddress())!
+
+    // THEN an authorized trust line is created between the wallet and the other account.
+    assert.equal(createdTrustLine.account, classicAddress.address)
+    assert.equal(createdTrustLine.limit, '0')
+    assert.equal(createdTrustLine.currency, trustLineCurrency)
+    assert.isTrue(createdTrustLine.authorized)
+  })
```
amiecorso, line 452: Should we add any additional test cases here?  For example, what if we created a test where we first establish a non-zero trustline from the customer to the issuer, and THEN have the issuer call `authorizeTrustLine` and observe that the trust line exists and has a non-0 limit?  Or a test where the issuer requires auth trust lines, does not authorize a line to a customer, and then tries to send money to the customer and it fails (<-- actually, this case I think belongs more in `sendIssuedCurrency` and is currently an open TODO for me, pending this PR, so I can take care of that one).

amiecorso (changes requested):

src/XRP/issued-currency-client.ts
tedkalaw: whoops, just got this one

src/XRP/issued-currency-client.ts
tedkalaw: i think in general our functions should have `wallet` first, but i noticed that pattern in a lot of other places in the sdk. having the most "important"/fixed params earlier typically feels more logical to me, but also unlocks things like currying: https://javascript.info/currying-partials

test/XRP/issued-currency-client-integration.test.ts
amiecorso: yeah I've been having a bit of hard time trying to think about where the boundaries lie in terms of logic that should be tested here v.s. logic that is part of the XRPL and shouldn't be tested here.... I'm fine leaving this test file as-is for now.  A demo will be the real test.

amiecorso (approved):

mvadari (approved):

PR #849, opened 2020-10-27
Creates `monitorAccountTransactions` on `IssuedCurrencyClient`

mvadari:

Gateways need a subscription method on the SDK, to be alerted when transactions occur that trigger off-ledger actions. This is very possible with the new IOU support that is currently being implemented in the SDK (for example, when IOUs are redeemed and gateways need to repay people off-ledger). This PR adds that subscription method for accounts. 

### Context of Change

A consequence of the IOU effort

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

IssuedCurrencyClient has a new method for subscribing to incoming payments for an account. 

## Test Plan

CI passes.

## Future Tasks
* More robust testing, to cover various use cases (such as sending issued currency payments). 
* Add an unsubscribe method.

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -220,7 +220,7 @@ export default class IssuedCurrencyClient {
       throw XrpError.xAddressRequired
     }
 
-    const id = 'monitor_transactions_${account}'
+    const id = `monitor_transactions_${account}`
 
     return await this.webSocketNetworkClient.subscribeToAccount(
       classicAddress.address,
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -38,10 +38,7 @@ import GatewayBalances, {
   gatewayBalancesFromResponse,
 } from './shared/gateway-balances'
 import TrustLine from './shared/trustline'
-import {
-  WebSocketStatusResponse,
-  TransactionResponse,
-} from './shared/rippled-web-socket-schema'
+import { TransactionResponse } from './shared/rippled-web-socket-schema'
 import { WebSocketNetworkClientInterface } from './network-clients/web-socket-network-client-interface'
 import WebSocketNetworkClient from './network-clients/web-socket-network-client'
 import { SendMax } from 'xpring-common-js/build/src/XRP/generated/org/xrpl/rpc/v1/common_pb'
@@ -215,7 +212,7 @@ export default class IssuedCurrencyClient {
   public async monitorAccountTransactions(
     account: string,
     callback: (data: TransactionResponse) => void,
-  ): Promise<WebSocketStatusResponse> {
+  ): Promise<boolean> {
     const classicAddress = XrpUtils.decodeXAddress(account)
     if (!classicAddress) {
       throw XrpError.xAddressRequired
@@ -223,11 +220,12 @@ export default class IssuedCurrencyClient {
 
     const id = `monitor_transactions_${account}`
 
-    return await this.webSocketNetworkClient.subscribeToAccount(
+    const response = await this.webSocketNetworkClient.subscribeToAccount(
       classicAddress.address,
       id,
       callback,
     )
+    return response.status === 'success'
   }
 
   /**
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -26,8 +26,8 @@ import { AccountLinesResponse } from './shared/rippled-json-rpc-schema'
 import TrustLine from './shared/trustline'
 import { TransferRate } from './Generated/node/org/xrpl/rpc/v1/common_pb'
 import {
-  WebSocketResponse,
   WebSocketStatusResponse,
+  WebSocketTransactionResponse,
 } from './shared/rippled-web-socket-schema'
 import { WebSocketNetworkClientInterface } from './network-clients/web-socket-network-client-interface'
 import WebSocketNetworkClient from './network-clients/web-socket-network-client'
@@ -117,7 +117,7 @@ export default class IssuedCurrencyClient {
 
   public async monitorIncomingPayments(
     account: string,
-    callback: (data: WebSocketResponse) => void,
+    callback: (data: WebSocketTransactionResponse) => void,
   ): Promise<WebSocketStatusResponse> {
     const id = 'monitor_transactions_' + account
     return await this.webSocketNetworkClient.subscribeToAccount(
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -40,7 +40,7 @@ import GatewayBalances, {
 import TrustLine from './shared/trustline'
 import {
   WebSocketStatusResponse,
-  WebSocketTransactionResponse,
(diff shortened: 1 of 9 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,34 @@ export default class IssuedCurrencyClient {
...
+   */
+  public async monitorIncomingPayments(
+    account: string,
+    callback: (data: WebSocketTransactionResponse) => void,
+  ): Promise<WebSocketStatusResponse> {
+    const classicAddress = XrpUtils.decodeXAddress(account)
+    if (!classicAddress) {
+      throw XrpError.xAddressRequired
+    }
+    const address = classicAddress.address
```
tedkalaw, line 222: typescript snob nit - it's common these days to do `const {address} = classicAddress` i guess. some of our other repos have this turned on as a lint rule

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,34 @@ export default class IssuedCurrencyClient {
...
+
+    const id = 'monitor_transactions_' + account
```
tedkalaw, line 224: can you use interpolation here?

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -21,22 +23,35 @@ const testAddressWithTrustLines =
...
+  let wallet2: Wallet
+  beforeEach(async function () {
     this.timeout(timeoutMs)
     wallet = await XRPTestUtils.randomWalletFromFaucet()
+    wallet2 = await XRPTestUtils.randomWalletFromFaucet()
+  })
+
+  after(function (done) {
+    issuedCurrencyClient.webSocketNetworkClient.close()
+    done()
```
tedkalaw, line 54: nice!

tedkalaw (approved):
this pr is super based ty

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,33 @@ export default class IssuedCurrencyClient {
     }
   }
 
+  /**
+   * Subscribes to an account's incoming transactions, and triggers a callback upon recieving each transaction.
```
amiecorso, line 206:
```suggestion
   * Subscribes to all transactions that affect the specified account, and triggers a callback upon receiving each transaction.
```

amiecorso: Correct me if I'm wrong, but from what I understood in the docs and I think we discussed, there will be an event for any transaction that affects this account, including an outgoing transaction that modifies its balance.

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,33 @@ export default class IssuedCurrencyClient {
...
+   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   * @see https://xrpl.org/subscribe.html
+   *
+   * @param account The account from which to subscribe to incoming transactions, encoded as an X-Address.
```
amiecorso, line 210:
```suggestion
   * @param account The account for which to subscribe to relevant transactions, encoded as an X-Address.
```

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,33 @@ export default class IssuedCurrencyClient {
...
+   * @param callback The function to trigger upon recieving each transaction from the ledger.
```
amiecorso, line 211:
```suggestion
   * @param callback The function to trigger upon receiving a transaction event from the ledger.
```

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,33 @@ export default class IssuedCurrencyClient {
...
+   * @returns The response from the websocket confirming the subscription.
```
amiecorso, line 212:
```suggestion
   * @returns The response from the websocket indicating the result of the subscription request.
```

amiecorso: I'm wondering if it might not make sense to process the response a little bit in this method instead of exposing the raw `WebSocketStatusResponse`?  for example, returning just an indicator of success or failure... I don't know, that might cause some information loss such that an alternative solution wouldn't really be any cleaner than what we already have...

src/XRP/issued-currency-client.ts
```diff
@@ -192,6 +202,33 @@ export default class IssuedCurrencyClient {
...
+   */
+  public async monitorIncomingPayments(
```
amiecorso, line 214: again, I think we need to be clear on what we're monitoring here... if this is subscribing to all transactions that affect the account, that might involve outgoing payments, as well as other transaction types that modify the account settings, trust lines, or other objects.  We should probably call this `monitorAccountTransactions` or something more generic.

amiecorso: unless we want to get fancy and provide our own callback that takes a callback, where our callback first filters things for payments-only and then iff the transaction was a payment executes the user's callback.  I think that could be achieved with like... currying or some sort of functional fanciness?  I don't think we need to do that though, I think we're better off just naming this better

amiecorso (changes requested):
this PR is indeed based.  few questions mostly about names and docs

src/XRP/issued-currency-client.ts
mvadari: Ah yeah missed that when merging those changes in

src/XRP/issued-currency-client.ts
mvadari: I wasn't really sure what to return tbh - I could just return the status string. There isn't much else returned in the status object, so it's fine. 
```interface WebSocketStatusResponse {
  id: string
  result: WebSocketTransactionResponse | EmptyObject
  status: string
  type: string
}

tedkalaw: i do think it's a little weird that we're exposing some details about the implementation here. specifically, we're returning a `WebSocketTransactionResponse`, but i would expect `IssuedCurrencyClient` to abstract away the concept of `WebSocket`, if that make sense

mvadari: Unless I just change the name of `WebSocketTransactionResponse`, I'm not really sure how to fix that, because you _are_ expecting the whole transaction response object.

amiecorso: Wait I don't know if we're talking about the same thing here:  we're looking at `monitorAccountTransactions` right? this method just needs to return an indication of success around establishing the subscription... the `WebSocketTransactionResponse` will be delivered to the registered callback upon transaction events... so that's a different thing entirely.  I think this method, `monitorAccountTransactions` could probably just return the status string `success`?  Or an error?  (Returning a string still seems a little weird?  is there a better or more conventional way to indicate success?)

amiecorso (approved):
just added one more comment to our discussion about what to return from `monitorAccountTransactions` so check that out but otherwise I think this looks good!

src/XRP/issued-currency-client.ts
mvadari: I was thinking I could just return `true` maybe? (or whether the status is `'success'`)

PR #850, opened 2020-10-27
Eliminates use of the JSON RPC Network Client

mvadari:

The Issued Currency client currently has 3 different network clients attached to it: gRPC, JSON RPC, and Web Socket. This is a little excessive. All the API endpoints called by the JSON API can also be called by the Web Socket API, so this PR removes all traces of the JSON API calls, migrating them to the Web Socket API. 

### Context of Change

IOU MVP

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

The JSON RPC network client is no longer used and no longer exists. 

## Test Plan

CI passes.


src/XRP/network-clients/web-socket-network-client-interface.ts
```diff
@@ -22,5 +24,30 @@ export interface WebSocketNetworkClientInterface {
...
+  getAccountLines(
+    account: string,
+    peerAccount?: string,
+  ): Promise<WebSocketAccountLinesResponse>
+
+  /**
+   * Submits a gateway_balances request to the rippled WebSocket API.
+   * @see https://xrpl.org/gateway_balances.html
+   *
+   * @param account The XRPL account for which to retrieve balances.
```
amiecorso, line 43:
```suggestion
   * @param account The XRPL account for which to retrieve issued currency balances.
```

amiecorso: I know I wrote this doc originally -- think this will be a bit better

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -130,7 +136,7 @@ export default class WebSocketNetworkClient {
 
   /**
    * Subscribes for notification about every validated transaction that affects the given account.
-   * @see https://xrpl.org/monitor-incoming-payments-with-websocket.html
+   * @see https://xrpl.org/subscribe.html
```
amiecorso, line 139: nice

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -159,6 +165,53 @@ export default class WebSocketNetworkClient {
...
+    }
+    const accountLinesResponse = await this.sendApiRequest(accountLinesRequest)
+    return accountLinesResponse as WebSocketAccountLinesResponse
+  }
+
+  /**
+   * Submits a gateway_balances request to the rippled WebSocket API.
+   * @see https://xrpl.org/gateway_balances.html
+   *
+   * @param account The XRPL account for which to retrieve balances.
```
amiecorso, line 193:
```suggestion
   * @param account The XRPL account for which to retrieve issued currency balances.
```

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -159,6 +165,53 @@ export default class WebSocketNetworkClient {
...
+   * @param addressesToExclude (Optional) An array of operational address to exclude from the balances issued.
+   * @see https://xrpl.org/issuing-and-operational-addresses.html
+   */
+  public async getGatewayBalances(
+    account: string,
+    addressesToExclude?: Array<string>,
+  ): Promise<WebSocketGatewayBalancesResponse> {
+    const gatewayBalancesRequest = {
+      id: 'gateway_balances_' + account,
```
amiecorso, line 202: what if multiple requests for gateway balances for the same account are made in a short window?  do we need an additional piece of identifying information here?  Oh... I guess they won't be because the actual implementation of this on the issued currency client is going to block until the response comes back... so even if there are successive calls to `getGatewayBalances` using the issued currency client, we'll essentially always be dealing with just one thing in the waiting queue anyway.  Tell me if I'm wrong but I think you're good

amiecorso (changes requested):
beautiful

src/XRP/network-clients/web-socket-network-client.ts
mvadari: I was thinking about adding an extra number that keeps increasing to the end of the IDs here to avoid that problem - lmk if you think that's worth

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -159,6 +165,53 @@ export default class WebSocketNetworkClient {
...
+      command: 'gateway_balances',
+      account,
+      strict: 'true',
```
tedkalaw, line 205: should this be `true` instead of `'true'`? in `GatewayBalancesRequest` below, the type seems to suggest that this is a `boolean` instead of a string

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -80,6 +123,70 @@ interface WebSocketTransactionResponse {
...
+
+interface WebSocketAccountLinesFailureResponse {
+  id: number | string
+  status: string
+  type: string
+  error: string
+  error_code: number
+  error_message: string
+  request: AccountLinesRequest
+}
```
tedkalaw, line 155: i'll need to look at the other responses some more, but these seem to share some fields. can we make a base type that they extend? like

```
interface BaseWebSocketAccountLinesResponse {
  id: number | string
  status: string
  type: string
}

interface WebSocketAccountLinesSuccessfulResponse extends BaseWebSocketAccountLinesResponse {
  result: // result thing
}

interface WebSocketAccountLinesFailureResponse extends Base {
  // failure response specific fields
}
```

src/XRP/network-clients/web-socket-network-client.ts
tedkalaw: you'll probably want to do that eventually - might be worth tracking with a task right now and leaving as is for now

mvadari: Just did it now

src/XRP/shared/rippled-web-socket-schema.ts
mvadari: I handle that in #860 but I can move it to here if you think that would be better

amiecorso (approved):
I'm down with you refactoring the types in the other PR - looks good

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -28,7 +33,11 @@ export default class WebSocketNetworkClient {
     string,
     (data: WebSocketResponse) => void
   > = new Map()
-  private waiting: Map<string, WebSocketResponse | undefined> = new Map()
+  private waiting: Map<
+    number | string,
+    WebSocketResponse | undefined
+  > = new Map()
+  private idNumber = 0
```
tedkalaw, line 40: maybe a small comment describing what this var is for?

src/XRP/shared/rippled-web-socket-schema.ts
tedkalaw: nah that's cool

tedkalaw (approved):
lgtm!!

PR #851, opened 2020-10-28
Adds IssuedCurrencyClient::freezeTrustLine

tedkalaw:
This PR adds `IssuedCurrencyClient::freezeTrustLine` and a test.

The next PR will add `unfreezeTrustLine`.

### Context of Change
We already have account level freezing. This PR adds it for individual trust lines.

https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`IssuedCurrencyClient` has a `freezeTrustLine` function

## Test Plan
CI, the test I added

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -459,7 +459,7 @@ export default class IssuedCurrencyClient {
    *
    * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
    *
-   * @param accountToFreeze The X-Address of the account which to to freeze a trust line.
+   * @param accountToFreeze The X-Address of the account involved in the trust line being frozen.
    * @param currencyName The currency of the trust line to freeze.
    * @param wallet The wallet freezing the trust line.
    */
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -10,6 +10,7 @@ import {
 import {
   TrustSet,
   AccountSet,
+  Transaction,
 } from './Generated/web/org/xrpl/rpc/v1/transaction_pb'
 
 import isNode from '../Common/utils'
@@ -25,6 +26,7 @@ import TransactionResult from './shared/transaction-result'
 import { AccountLinesResponse } from './shared/rippled-json-rpc-schema'
 import TrustLine from './shared/trustline'
 import { TransferRate } from './Generated/node/org/xrpl/rpc/v1/common_pb'
+import TrustSetFlag from './shared/trust-set-flag'
 
 /**
  * IssuedCurrencyClient is a client for working with Issued Currencies on the XRPL.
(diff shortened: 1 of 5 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -389,6 +389,22 @@ export default class IssuedCurrencyClient {
...
+    accountToFreeze: string,
+    currencyName: string,
+    wallet: Wallet,
+  ): Promise<TransactionResult> {
+    return await this.sendTrustSetTransaction(
+      accountToFreeze,
+      currencyName,
+      // You can change the trust line when you freeze it, but an amount of 0
+      // would be the most conservative amount.
+      '0',
```
tedkalaw, line 412: so this is pretty opinionated of me. however, you would typically freeze a trust line if something sketchy happened, and the actual balance can still be reduced by the counter party sending ious back to the issuer. it didn't really feel like it made too much sense to have an exposed `amount` param when all you want to do is freeze the tl

lemme know what you think!

mvadari: seems reasonable - it appears to also be recommended

mvadari (approved):

src/XRP/issued-currency-client.ts
amiecorso: agreed - we can change this in the future if we really have to but I think the benefits are worth it

src/XRP/issued-currency-client.ts
```diff
@@ -453,6 +453,32 @@ export default class IssuedCurrencyClient {
...
+   * Note that the trust line's limit is set to 0.
+   *
+   * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
+   *
+   * @param accountToFreeze The X-Address of the account which to to freeze a trust line.
+   * @param currencyName The currency of the trust line to freeze.
+   * @param wallet The wallet freezing the trust line.
+   */
+  public async freezeTrustLine(
+    accountToFreeze: string,
```
amiecorso, line 467: I think that in keeping with the rest of the SDK, we need to check that this is an X-Address and throw if it's not?  We should consider as a group whether it makes sense to ALWAYS require X-Addresses everywhere even in situations where there can't possibly BE a tag... in which case we might want to support both address types.

src/XRP/issued-currency-client.ts
```diff
@@ -453,6 +453,32 @@ export default class IssuedCurrencyClient {
     )
   }
 
+  /**
+   * Freezes the trust line between this account (issuing account) and another account.
+   * Note that the trust line's limit is set to 0.
+   *
+   * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
+   *
+   * @param accountToFreeze The X-Address of the account which to to freeze a trust line.
```
amiecorso, line 462: I think we actually need a better name for this parameter, since technically you can actually freeze a whole account (that's what the global freeze is).  Don't want to create confusion there... we could call it `counterpartyAccount` although we've never used that term anywhere else... we could call it `accountOnTrustline` or `trustlineAccount` or `trustlineCounterparty` or....  And actually, maybe we could rename `currencyName` below to `trustlineCurrencyName` and then use `trustlineAccount` to make a little more clear and consistent that these data are what specify the trustline that we're freezing.

amiecorso, line 462:
```suggestion
   * @param accountToFreeze The X-Address of the account involved in the trust line being frozen.
```

amiecorso: lol

amiecorso (changes requested):
Just the one note about verifying the X-address of the accountToFreeze

amiecorso (approved):

src/XRP/issued-currency-client.ts
tedkalaw: gotcha, that's a great point

PR #858, opened 2020-10-29
Adds IssuedCurrencyClient::unfreezeTrustLine

tedkalaw:
This PR adds IssuedCurrencyClient::unfreezeTrustLine and a test.

### Context of Change
This is the opposite of https://github.com/xpring-eng/Xpring-JS/pull/851, and allows unfreezing a frozen trust line.

https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`IssuedCurrencyClient` has `unfreezeTrustLine`

## Test Plan
CI, test

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -479,6 +479,22 @@ export default class IssuedCurrencyClient {
     )
   }
 
+  public async unfreezeTrustLine(
+    accountToUnfreeze: string,
+    currencyName: string,
+    wallet: Wallet,
+  ): Promise<TransactionResult> {
+    return await this.sendTrustSetTransaction(
+      accountToUnfreeze,
+      currencyName,
+      // You can change the trust line when you unfreeze it, but this would typically
+      // be used by gateways, who will maintain an amount of 0.
+      '0',
+      TrustSetFlag.tfClearFreeze,
+      wallet,
+    )
+  }
+
   /*
    * Creates and sends a TrustSet transaction to the XRPL.
    *
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -10,6 +10,7 @@ import {
 import {
   TrustSet,
   AccountSet,
+  Transaction,
 } from './Generated/web/org/xrpl/rpc/v1/transaction_pb'
 
 import isNode from '../Common/utils'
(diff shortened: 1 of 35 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -479,6 +479,32 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @param accountToUnfreeze The X-Address of the account which to unfreeze a trust line.
+   * @param currencyName The currency of the trust line to unfreeze.
+   * @param wallet The wallet unfreezing the trust line.
+   */
+  public async unfreezeTrustLine(
+    accountToUnfreeze: string,
+    currencyName: string,
+    wallet: Wallet,
+  ): Promise<TransactionResult> {
```
mvadari, line 496: Should there be an optional parameter of amount?

src/XRP/issued-currency-client.ts
```diff
@@ -479,6 +479,32 @@ export default class IssuedCurrencyClient {
     )
   }
 
+  /**
+   * Unfreezes the trust line between this account (issuing account) and another account.
+   * Note that the trust line's limit is set to 0.
+   *
+   * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
+   *
+   * @param accountToUnfreeze The X-Address of the account which to unfreeze a trust line.
```
amiecorso, line 488: same note here about this confusing name...  I'd also call this `trustlineAccount` and the second parameter `trustlineCurrencyName` or just `trustlineCurrency` ...

amiecorso, line 488:
```suggestion
   * @param accountToUnfreeze The X-Address of the account involved in the trust line being unfrozen.
```

amiecorso: lol

src/XRP/issued-currency-client.ts
amiecorso: I think that we can operate on the assumption that gateways will never be using trustlines with non-zero limits extended to a counterparty... so we can probably get await without an `amount` param.  However, if we find we need to change this in future we can just add the optional param as a non-breaking change so we can do less now.

amiecorso (approved):
approving now but check my notes about the names and doc

mvadari (approved):

src/XRP/issued-currency-client.ts
tedkalaw: got it, per our conversation off line i changed this to `trustLinePeerAccount`

tedkalaw: got it

PR #862, opened 2020-10-30
Adds `sendIssuedCurrencyPayment` Method to `IssuedCurrencyClient`

amiecorso:
This PR adds `sendIssuedCurrencyPayment` to `IssuedCurrencyClient`, which is an opinionated public API for sending issued currency from one (non-issuing) account to another, and is parameterized with only the minimum, required information to achieve this.

### Context of Change
IOUs MVP

### Type of Change

## Before / After
`sendIssuedCurrencyPayment` exists on `IssuedCurrencyClient`

## Test Plan
Integration test for the method, CI passes

## Future Task

Do this for `sendCrossCurrencyPayment`, and `exchangeCurrency`


amiecorso:
> oof ouch owie that diff coverage

hahaha yeahhhh I think I realize where I need more tests

src/XRP/issued-currency-client.ts
```diff
@@ -689,6 +689,62 @@ export default class IssuedCurrencyClient {
...
+      throw new XrpError(
+        XrpErrorType.InvalidInput,
+        'The destination address cannot be the same as the issuer. To redeem issued currency, use `redeemIssuedCurrency`.',
+      )
+    }
+    // TODO: (acorso) A payment like this might fail if:
+    // - the issuer has not enabled rippling on its account
+    // - the issuer has enabled authorized trust lines, and the recipient of the payment does NOT have an authorized trust line to the issuer
+    // We may want to decide if it's worth incurring the cost of querying for this information BEFORE sending a payment that will destroy
+    // a transaction cost.  Or maybe we provide a different public method that can perform these checks if the user cares.
```
tedkalaw, line 735: i think you could get away with including this in the docs for now...

might be good to track this with a task somehow as well

src/XRP/issued-currency-client.ts
```diff
@@ -704,10 +760,15 @@ export default class IssuedCurrencyClient {
...
+   * @param issuer The issuer to specify in the destination amount field of the payment, encoded as an X-address.
+   *               Typically (but not always) the original issuer of the currency.  See https://xrpl.org/payment.html#special-issuer-values-for-sendmax-and-amount.
+   * @param amount The amount of issued currency to pay to the destination.
+   * @param transferFee (Optional) The transfer fee associated with the issuing account, expressed as a percentage. (i.e. a value of .5 indicates
+   *                    a 0.5% transfer fee).  Supply this field for automatic calculation of the sendMax value for this payment.
+   *                    Either this or sendMaxvalue may be specified, but not both.
+   * @param sendMaxValue (Optional) A manual specification of the maximum amount of source currency this payment is allowed to cost,
+   *                      including transfer fees, exchange rates, and slippage. Does not include the XRP destroyed as a cost for submitting \
+   *                      the transaction. Either this or transferFee may be specified, but not both.
```
tedkalaw, line 771: nit - the indentation of the newlines is different for each of these comments. i think first indentation level for issuer is the correct one

src/XRP/issued-currency-client.ts
```diff
@@ -783,10 +853,25 @@ export default class IssuedCurrencyClient {
...
+      const sendMaxCurrencyAmount = new CurrencyAmount()
+      sendMaxCurrencyAmount.setIssuedCurrencyAmount(sendMaxIssuedCurrencyAmount)
+
+      const sendMax = new SendMax()
+      sendMax.setValue(sendMaxCurrencyAmount)
+
+      payment.setSendMax(sendMax)
+    }
+    // manual specification of sendMax value
+    if (sendMaxValue !== undefined) {
```
tedkalaw, line 874: these two logical statements are huge and mutually exclusive; it seems like at the very least there should be an `else` here?

also, the code in these two else blocks is identical. it seems like you could probably do something like

```
// This would go outside of this function
private function calculateSendMax(transferFee: string): string {
  // do the calculation
  return transferFee;
}

// in issuedCurrencyPayment
...
const sendMaxAmount = transferFee !== undefined
  ? this.calculateSendMax(transferFee)
  : sendMaxValue

// construct sendmax
// set it on the payment
```

src/XRP/issued-currency-client.ts
```diff
@@ -783,10 +853,25 @@ export default class IssuedCurrencyClient {
       const rawSendMaxValue = (1 + transferFee / 100) * numericValue
       const numDigitsLeftOfDecimal = String(Math.ceil(rawSendMaxValue)).length
       const allowedDigitsRightOfDecimal = 15 - numDigitsLeftOfDecimal
-      const sendMaxValue = String(
+      const calculatedSendMaxValue = String(
```
tedkalaw, line 856: i know this is from before but it's more idiomatic to use `Number.toString` instead of the `String` constructor

also - `rawSendMaxValue.toFixed(allowedDigitsRightOfDecimal)` is already a string so i'm not sure you need to do this?

tedkalaw (changes requested):

src/XRP/issued-currency-client.ts
```diff
@@ -674,8 +674,8 @@ export default class IssuedCurrencyClient {
    */
   public async redeemIssuedCurrency(
     sender: Wallet,
```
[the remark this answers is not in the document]
amiecorso: I'm going to make an Asana task for this since I agree, but figure it'll be best to do all at once and once everything's implemented so we can guarantee consistency.  https://app.asana.com/0/1190819649053561/1199151116842789/f

src/XRP/issued-currency-client.ts
amiecorso: Removed comment, created a task: https://app.asana.com/0/1190819649053561/1199151116842791/f

src/XRP/issued-currency-client.ts
amiecorso: mmm yes, this is a great idea.  and that'll actually allow me to write some unit tests specifically for the sendMax calculation which I think I need to do.

src/XRP/issued-currency-client.ts
amiecorso: ohh I think you're right -- thought it left it as a number but nah

src/XRP/issued-currency-client.ts
```diff
@@ -860,4 +911,34 @@ export default class IssuedCurrencyClient {
...
+  /**
+   * Calculates the sendMaxValue by applying the transferFee to the amount being sent and ensuring the total amount
+   * fits within maximum decimal precision.
+   * @see https://xrpl.org/currency-formats.html#issued-currency-precision
+   *
+   * @param amount The amount of issued currency to pay to the destination.
+   * @param transferFee The transfer fee associated with the issuing account, expressed as a percentage.
+   *                    (i.e. a value of .5 indicates a 0.5% transfer fee).
+   */
+  public calculateSendMaxValue(amount: string, transferFee: number): string {
```
tedkalaw, line 924: NICE, this is far easier to understand

one minor architecture thing: it's a bit strange that `IssuedCurrencyClient` exposes `calculateSendMaxValue`, since it doesn't seem like something that a caller would care about. i'm guessing we're doing this so that it's testable.

i think this is fine, but one way you could fix this is by making a module called something like `IssuedCurrencyUtil` that has this function in it. that way, you can still test it, but it keeps the abstraction tighter. the downside is another level of indirection; the upside is that IssuedCurrencyClient becomes much dumber

in general, you don't really want to test private functions because they tend to be implementation details, and you're often better off testing them indirectly via the public interface. that being said, you could also make this function private and use `@ts-ignore` to ignore the private typecheck. types aren't "real" anyway

src/XRP/issued-currency-client.ts
```diff
@@ -860,4 +911,34 @@ export default class IssuedCurrencyClient {
...
+
+    // calculate the total sendMaxValue with any number of decimal places.
+    const rawSendMaxValue: BigNumber = numericAmount.multipliedBy(transferRate)
+
+    if (rawSendMaxValue.toFixed().length <= 15) {
+      return rawSendMaxValue.toFixed()
+    }
+
+    // If we have too many digits of precision, express with scientific notation but fit within decimal precision.
+    return rawSendMaxValue.toExponential(14, 0).replace(/e\+/, 'e')
```
tedkalaw, line 942: might be good to make consts for these magic numbers

tedkalaw (approved):

src/XRP/issued-currency-client.ts
amiecorso: yeah so... I actually thought about these considerations as well, and the reason I decided to just make it simple and leave it public is:  I think it may likely be the case that a user of the SDK will actually want to be able to calculate the sendMax value separately from this `sendIssuedCurrencyPayment` method in order to decide if it's viable/cheap enough, before actually submitting the transaction...  So I think for now I want to leave it public bc I can actually see us using this in our demo, and if we end up not needing it or can think of a better way to obscure it (or feel that it makes more sense in a utility class alongside maybe some other stuff we end up writing) I think we should move it then.

src/XRP/issued-currency-client.ts
amiecorso: yeah can do I'll call this thing maxDecimalPrecision or something

mvadari (approved):

PR #871, opened 2020-11-05
Adds IssuedCurrencyClient::disableRipplingForTrustLine

tedkalaw:
This PR adds `IssuedCurrencyClient:: disableRipplingForTrustLine `, which disables rippling on a specified trust line between two accounts.

### Context of Change
https://xrpl.org/trust-lines-and-issuing.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`IssuedCurrencyClient` can now set `noRippling` on a specific trust line instead of at the account level

## Test Plan
CI, the test I added


src/XRP/issued-currency-client.ts
```diff
@@ -505,6 +505,32 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @param trustLinePeerAccount The X-Address of the account involved in the trust line to disable rippling.
+   * @param currencyName The currency of the trust line to disable rippling.
+   * @amount The maximum amount of debt to allow on this trust line.
+   * @param wallet The wallet disabling rippling on the trust line.
+   */
+  public async setNoRippling(
+    trustLinePeerAccount: string,
+    currencyName: string,
+    amount: string,
```
tedkalaw, line 522: i found that if an existing trust line existed between the account and peer account, even if it was authorized, setting this to 0 would destroy this trust line. if, however, this was non-zero, then the `noRipple` would be properly set

mvadari (approved):

src/XRP/issued-currency-client.ts
amiecorso: On some level it seems weird that it destroys the trustline... but I suppose that if you have a trustline that has 0 balance AND is frozen it's effectively useless and would need to be re-established by whoever has the power to unfreeze it anyway.

amiecorso: Looking at the test... I wonder if a `noRipple` trustset transaction with a 0 limit would destroy the trustline if the counterparty had extended a trustline with a non-zero balance to the issuer?  I don't think it really matters for our purposes here since we can see that at least in the non-zero case the flag gets set.  The rest of this logic seems like rippled curiosities to me.

src/XRP/issued-currency-client.ts
```diff
@@ -505,6 +505,32 @@ export default class IssuedCurrencyClient {
     )
   }
 
+  /**
+   * Disables rippling on the trust line between this account (issuing account) and another account.
+   * Note that the trust line's limit is set to 0.
```
amiecorso, line 510: Is this true??  I don't see this noted in the documentation, and the final assert statement in your test suggests otherwise.

src/XRP/issued-currency-client.ts
```diff
@@ -505,6 +505,32 @@ export default class IssuedCurrencyClient {
...
+   *
+   * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
+   *
+   * @param trustLinePeerAccount The X-Address of the account involved in the trust line to disable rippling.
+   * @param currencyName The currency of the trust line to disable rippling.
+   * @amount The maximum amount of debt to allow on this trust line.
```
amiecorso, line 516:
```suggestion
   * @param amount The maximum amount of debt to allow on this trust line.
```

amiecorso (approved):

src/XRP/issued-currency-client.ts
```diff
@@ -505,6 +505,32 @@ export default class IssuedCurrencyClient {
     )
   }
 
+  /**
+   * Disables rippling on the trust line between this account (issuing account) and another account.
+   * Note that the trust line's limit is set to 0.
+   *
+   * @see https://xrpl.org/freezes.html#enabling-or-disabling-individual-freeze
```
amiecorso, line 512:
```suggestion
   * @see https://xrpl.org/rippling.html#enabling-disabling-no-ripple
```

src/XRP/issued-currency-client.ts
tedkalaw: it's false, deleted

src/XRP/issued-currency-client.ts
tedkalaw: got it

src/XRP/issued-currency-client.ts
tedkalaw: lol

PR #876, opened 2020-11-06
Refactors `XrpClient` integration tests to use `sendXrp` instead of deprecated `send`

amiecorso:
This PR refactors the `XrpClient` integration tests to use the improved `sendXrp` method (which returns a full transaction result) instead of the deprecated `send` method that only returned a hash.

### Type of Change

## Before / After
All `XrpClient` tests should now be testing the right method.

## Test Plan
CI still passes.


tedkalaw (approved):
thank you!!

test/XRP/xrp-client-integration.test.ts
```diff
@@ -43,33 +43,35 @@ describe('XrpClient Integration Tests', function (): void {
...
-    const transactionHash = await xrpWebClient.send(
+    const transactionResult = await xrpWebClient.sendXrp(
       amount,
       recipientAddress,
       wallet,
     )
     const transactionStatus = await xrpWebClient.getPaymentStatus(
-      transactionHash,
+      transactionResult.hash,
     )
```
mvadari, line 53: Couldn't you discard this and just use `transactionResult.status` instead of calculating `transactionStatus`?

amiecorso: oh you mean like capitalize on the entire point of this refactor in the first place?  not a bad idea lol

amiecorso: changed everywhere possible in this file, nice catch mayukha hahaha

tedkalaw (approved):
cardi-yeee

mvadari (approved):

mvadari (changes requested):
You're missing some other test files, like the issued currency client and web socket integration tests

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -223,33 +223,29 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
       wallet.getAddress(),
       wallet2,
     )
-    const transactionStatus = await xrpClient.getPaymentStatus(transactionHash)
 
     // THEN the transaction fails.
-    assert.exists(transactionHash)
-    assert.equal(transactionStatus, TransactionStatus.Failed)
+    assert.exists(transactionResult.hash)
+    assert.equal(transactionResult.status, TransactionStatus.ClaimedCostOnly)
```
amiecorso, line 234: Using the actual `TransactionResult`, I noticed that the original source of the transaction status (`getPaymentStatus`) was not creating a granular enough distinction from the raw server transaction code (like `'tecPATH_DRY` for example).  These tests are now correct but I've made a task to update our logic around converting from raw string code to a `TransactionStatus` so that we can make sure that the core logic exists only once and is reused wherever this conversion is happening: https://app.asana.com/0/1190819649053561/1199159961044989/f

amiecorso:
> You're missing some other test files, like the issued currency client and web socket integration tests

ugh, yep, good catch.  This update also illuminated something else that needs cleaning which you'll see in my comments.

mvadari (approved):

PR #885, opened 2020-11-10
Implements `findRipplePath` on `WebSocketNetworkClient`

mvadari:

In order to implement cross-currency payment, we need to be able to find paths from the source to the address. This PR implements the API call to the ledger that allows us to do that.

### Context of Change

IOUs

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

The web socket network client has a new method, `findRipplePath`

## Test Plan

Added new tests. CI passes.

## Future Tasks
Implement cross-currency payment


src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -275,6 +279,40 @@ export default class WebSocketNetworkClient {
     return accountOffersResponse as AccountOffersResponse
   }
 
+  /**
+   * Submits a ripple_path_find request to the rippled WebSocket API.
+   * @see https://xrpl.org/ripple_path_find.html
+   *
+   * @param sourceAccount
```
amiecorso, line 286: missing param docs here

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -275,6 +279,40 @@ export default class WebSocketNetworkClient {
...
+    sourceAccount: string,
+    destinationAccount: string,
+    destinationAmount: string | IssuedCurrency,
+    sendMax?: string | IssuedCurrency,
+    sourceCurrencies?: SourceCurrency[],
+  ): Promise<RipplePathFindResponse> {
+    const ripplePathFindRequest: RipplePathFindRequest = {
+      id: `${RippledMethod.ripplePathFind}_${sourceAccount}_${this.idNumber}`,
+      command: RippledMethod.ripplePathFind,
+      source_account: sourceAccount,
```
amiecorso, line 298: does this assume that sourceAccount and destinationAccount are encoded as classic addresses?  we might want to include that in the documentation, or add a check where if they're X-addresses, they get decoded before being submitted

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -275,6 +279,40 @@ export default class WebSocketNetworkClient {
...
+      destination_account: destinationAccount,
+      destination_amount: destinationAmount,
+      send_max: sendMax,
+      source_currencies: sourceCurrencies,
+    }
+    this.idNumber++
+
+    console.log('ripplePathFindRequest: ')
+    console.log(ripplePathFindRequest)
```
amiecorso, line 307: probably want to get rid of these

amiecorso: I probably left these here haha

test/XRP/web-socket-network-client.test.ts
```diff
@@ -27,20 +30,28 @@ const webSocketNetworkClient = new WebSocketNetworkClient(
...
+const grpcNetworkClient = new GrpcNetworkClient(rippledGrpcUrl)
+
 function sleep(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms))
 }
 
 describe('WebSocket Tests', function (): void {
   // Retry integration tests on failure.
-  this.retries(3)
+  //this.retries(3)
```
amiecorso, line 41: sorry something I also left.... we could probably get away with removing this too.  this was a relic from when we weren't generating new wallets for new test runs, and simultaneous CI runs were creating sequence number clashes that created occasional failures.

test/XRP/web-socket-network-client.test.ts
```diff
@@ -266,4 +277,298 @@ describe('WebSocket Tests', function (): void {
...
+      value: '200',
+    }
+
+    const takerPaysAmount = '200000000' // 200 XRP, 1:1 exchange rate
+    const offerCreateResult = await issuedCurrencyClient.createOffer(
+      offerCreatorWallet,
+      takerGetsAmount,
+      takerPaysAmount,
+    )
+    console.log('result of creating offer: ')
```
amiecorso, line 500: couple of log statements in this test that we can take out

amiecorso (approved):
This is looking pretty good.  Just some cleanup comments... I think RE the sendMax case... we have a test in there for the special sendMax case.  I think that's enough -- I don't think we need to try to get the rest of sendMax figured out right here right now (it is, after all, effectively testing rippled logic -- we can see already that a call to `ripple_path_find` is working.

amiecorso:
Realizing this from work on the cross currency payment PR, but could we add `findRipplePath` to the `WebSocketNetworkClientInterface` and then also to the `FakeWebSocketNetworkClient` in this PR?

tedkalaw (approved):

PR #893, opened 2020-11-11
Adds `createOffer` to `IssuedCurrencyClient`

amiecorso:
This PR creates a `createOffer` API on the `IssuedCurrencyClient` that essentially mirrors the rippled JSON API for creating offers in the DEX.  While we initially considered creating/cancelling offers to be something that could be left to V2 functionality, we realized that creating offers is essential to being able to test cross-currency payments and currency exchange payments... They're also relatively simple, so we might as well implement.

NOTE: Used npm link on `xpring-common-js` to develop this locally while I wait for a serializing/signing bug fix for offer transactions to make it through release process.  This will fail CI until the next `xpring-common-js` release / version bump.

### Context of Change
IOU MVP

### Type of Change

## Before / After
`createOffer` API exists on `IssuedCurrencyClient`

## Test Plan
New unit/integration tests added for `createOffer`.

## Future Tasks

Equivalent API for `OfferCancel` transactions


src/XRP/issued-currency-client.ts
```diff
@@ -1024,4 +1033,107 @@ export default class IssuedCurrencyClient {
...
+   *               (See https://xrpl.org/basic-data-types.html#specifying-time)
+   */
+  public async createOffer(
+    sender: Wallet,
+    takerGetsAmount: string | IssuedCurrency,
+    takerPaysAmount: string | IssuedCurrency,
+    offerSequence?: number,
+    expiration?: number,
+  ): Promise<TransactionResult> {
+    // construct TakerGets
```
tedkalaw, line 1060: remove this?

src/XRP/shared/issued-currency.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Represents an issued currency on the XRP Ledger.
+ * @see https://xrpl.org/currency-formats.html#issued-currency-amounts
+ */
+export default interface IssuedCurrency {
+  issuer: string
+  currency: string
+  value: string
```
tedkalaw, line 8: alpha?

src/XRP/issued-currency-client.ts
```diff
@@ -1024,4 +1033,107 @@ export default class IssuedCurrencyClient {
...
+
+      const takerPaysIssuedCurrencyAmount = new IssuedCurrencyAmount()
+      takerPaysIssuedCurrencyAmount.setIssuer(takerPaysAccountAddress)
+      takerPaysIssuedCurrencyAmount.setCurrency(takerPaysCurrency)
+      takerPaysIssuedCurrencyAmount.setValue(takerPaysAmount.value)
+
+      takerPaysCurrencyAmount.setIssuedCurrencyAmount(
+        takerPaysIssuedCurrencyAmount,
+      )
+    }
```
mvadari, line 1106: Since this logic is identical for takerGets and takerPays, you could put it in a helper function

mvadari (approved):

src/XRP/issued-currency-client.ts
```diff
@@ -1024,4 +1033,107 @@ export default class IssuedCurrencyClient {
...
+   */
+  public async createOffer(
+    sender: Wallet,
+    takerGetsAmount: string | IssuedCurrency,
+    takerPaysAmount: string | IssuedCurrency,
+    offerSequence?: number,
+    expiration?: number,
+  ): Promise<TransactionResult> {
+    // construct TakerGets
+    const takerGetsCurrencyAmount = new CurrencyAmount()
```
tedkalaw, line 1061: this is super cool, but i think it may be better to have this be a `string` to start, since it matches the rest of our apis. we don't rly seem to have too many apis that use protobuffs in this way (i guess we could though?)

 it's also a bit less risky to start with a more opinionated/restrictive api and make it accept more later...would strongly recommend keeping this to be just a string to start and expanding later if we like the approach of having a protobuf passed in

tedkalaw (approved):
this all seems fairly reasonable. would slightly prefer not having IssuedCurrency be a type passed in, but up to you - it doesn't match our other apis

src/XRP/issued-currency-client.ts
amiecorso: good idea, done

src/XRP/issued-currency-client.ts
amiecorso: hmmm, I'm not sure what you mean here?  it can't just be a string because that's how you specify an XRP amount, but we have to be able to specify both an XRP amount AND an issued currency amount (especially bc if either one is an XRP amount, the other one has to be issued currency (otherwise it's not an issued currency).  the opposite isn't true, since there could be two different issued currencies specified).  Also, here an `IssuedCurrency` isn't a protobuf, it's an interface that we've defined to bundle the three piece of information together.  You're right that our other APIs actually don't take `IssuedCurrencies` and instead accept the information piecemeal... something that we have an asana task out to change, bc this is much more logical.

PR #908, opened 2020-11-18
Implements `getAccountOffers` on `WebSocketNetworkClient`

mvadari:

Now that we have offers implemented, we need to be able to retrieve info about outstanding offers for accounts, via the API. https://xrpl.org/account_offers.html

### Context of Change

IOUs

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

The web socket network client has a new method, `getAccountOffers`

## Test Plan

Added new tests. CI passes.


amiecorso (approved):
yeet

src/XRP/shared/rippled-web-socket-schema.ts
```diff
@@ -276,21 +316,24 @@ interface CurrencyValuePair {
...
   GatewayBalancesResponse,
   GatewayBalancesSuccessfulResponse,
-  WebSocketReadyState,
-  RippledMethod,
-  ResponseStatus,
-  SubscribeRequest,
-  AccountLinesRequest,
-  GatewayBalancesRequest,
+  AccountOffersResponse,
+  AccountOffersSuccessfulResponse,
```
ledhed2222, line 336: why did the order of these exports change? not a big deal, but messes with our commit history

test/XRP/web-socket-network-client.test.ts
```diff
@@ -176,4 +181,89 @@ describe('WebSocket Tests', function (): void {
       }
     }
   })
+
+  it('getAccountOffers - valid requests', async function (): Promise<void> {
```
ledhed2222, line 185: why are we using the `function` keyword instead of arrow functions?

```javascript
it('getAccountOffers - valid requests', async (): Promise<void> => {
   ...
})
```

test/XRP/web-socket-network-client.test.ts
```diff
@@ -176,4 +181,89 @@ describe('WebSocket Tests', function (): void {
...
+    await issuedCurrencyClient.createOffer(
+      wallet,
+      takerGetsIssuedCurrency,
+      takerPaysXrp,
+    )
+
+    // WHEN getAccountOffers is called for that address
+    const accountOfferResponse2 = await webSocketNetworkClient.getAccountOffers(
+      address,
+    )
```
ledhed2222, line 233: there's an opportunity to use `Promise.all` here instead of two successive `await`s:

```javascript
const issuedCurrencyPromise = issuedCurrencyClient.createOffer(
  wallet,
  takerGetsIssuedCurrency,
  takerPaysXrp,
)
const accountOfferPromise = webSocketNetworkClient.getAccountOffers(
  address,
)
const [_issuedCurrencyResponse, accountOfferResponse2] = await Promise.all([issuedCurrencyPromise, accountOfferPromise])
```

test/XRP/web-socket-network-client.test.ts
```diff
@@ -176,4 +181,89 @@ describe('WebSocket Tests', function (): void {
...
+
+    const offer = result2.offers[0]
+
+    assert.equal(offer.taker_pays, takerPaysXrp)
+    assert.deepEqual(offer.taker_gets, takerGetsIssuedCurrency)
+
+    issuedCurrencyClient.webSocketNetworkClient.close()
+  })
+
+  it('getAccountOffers - bad address', async function (): Promise<void> {
```
ledhed2222, line 253: arrow function

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -1,19 +1,21 @@
...
   GatewayBalancesRequest,
+  AccountOffersRequest,
   WebSocketResponse,
   WebSocketFailureResponse,
+  ResponseStatus,
+  AccountLinesResponse,
+  GatewayBalancesResponse,
+  StatusResponse,
+  TransactionResponse,
+  AccountOffersResponse,
```
ledhed2222, line 18: why did the order of these imports change? it doesn't matter but it will mess up commit history. just curious.

src/XRP/shared/rippled-web-socket-schema.ts
mvadari: I just reordered them so the order made more sense, instead of being really messy.

test/XRP/web-socket-network-client.test.ts
mvadari: I think that's how all the other test files do it (and maybe eslint gets mad at it?)

test/XRP/web-socket-network-client.test.ts
mvadari: Does this still apply if I want the second `await` to only run after the first one has finished? Because that's the case here.

src/XRP/network-clients/web-socket-network-client.ts
mvadari: see above comment

test/XRP/web-socket-network-client.test.ts
mvadari: see above comment

test/XRP/web-socket-network-client.test.ts
ledhed2222: ah i see

ledhed2222 (approved):

PR #915, opened 2020-11-21
[DO NOT MERGE] Expose `IssuedCurrencyClient` for CBDC Demo Build

amiecorso:
This branch will serve as a place to expose the `IssuedCurrencyClient` and related infrastructure so it can be used to build a CBDC demo and test the API before releasing publicly.
This can be achieved for local development with `npm link`, or we may choose to temporarily publish a private npm package.

### Context of Change
CBDC demo, issued currency dogfood

## Test Plan
Crash and burn.


package.json
```diff
@@ -22,6 +22,7 @@
     "lint": "eslint . --ext .ts --fix",
     "lintNoFix": "eslint . --ext .ts",
     "test": "nyc mocha",
+    "prepare": "git submodule update --init --recursive && npm run build",
```
dino-rodriguez, line 25: Added this prepare script to trigger `xpring-js` build when pulling this in locally as a `git` branch through `npm`.

"NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed."

From: https://docs.npmjs.com/cli/v6/using-npm/scripts (for more information on what the prepare script does)

package.json
```diff
@@ -72,13 +73,14 @@
     "crypto-browserify": "^3.12.0",
     "google-protobuf": "^3.12.2",
     "grpc-web": "1.2.1",
+    "isomorphic-ws": "^4.0.1",
```
dino-rodriguez, line 76: Switched to `isomorphic-ws` for browser compatibility (we want to use `WebSocket` in the browser instead of `ws`).

scripts/init_mac.sh
```diff
@@ -0,0 +1,12 @@
+# Pull Submodules
```
dino-rodriguez, line 1: Added a script for OSX to set up a local development environment for `xpring-js` using `brew` according to the `CONTRIBUTING.md`.

src/XRP/network-clients/web-socket-network-client.ts
```diff
@@ -1,4 +1,4 @@
-import WebSocket = require('ws')
+import WebSocket = require('isomorphic-ws')
```
dino-rodriguez, line 1: Only reference found using grep.

src/XRP/issued-currency-client.ts
```diff
@@ -810,12 +814,14 @@ export default class IssuedCurrencyClient {
...
    * @param currency The three-letter currency code of the issued currency being redeemed.
    * @param amount The amount of issued currency to redeem.
+   * @param sendMaxValue The max amount to send.
    */
   public async redeemIssuedCurrency(
     sender: Wallet,
     currency: string,
     issuer: string,
     amount: string,
+    sendMaxValue?: string,
```
dino-rodriguez, line 824: When we were attempting to get transaction history for IOUs, we were getting an error specifying that we needed to provide a `sendMax`. From our slack convo, it sounds like this fix might already be included in another PR.

package.json
```diff
@@ -72,13 +73,14 @@
...
     "os-browserify": "^0.3.0",
     "stream-browserify": "^3.0.0",
     "url": "^0.11.0",
     "util": "^0.12.3",
-    "ws": "^7.3.1",
+    "ws": "^7.4.0",
     "xhr2": "0.2.0",
-    "xpring-common-js": "6.2.7"
+    "xpring-common-js": "xpring-eng/xpring-common-js#cbdc-demo"
```
dino-rodriguez, line 83: This should be updated once we release the changes in `xpring-common-js`.

package.json
amiecorso: Is this still a useful thing to have generally?  Or irrelevant outside the context of using xpring-js as a git branch?

package.json
amiecorso: sounds like we definitely want to go this route since xpring-js needs to be browser compatible

scripts/init_mac.sh
amiecorso: this also seems like a useful thing to leave in if it worked for you... maybe I should add a note about this in CONTRIBUTING

src/XRP/network-clients/web-socket-network-client.ts
amiecorso: yeah this should be the only ref

src/XRP/issued-currency-client.ts
amiecorso: this should already be fixed via another PR (already merged) so we definitely don't want this additional param here.

package.json
amiecorso: gotcha, which is why you suggested first cutting a release of xpring-common-js

package.json
dino-rodriguez: The things it does:

```
Runs BEFORE the package is packed
Runs BEFORE the package is published
Runs on local npm install without any arguments
Run AFTER prepublish, but BEFORE prepublishOnly
NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.
```

I would say probably irrelevant unless one of the use cases stands out to you, mainly because you already have a `prePublishOnly` script anyways.

theotherian: Yeah, this _seems_ irrelevant to me since we're getting away from the branch dependency model.

package.json
dino-rodriguez: yup

theotherian:
Fixed the merge conflicts.

mvadari (approved):

theotherian (approved):

PR #940, opened 2020-12-01
Refactors `IssuedCurrencyClient` methods to use the `IssuedCurrency` interface

mvadari:
To clean up the IOU functions, this PR replaces all instances where there are 3 parameters in a function for the 3 elements of an issued currency (currency, value, issuer) with an issued currency interface. 

### Context of Change

Cleaning up IOU code before deployment

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Where possible, the public IOU methods use `IssuedCurrency` interfaces instead of 3 separate parameters.

## Test Plan

CI passes.

### DIFF
--- a/src/XRP/issued-currency-client.ts
+++ b/src/XRP/issued-currency-client.ts
@@ -793,13 +793,12 @@ export default class IssuedCurrencyClient {
     amount: string,
   ): Promise<TransactionResult> {
     const issuer = sender.getAddress()
-    return await this.issuedCurrencyPayment(
-      sender,
-      destination,
+    const issuedCurrency = {
       currency,
       issuer,
-      amount,
-    )
+      value: amount,
+    }
+    return await this.issuedCurrencyPayment(sender, destination, issuedCurrency)
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/XRP/issued-currency-client.ts
```diff
@@ -793,13 +793,12 @@ export default class IssuedCurrencyClient {
...
-      sender,
-      destination,
+    const issuedCurrency = {
       currency,
       issuer,
-      amount,
-    )
+      value: amount,
+    }
+    return await this.issuedCurrencyPayment(sender, destination, issuedCurrency)
```
ledhed2222, line 801: I'm sure this is passing the typescript checks, but I'm confused by something. If this is supposed to return a `Promise<TransactionResult>` then wouldn't you `return this.issuesCurrencyPayment...`, IE without `await`? Also, IMO it might make more sense to in fact return the promise, so that the caller can have more flexibility on when it's triggered

src/XRP/issued-currency-client.ts
```diff
@@ -813,20 +812,21 @@ export default class IssuedCurrencyClient {
...
+      value: issuedCurrency.value,
+    }
     return await this.issuedCurrencyPayment(
       sender,
-      issuer,
-      currency,
-      sender.getAddress(),
-      amount,
+      issuedCurrency.issuer,
+      specialIssuedCurrency,
```
ledhed2222, line 829: same as above

src/XRP/issued-currency-client.ts
```diff
@@ -813,20 +812,21 @@ export default class IssuedCurrencyClient {
...
   ): Promise<TransactionResult> {
     // Redemption of issued currency is achieved by sending issued currency directly to the original issuer.
     // However, the issuer field specified in the amount is treated as a special case in this circumstance, and should be
     // set to the address of the account initiating the redemption.
     // See: https://xrpl.org/payment.html#special-issuer-values-for-sendmax-and-amount
+    const specialIssuedCurrency: IssuedCurrency = {
+      currency: issuedCurrency.currency,
+      issuer: sender.getAddress(),
+      value: issuedCurrency.value,
+    }
```
ledhed2222, line 825: I think this might look cleaner like this. What do you think?

```javascript
const issuedCurrency: IssuedCurrency = {
  ...issuedCurrency,
  issuer: sender.getAddress(),
}
```

ledhed2222 (approved):

amiecorso (approved):
Thanks for doing this one! I added some documentation to the `IssuedCurrency` interface since we're losing it in the params of these methods... other than that I agree with Greg's comments (although this may mean we want to make sure ALL of our methods behave this way to maintain consistency).  This looks good.

tedkalaw (approved):

PR #942, opened 2020-12-01
Refactors integration tests to use `XrpTestUtils.randomWalletFromFaucet` instead of deprecated `Wallet.generateRandomWallet`

mvadari:

This PR refactors all the integration test files to stop using the now-deprecated method `Wallet.generateRandomWallet` and instead use `XrpTestUtils.randomWalletFromFaucet`. 

### Context of Change

Code cleanup

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
CI passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,10 +125,11 @@ Xpring-JS can generate a new and random HD Wallet. The result of a wallet genera
 - A reference to the new wallet
 
 ```javascript
-const { Wallet } = require('xpring-js')
+const { WalletFactory, XrplNetwork } = require('xpring-js')
 
 // Generate a random wallet.
-const generationResult = Wallet.generateRandomWallet()
+const walletFactory = new WalletFactory(XrplNetwork.Test)
+const generationResult = await walletFactory.generateRandomWallet()!
 const newWallet = generationResult.wallet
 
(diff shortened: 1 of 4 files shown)


ledhed2222 (commented):
Looks good to me, but from my `grep`-ing on master looks like you missed: `README`, `test/XRP/helpers/xrp-test-utils`, `test/XRP/issued-currency-client-integration.test.js`, `test/XRP/issued-currency-client.test.js`

mvadari:
Missed `README`, thanks. The rest are all `walletFactory.generateRandomWallet`, which is not deprecated

ledhed2222 (approved):

amiecorso (commented):
Another thanks! So my only note here is that all of the places where we were using `Wallet.generateRandomWallet()`... we should use `walletFactor.generateRandomWallet()` and NOT `XRPTestUtils.randomWalletFromFaucet()` because the faucet wallet is actually creating a funded wallet on the XRPL (takes time), and is not necessary for anything that's not an integration test because we're not actually hitting the ledger.  `generateRandomWallet` generates a fresh, valid address, it just doesn't actually exist on the ledger which is fine for these unit tests.

amiecorso (approved):
looks great!

PR #947, opened 2020-12-04
Removes Unnecessary Check from Construction of `XrpPayment` Object

amiecorso:
This PR removes a safety check in the construction of `XrpPayment` objects that wanted to see a value for the `sendMax` field in ANY payment that involved an issued currency.  This was probably just due to a misinterpretation of the documentation, but now that we are issued currency experts, we know that issued currency payments can actually omit the `sendMax` sometimes, since it is only _required_ for cross-currency payments (not required for issuances or redemptions).

(The documentation reads "For non-XRP amounts, the nested field names MUST be lower-case. Must be supplied for cross-currency/cross-issue payments. Must be omitted for XRP-to-XRP payments.")

This came to light through the cbdc demo work, since their attempt to get an account's payment history using the `XrpClient.paymentHistory` was throwing while trying to construct `XrpPayment` objects from issued currency payment transactions.

### Context of Change
IOU MVP Dogfood

### Type of Change

## Test Plan
Wrote myself an integration test that verifies `paymentHistory` can handle issued currency payments once this check is removed.  It is not included here because I don't want to mix clients yet, or add time to our integration tests if we can avoid it.

### DIFF
--- a/src/XRP/protobuf-wrappers/xrp-payment.ts
+++ b/src/XRP/protobuf-wrappers/xrp-payment.ts
@@ -74,13 +74,6 @@ export default class XrpPayment {
         'Payment protobuf `sendMax` field should be omitted for XRP-to-XRP interactions.',
       )
     }
-    // For non-XRP amounts, the nested field names in `sendMax` MUST be lower-case.
-    if (sendMax === undefined && amount.issuedCurrency !== undefined) {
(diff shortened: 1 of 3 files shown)


mvadari (approved):
whoops

tedkalaw (approved):

PR #948, opened 2020-12-05
Adds flag optional params to `IssuedCurrencyClient::createOffer`

mvadari:

The `CreateOffer` transaction type has flag options that can be applied to the resulting offers. This PR adds those as optional parameters to the `createOffer` method in `IssuedCurrencyClient`.

### Context of Change

IOUs

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

`IssuedCurrencyClient::createOffer` now has additional optional parameters for the flags

## Test Plan

Added tests for the flags, including invalid combinations. CI passes.


src/XRP/issued-currency-client.ts
```diff
@@ -1222,13 +1223,30 @@ export default class IssuedCurrencyClient {
    *               It is not considered an error if the offer specified does not exist.
    * @param expiration (Optional) Time after which the offer is no longer active, in seconds since the Ripple Epoch.
    *               (See https://xrpl.org/basic-data-types.html#specifying-time)
+   * @param passive (Optional) If enabled, the offer does not consume offers that exactly match it, and instead becomes
```
amiecorso, line 1226: I wonder if we should specify "Defaults to false." for all of these opt flag params?

test/XRP/issued-currency-client-integration.test.ts
```diff
@@ -873,6 +873,98 @@ describe('IssuedCurrencyClient Integration Tests', function (): void {
...
+
+    const offerSequenceNumber = 1
+
+    const rippleEpochStartTimeSeconds = 946684800
+    const currentTimeUnixEpochSeconds = Date.now() / 1000 // 1000 ms/sec
+    const currentTimeRippleEpochSeconds =
+      currentTimeUnixEpochSeconds - rippleEpochStartTimeSeconds
+    const expiration = currentTimeRippleEpochSeconds + 60 * 60 // roughly one hour in future
+
+    // WHEN the wallet creates an offer to exchange (receive) their own issued currency for their XRP (deliver)
```
amiecorso, line 948: make a note about the flags we're testing here in the WHEN statement?  it seems that immediateOrCancel and fillOrKill are mutually exclusive? but it's not noted here

src/XRP/shared/offer-create-flags.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Flags used in OfferCreate transactions.
+ *
+ * @see https://xrpl.org/offercreate.html#offercreate-flags
+ */
+class OfferCreateFlags {
+  static TF_PASSIVE = 0x00010000
+  static TF_IMMEDIATE_OR_CANCEL = 0x00020000
+  static TF_FILL_OR_KILL = 0x00040000
+  static TF_SELL = 0x00080000
+}
```
tedkalaw, line 11: any reason to not use an enum for these instead? they end up being implemented pretty much the same under the covers - objects with named properties

tedkalaw (approved):

amiecorso (approved):
just a few notes!

